PR #4, opened 2020-09-02
feat(auth): add login page and user authentication

MrTimscampi:


plugins/axios.ts
```diff
@@ -0,0 +1,10 @@
+export default function ({ NuxtAxiosInstance$axios, store }) {
```
cromefire, line 1: Shouldn't this be typed?

Camc314 (approved):

plugins/axios.ts
MrTimscampi: This was an unremoved experiment :p

PR #11, opened 2020-09-03
feat(nuxt.config.ts): added Jellyfin accent color (blue) as primary V…

ThibaultNocchi:
…uetify color


ThibaultNocchi:
Well, it seems cz isn't counting the characters as it should ^^"

PR #15, opened 2020-09-04
feat(loginform.vue): using snackbar for login errors

ThibaultNocchi:
it follows the use of a similar snackbar in the current web client

### DIFF
--- a/components/LoginForm.vue
+++ b/components/LoginForm.vue
@@ -28,9 +28,9 @@
         >submit</v-btn
       >
     </v-form>
-    <v-alert v-if="errorMessage" class="mt-5" outlined type="error"
-      >{{ errorMessage }}
-    </v-alert>
+    <v-snackbar color="red" bottom left v-model="errorMessageSnackbar">
+      {{ errorMessage }}
+    </v-snackbar>
   </div>
 </template>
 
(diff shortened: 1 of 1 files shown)


ThibaultNocchi:
Here it is in red
print_screen_2020-09-04-11-07-41

Camc314:
How well does this work on mobile, if there is a keyboard covering ~50% of the screen?

ThibaultNocchi:
Ok they suck for now haha, they appear under the screen and you need to scroll. I'll look into it.

MrTimscampi:
You may need to add the `app` prop so it respects the layout. That might make it work with virtual keyboards (not sure, just a guess)

At some point, it would likely be better to use an event bus for this and make a notification component that would contain the snackbar, the code to handle messages and put that into the layouts directly instead of in individual components.

We could then use `this.$nuxt.$emit();` to emit app-wide events that could be caught by other components.

It's no big deal for now, but definitely something we want to look into.

ThibaultNocchi:
I do agree with the event bus, it'd be then easier to have a common component of a snackbar, with different levels for different colors, etc.

Also, our initial problem on mobile seems to be a bug, see https://github.com/vuetifyjs/vuetify/issues/11781#issuecomment-655689025

The given fix is a CSS one, so it'd totally be suitable to put it into a component.

MrTimscampi:
Fix sounds good to me :)

I'll create an issue to keep track of the event bus thing.

MrTimscampi:
For reference: #18

MrTimscampi (approved):

PR #25, opened 2020-09-04
Global snackbar

### LINKED ISSUE #18: Move snackbar to a dedicated component and use a Vuex store to pass state to it
The snackbar should be its own component, controlled by a Vuex store in order to be available across the app in a standard way.

A good example of an implantation would be the following, using a store and a plugin for easy access: https://dev.to/stephannv/how-to-create-a-global-snackbar-using-nuxt-vuetify-and-vuex-1bda

ThibaultNocchi:
This global snackbar can be called in Vue instances, stores and Nuxt contexts.

The method signature is $snackbar(message: string, color: string|undefined|null) and it will appear. If no color is specified, an empty string is given to the Vuetify component and it applies the default of the theme (black for our dark theme).

Closes #18

### DIFF
--- a/components/LoginForm.vue
+++ b/components/LoginForm.vue
@@ -28,9 +28,6 @@
         >submit</v-btn
       >
     </v-form>
-    <v-snackbar color="red" bottom left v-model="errorMessageSnackbar">
-      {{ errorMessage }}
-    </v-snackbar>
   </div>
 </template>
 
@@ -44,8 +41,6 @@ export default {
         pw: ''
       },
       showPassword: false,
-      errorMessageSnackbar: false,
-      errorMessage: '',
       validInputs: false,
       rules: {
         serverUrlTest: [
@@ -72,16 +67,15 @@ export default {
         this.$auth.setUser(response.data.User);
       } catch (error) {
         console.error('Failed to login:', error);
+        let errorMessage = 'Unexpected Error';
         if (!error.response) {
-          this.errorMessage = 'Server Not Found';
+          errorMessage = 'Server Not Found';
         } else if (error.response.status === 500) {
-          this.errorMessage = 'Incorrect Password';
+          errorMessage = 'Incorrect Password';
         } else if (error.response.status === 400) {
-          this.errorMessage = 'Bad Request. Try Again';
-        } else {
-          this.errorMessage = 'Unexpected Error';
+          errorMessage = 'Bad Request. Try Again';
         }
-        this.errorMessageSnackbar = true;
+        this.$snackbar(errorMessage, 'red');
       }
     }
   }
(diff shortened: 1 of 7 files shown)


components/Snackbar.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <v-snackbar v-model="model" :color="color" bottom left>
+    {{ message }}
+  </v-snackbar>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+export default Vue.extend({
```
MrTimscampi, line 9:
```suggestion
import Vue from 'vue';

export default Vue.extend({
```

components/LoginForm.vue
```diff
@@ -72,16 +67,15 @@ export default {
...
-          this.errorMessage = 'Incorrect Password';
+          errorMessage = 'Incorrect Password';
         } else if (error.response.status === 400) {
-          this.errorMessage = 'Bad Request. Try Again';
-        } else {
-          this.errorMessage = 'Unexpected Error';
+          errorMessage = 'Bad Request. Try Again';
         }
-        this.errorMessageSnackbar = true;
+        this.$snackbar(errorMessage, 'red');
```
MrTimscampi, line 78:
```suggestion
        this.$snackbar(errorMessage, 'error');
```

That way, it adapts to the user's color scheme.

components/Snackbar.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <v-snackbar v-model="model" :color="color" bottom left>
```
MrTimscampi, line 2: You can probably add the `app` prop here, so that it respects the layout of the various app components (Navigation drawer, footer, app bar, etc).

It's especially important once we set up a mini player in the app footer.

layouts/default.vue
```diff
@@ -72,12 +72,14 @@
         <nuxt />
       </v-container>
     </v-main>
+    <Snackbar></Snackbar>
```
MrTimscampi, line 75:
```suggestion
    <snackbar />
```

layouts/fullpage.vue
```diff
@@ -3,5 +3,13 @@
     <v-main>
       <nuxt />
     </v-main>
+    <Snackbar></Snackbar>
   </v-app>
 </template>
+
+<script lang="ts">
+import Snackbar from '../components/Snackbar.vue';
+export default {
```
MrTimscampi, line 12:
```suggestion
import Snackbar from '../components/Snackbar.vue';

export default {
```

MrTimscampi (changes requested):
Some small style issues.

ThibaultNocchi:
Done, and the `app` prop worked as intended over the nav drawer.

MrTimscampi (approved):

PR #37, opened 2020-09-05
refactor: removed component imports

ThibaultNocchi:
In nuxt.config.ts, the option 'components' is set to true. Nuxt auto imports used components in templates from the components/ directory.


MrTimscampi (approved):
Good to know, that's some nice cleanup :)

ThibaultNocchi:
Yeah, I was curious of this option when checking over the config file. :)

PR #41, opened 2020-09-05
"Not found" component

ThibaultNocchi:
This a custom component which display an "alert" icon in orange, with a message below. This message is by default "Not found" and can be customize by passing a `custom-message` prop to the component.

Also, there's a linting fix.

Maybe this could need discussion, as those sort of errors could be handled in another way. Such as with a redirection to a 404 page or something.

Here I used the /library/ page when the given ID is wrong to put it as a POC.

### DIFF
--- a/layouts/error.vue
+++ b/layouts/error.vue
@@ -1,0 +1,40 @@
+<template>
+  <div>
+    <v-row justify="center" class="mt-16 mb-4">
+      <v-icon :color="icon.color" size="72">{{ icon.icon }}</v-icon>
+    </v-row>
+    <v-row justify="center">
+      <h1 class="text-h4 text-center">{{ error.message }}</h1>
+    </v-row>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue, { PropType } from 'vue';
+import { NuxtError } from '@nuxt/types';
+
+interface Icon {
+  icon: string;
+  color: string | undefined | null;
+}
+
+export default Vue.extend({
+  props: {
+    error: {
+      type: Object as PropType<NuxtError>,
+      default: { statusCode: -1, message: '' }
+    }
+  },
+  computed: {
+    icon(): Icon {
+      switch (this.error.statusCode) {
+        case 404:
+          return { icon: 'mdi-alert', color: 'warning' };
+
+        default:
+          return { icon: 'mdi-alert-circle', color: 'error' };
+      }
+    }
+  }
+});
+</script>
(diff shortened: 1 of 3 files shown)


MrTimscampi:
Nuxt can use an `error.vue` file in the layouts folder for this, which would be cleaner: https://nuxtjs.org/guide/views/#error-page.

Nuxt has an example here: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/components/nuxt-error.vue

components/NotFound.vue
```diff
@@ -0,0 +1,20 @@
+<template>
```
Camc314, line 1: > Warning: Though this file is placed in the layouts folder, it should be treated as a page.

From [here](https://nuxtjs.org/guide/views/#error-page)

Move this to layouts?

ThibaultNocchi: No worry, I just finished it to use their native error call ;)

ThibaultNocchi:
So I used their error calling method (see example [here](https://github.com/jellyfin/jellyfin-vue/pull/41/files#diff-81a9fca19fc3e653641ed4ee02718ef2R53)) and you can pass a status code and message. Our custom component associates an icon and error with a status code, with already one for a 404 and one for default.

What do you think?

layouts/error.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <div>
+    <v-row justify="center" class="mt-16 mb-4">
+      <v-icon :color="icon.color" size="72">{{ icon.icon }}</v-icon>
+    </v-row>
+    <v-row justify="center">
+      <h1 class="text-h4">{{ error.message }}</h1>
```
Camc314, line 7: Looks good.
I would make this slightly smaller though

ThibaultNocchi: Switching to h5 is really smaller though, I'll look into other size classes

ThibaultNocchi: Oh maybe you're talking about the icon?

ThibaultNocchi: Screenshot_2020-09-05 jellyfin-vue(1)
Screenshot_2020-09-05 jellyfin-vue

MrTimscampi: Looks good like this, I think :+1:

Camc314: Actually yeah h4 is too small, h5 looks good to me

Camc314, line 7:
```suggestion
      <h1 class="text-h4 text-center">{{ error.message }}</h1>
```

just centers text on small screens.

ThibaultNocchi: Nice catch, done

Camc314 (approved):

PR #53, opened 2020-09-06
Design improvements

ThibaultNocchi:
* Removing one of the dividers in the nav drawer when no library is in the list, to avoid having two dividers displayed
* Hidden the button which enabled / disabled the mini variant of navigation drawer. All the code is still there, only the button is commented in HTML
* Hiding the drawer by default, better for mobile usage in order to avoid having to close it almost every time
* Removed the flex usage in the _viewId page. We should use rows and cols to make a responsive grid.
* Removed the .card and .cardImage classes. First one statically applied a size to the cards, which at least shouldn't be set to the component. Second one didn't have a definition anywhere else.

Feel free to comment on anything. :)

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -23,7 +23,7 @@
           </v-list-item-content>
         </v-list-item>
       </v-list>
-      <v-divider></v-divider>
+      <v-divider v-if="libraries.length > 0"></v-divider>
       <v-list>
         <v-list-item
           v-for="(library, i) in libraries"
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -1,12 +1,6 @@
 <template>
   <v-app dark>
-    <v-navigation-drawer
-      v-model="drawer"
-      :mini-variant="miniVariant"
-      :clipped="clipped"
-      fixed
-      app
-    >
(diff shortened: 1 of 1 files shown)


MrTimscampi:
> Hidden the button which enabled / disabled the mini variant of navigation drawer. All the code is still there, only the button is commented in HTML

We can probably remove this completely. It was part of the boilerplate and can be re-added very easily if we with to.

> Hiding the drawer by default, better for mobile usage in order to avoid having to close it almost every time

We can use `this.$vuetify.breakpoint.mobile` to close it automatically if the window is too small.

> Removed the flex usage in the _viewId page. We should use rows and cols to make a responsive grid.

This may conflict with some changes from #45, where I messed with the layout a bit to use columns.

> Removed the .card and .cardImage classes. First one statically applied a size to the cards, which at least shouldn't be set to the component. Second one didn't have a definition anywhere else.

Already taken care of in #45 :smiley:

ThibaultNocchi:
Ok, on the two first points I'll apply your suggestions. On the third I'll try to look into it. And the fourth well that's great haha.

ThibaultNocchi:
Done, I've also removed the commit about the views you've redone.

MrTimscampi (approved):

PR #54, opened 2020-09-06
feat(itemid.vue): responsive columns size

ThibaultNocchi:
Those are subjective, but I think they work well on the minimum size of each screen size step.

### DIFF
--- a/pages/library/_viewId.vue
+++ b/pages/library/_viewId.vue
@@ -8,7 +8,7 @@
       </v-col>
     </v-row>
     <v-row>
-      <v-col v-for="item in items" :key="item.Id" cols="2">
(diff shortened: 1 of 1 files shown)


pages/library/_viewId.vue
```diff
@@ -49,6 +49,9 @@ export default Vue.extend({
         });
 
         this.items = itemsResponse.data.Items || [];
+        for (let i = 0; i < 11; ++i) {
+          this.items.push(this.items[0]);
+        }
```
MrTimscampi, line 54: ?

ThibaultNocchi: Oops my bad, I only got one item in my library so I practiced some multiplications

ThibaultNocchi: Done

MrTimscampi (approved):

PR #61, opened 2020-09-07
German translation

cromefire:
Nothing fancy just a translation


nielsvanvelzen:
https://translate.jellyfin.org/projects/jellyfin/jellyfin-vue/de/

cromefire:
Is it also added to weblate?

cromefire:
Apparently, but the site is just really confusing

cromefire:
But it still has to be added to the nuxt config, which probably doesn't work automatically

nielsvanvelzen:
@MrTimscampi will do that soon for all new languages

PR #75, opened 2020-09-12
Add Dockerfile

Maxr1998:


nvllsvm (approved):

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN yarn install
+
+# Copy resources
+ADD . .
+
+# Build static site
+RUN yarn generate
+
+# Deploy to nginx
+FROM nginx:latest
```
nvllsvm, line 22: nit - `nginx:alpine` is roughly 5x smaller

Maxr1998: Oops yes, gonna fix. Shouldn't have copy-pasted there, oof.

MrTimscampi (approved):
I'm definitely not a Docker guru (I've only written a few docker-compose files :p ) but LGTM.

Shadowghost:
Shouldn't we use `COPY` instead of `ADD` to comply with [best practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy)?

PR #179, opened 2020-10-09
feat(ci): add docker image pipeline

Shadowghost:
What this PR should accomplish:
* Build unstable image after every commit to master
* Build stable image whenever we have a versioned git tag
* Images are built for amd64, armhf and arm64 architectures
* Push images to `jellyfin/jellyfin-vue` on docker hub

It's my first time doing azure ci (code originates from the jellyfin server ci configs with some changes by me) so this may not work as intended - comments are welcome.

An optimized approach could be to create the static files in their own job and not within the Dockerfile and to just copy them over into the different base images in the Dockerfile, but this would prevent users from easily building the image by building the Dockerfile.

### DIFF
--- a/.ci/azure-pipelines-build.yml
+++ b/.ci/azure-pipelines-build.yml
@@ -24,3 +24,62 @@ jobs:
 
       - script: 'yarn build'
         displayName: 'Build'
+
+  - job: BuildDocker
+    displayName: 'Build Docker'
+
+    strategy:
+      matrix:
+        amd64:
+          BuildConfiguration: amd64
+        arm64:
+          BuildConfiguration: arm64
+        armhf:
+          BuildConfiguration: armhf
+
+    pool:
+      vmImage: 'ubuntu-latest'
+
+    variables:
+      - name: Version
+        value: 0.0.0
+
+    steps:
+      - script: echo "##vso[task.setvariable variable=Version]$( awk -F '/' '{ print $NF }' <<<'$(Build.SourceBranch)' | sed 's/^v//' )"
+        displayName: Set release version (stable)
+        condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/v')
+
+    steps:
+      - task: Docker@2
+        displayName: 'Push Unstable Image'
+        condition: startsWith(variables['Build.SourceBranch'], 'refs/heads/master')
+        inputs:
+          repository: 'jellyfin/jellyfin-vue'
+          command: buildAndPush
+          buildContext: '.'
+          Dockerfile: 'Dockerfile'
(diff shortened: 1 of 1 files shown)


.ci/azure-pipelines-build.yml
```diff
@@ -24,3 +24,53 @@ jobs:
...
+
+    pool:
+      vmImage: 'ubuntu-latest'
+
+    variables:
+      - name: Version
+        value: 0.0.0
+
+    steps:
+      - script: echo "##vso[task.setvariable variable=Version]$( awk -F '/' '{ print $NF }' <<<'$(Build.SourceBranch)' | sed 's/^v//' )"
```
EraYaN, line 48: Would something like this work? `$[replace(variables['Build.SourceBranch'],'refs/tags/v','')]`

EraYaN (approved):

.ci/azure-pipelines-build.yml
Shadowghost: I have no idea, I took this line from [jellyfin-server](https://github.com/jellyfin/jellyfin/blob/2375c35c4a966a469dd97f33deb57b342498b905/.ci/azure-pipelines-package.yml#L119).

h1dden-da3m0n: Just because I touched the Java API CI today this is how that CI gets the version (note that would still result in vX.Y.Z)

```yaml
      - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
        displayName: 'Set Version Variable'
```

h1dden-da3m0n: > Just because I touched the Java API CI today this is how that CI gets the version (note that would still result in vX.Y.Z)
> 
> ```yaml
> - script: 'echo "##vso[task.setvariable variable=JELLYFIN_VERSION]$(git describe --tags)"'
>   displayName: 'Set Version Variable'
> ```
> 
> /disclaimer: I did not touch this piece of the CI

okay due to this comment I may have found the reason for some artifact upload errors in the Java API ci. (**POSSIBLY**)
So pleas disregard my earlier comment

EraYaN: I mean it would just mean you don't need the extra step, you can assign the variable at the top of the file using that expression if required.

Shadowghost: Should be resolved with the latest commit if I understood you correctly.

EraYaN: Yes looks good.

EraYaN (approved):

PR #180, opened 2020-10-09
Refactor the API plugin

MrTimscampi:
Merges all the API plugins into one, and adds all of the exposed APIs into it.

This makes the tree a bit cleaner and provides one way to access all the APIs, which is a bit simpler.


Camc314 (approved):
LGTM

PR #184, opened 2020-10-11
feat(details): redesign item details

MrTimscampi:
I accidentally ended up redesigning the entire UI while doing this.

**Movies**
Screenshot_2020-10-22 jellyfin-vue

**Shows**
Screenshot_2020-10-22 jellyfin-vue(1)

**Person**
Screenshot_2020-10-22 jellyfin-vue(2)


neopc10:
Tangentially related here but for cast/crew, can we get multiple images available for each person?
For reference:
image


1983 hours and 11 minutes?

MrTimscampi:
> Tangentially related here but for cast/crew, can we get multiple images available for each person?

Unfortunately, that's completely different than what is being done here, and it'd have to be done on the server, not on a client. We only get the images sent to us by the server.

> 1983 hours and 11 minutes?

There's a bug in the current implementation of the runtime calculation in the client. It'll be fixed as part of this.

hamburglar2160:
Looks good.

What is the "16" on the City Hunter screenshot? Is that a rating? Are ratings for TV shows even relevant or necessary? I certainly see how they're relevant (but maybe not necessary) for movies.

Stylistically, I feel like we could do away with the "seconds" count for movie length or for anything above 60 minutes. It kind of feels like fluff? If it read "1h22m10s" (I think that'd look silly on non-mobile) then it'd be a different matter, but in its current state it's fluff?

Should genres get lumped in with the above as well? They appear to be missing for the TV show details.

To my eye, the padding between each of the cast/crew entries seems excessive. Is it any better at 20-40% less?

Are those collapsible views I see for "you may also like?" -- nice!

MrTimscampi:
> What is the "16" on the City Hunter screenshot? Is that a rating? Are ratings for TV shows even relevant or necessary? I certainly see how they're relevant (but maybe not necessary) for movies.

It the rating. It shows up on the current client as well.

It's important for people with children, who would want to know what something is rated as before watching something.

> Stylistically, I feel like we could do away with the "seconds" count for movie length or for anything above 60 minutes. It kind of feels like fluff? If it read "1h22m10s" (I think that'd look silly on non-mobile) then it'd be a different matter, but in its current state it's fluff?

It's an early version. It's planned to remove it. I'm just trying to not get too hung up on small details in order to get this out of the door quicker.

> Should genres get lumped in with the above as well? They appear to be missing for the TV show details.

They're not. The screenshots don't fully represent the current state of the branch. I can't update them every time I change something.

> To my eye, the padding between each of the cast/crew entries seems generous. Is it any better at 20-40% less?

That's the way they look by default. The "dense" mode pretty much just makes text smaller. There's no real point to making them less padded though. It'll squish info together too much. This way looks good, imo.

> Are those collapsible views I see for "you may also like?" -- nice!

They're not.

The "You may also like" section has two modes: vertical and horizontal.

Showing the seasons makes more sense than related items on TV shows, so the layout is adjusted to reflect that. On movies, on the other hand, there's no season to show, so it makes sense to help guide the user towards other relevant content.

mixins/timeUtils.ts
```diff
@@ -19,7 +19,7 @@ declare module '@nuxt/types' {
 
 declare module 'vue/types/vue' {
   interface Vue {
-    ticksToMs: (ticks: number) => number;
+    ticksToMs: (ticks: number | null | undefined) => number;
```
ferferga, line 22: The behaviour of this whole file can't be accomplished with date fns? What does ticks equals to? Never understood what it meant in web.

components/MediaInfo.vue
```diff
@@ -0,0 +1,95 @@
...
+        return this.$dateFns.format(Date.now() + seconds, 'p');
+      }
+    }
+  },
+  methods: {
+    getEndsAtTime(ticks: number): string {
+      const ms = this.ticksToMs(ticks);
+      const endTimeLong = new Date(Date.now() + ms);
+      // TODO: Respect user locale when rendering time
+      const endTimeShort = endTimeLong.toLocaleString('en-US', {
```
ferferga, line 67: Not tested:
```suggestion
      const endTimeShort = endTimeLong.toLocaleString($i18n.getLocale(), {
```

components/MediaInfo.vue
```diff
@@ -0,0 +1,95 @@
...
+    runtimeValue: {
+      get() {
+        const seconds = this.ticksToMs(this.item.RunTimeTicks);
+        return formatDuration(intervalToDuration({ start: 0, end: seconds }));
+      }
+    },
+    endsAtValue: {
+      get() {
+        const seconds = this.ticksToMs(this.item.RunTimeTicks);
+        return this.$dateFns.format(Date.now() + seconds, 'p');
```
ferferga, line 58: See my other comment about this mixin

ferferga (commented):
Does search work or it's just a placeholder? What's left for this to be ready?

MrTimscampi:
> Does search work or it's just a placeholder? What's left for this to be ready?

Placeholder, for now.

What's left is mainly cleanup. I'm still not sure if I'll do artists pages and such in another PR or not.

mixins/timeUtils.ts
MrTimscampi: Ticks is something from C#. It's basically the unit of representation of time over there.

The file is meant to convert ticks sent by the server (There are plans to change this in 11.0, I believe) to milliseconds in one central place to avoid all the undocumented `thing * 1000` and `thing / 1000` all over jf-web.

components/MediaInfo.vue
MrTimscampi: Same answer :) The mixin is to convert C# concepts into JS concepts.

ferferga:
> > Does search work or it's just a placeholder? What's left for this to be ready?
> 
> Placeholder, for now.
> 
> What's left is mainly cleanup. I'm still not sure if I'll do artists pages and such in another PR or not.

I would leave it for another PR, so we can get this in quicker

ferferga (approved):

PR #197, opened 2020-10-15
Properly scrollable locale switcher

ferferga:
Quick fix for an small quirk. Once I go deeper, I can do the same to the rest of the menus if needed.


components/LocaleSwitcher.vue
```diff
@@ -5,7 +5,8 @@
         <v-icon>mdi-web</v-icon>
       </v-btn>
     </template>
-    <v-list>
+    <v-list style="max-height: 90vh"
```
MrTimscampi, line 8: I'd rather have that in a scoped style block at the end of the file, if possible.

90vh is also probably too high. Something in EM would be better, for consistency.

MrTimscampi (changes requested):

MrTimscampi (dismissed):

MrTimscampi:
@ferferga 

```
/home/vsts/work/1/s/components/LocaleSwitcher.vue
  24:9  error  Insert `⏎`  prettier/prettier
```

MrTimscampi (approved):

PR #223, opened 2020-10-22
feat(progressbar): use nuxt loading and create custom progressbar component

ferferga:
Adds a progressbar to the top while switching pages like the one in YouTube.

progress

**Marked as a WIP because the custom progress bar component doesn't react to the methods called by the Nuxt loading module. There is still cleanup to be made**


ferferga:
### Now ready to be reviewed

**Changes:**

- Fully functional
- Moved the progressbar to the top of the page, as most people was gearing towards it in Matrix.

**To do**

- Implement the progress bar in all the routes and not only in library view. I added it to the default route but if stuff is cached it doesn't appear. I need to investigate how Nuxt handles route loading to implement that. This can be reviewed in the meantime (or merged, no problem in doing it in another PR for me as I might take a while).

new_progress

h1dden-da3m0n:
I have one question not related to this PR, but why is there an additional user icon on the top right if there is already the user in the menu to the left?

Sorry for this unrelated comment, but your latest showcase mad me notice :thinking:

Camc314:
> I have one question not related to this PR, but why is there an additional user icon on the top right if there is already the user in the menu to the left?

I think the icon to the left (kind of like a globe) is for changing languages

The other icon is for logging out

h1dden-da3m0n:
> I think the icon to the left (kind of like a globe) is for changing languages
> 
> The other icon is for logging out

Thanks for the comment, but I am aware of what the two icons do on the right.
However, why is there a "user section" in the menu on the left side of the screen too?

ferferga:
@h1dden-da3m0n Introduced in #184 

I think @MrTimscampi's plans are to leave only the right one for mobile interface and the right for desktop layout. The current interface is still pretty much alpha (resizing the window removes sidebar and you can't navigate, for example) so those quirks will be changed in the future 😁.

Also, I see no point in keeping the language switcher (great in the login screen, but you aren't changing languages all the day...) there, it makes more sense in settings. Might be useful for bilingual countries but anyway, afaik people usually stick to one.

h1dden-da3m0n:
Hey ferferga, thank you for the clarifications! I am aware that this UI is still very much in alpha :wink:
Anyway, I was just curious why there where two user "widgets" all of the sudden and misused your PR to fire the question, sorry for the later again.

Anyway, keep up the great work :+1:

KristupasSavickas:
IMO it would look better ath the very top of the screen - like it's done on Youtube. Right now it looks a bit weird being in between the navigation bar and the side bar.

ferferga:
@KristupasSavickas Latest commits have it a the very top, see the GIF.

components/loading/LoadingBar.vue
```diff
@@ -0,0 +1,59 @@
...
+      this.barColor = 'primary';
+      this.barValue = 0;
+      // Ensure that the 'slide up' animation appears after loading finishes. Without this
+      // the animation could be skipped due to DOM blocking.
+      timeout = window.setTimeout(() => (this.loading = false), 50);
+    },
+    fail() {
+      this.barColor = 'red';
+      this.indeterminate = false;
+      this.barValue = 30;
```
MrTimscampi, line 47: Why 30?

Seems like it'd be a perfect use of "indeterminate", or even put it at 100 so it covers the entire width?.

layouts/default.vue
```diff
@@ -147,14 +147,19 @@ export default Vue.extend({
...
+  mounted() {
+    this.$nuxt.$loading.finish();
   }
 });
 </script>
 
 <style lang="scss" scoped>
 .v-application {
   background-color: var(--v-background-base) !important;
+  margin-top: -4px;
```
MrTimscampi, line 162: Won't this cutoff part of the header app when the loading bar isn't being shown?

components/loading/LoadingBar.vue
```diff
@@ -0,0 +1,59 @@
...
+      :color="barColor"
+      :value="barValue"
+    ></v-progress-linear>
+  </v-app>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+let timeout: number;
```
MrTimscampi, line 18: This could probably use a better name and should, if possible, be part of the component (But I'm not sure if this component is destroyed and re-mounted or just hidden).

layouts/default.vue
ferferga: @mrtimscampi No, because I keep the component always at top (but hidden) so page doesn't jump when the progressbar disappears. If I don't do this, it's plain horrible.

components/loading/LoadingBar.vue
ferferga: @mrtimscampi Indeterminate doesn't make sense imo, as the progressbar will keep "flowing" in the event of an error, which doesn't make sense imo, as when an error loading happens it usually means that the loading process was halted, thus, no progress is happening. 

30 fills a quarter of the bar while being still visible with the sidebar open.

components/loading/LoadingBar.vue
ferferga: @MrTimscampi Is hidden. I thought setting this here "globally" made more sense, as it's not a variable that you will manipulate externally (from the Vue.js debugger or other components), you will only use it in the progress bar component itself. It isn't relevant for anything else but the progress bar either.

Agree with you about the name though, I'll make it clearer.

MrTimscampi: You generally want to keep variables either in components, or in a Vuex store :smiley: 

Defining variables just outside of a component like this, to be then used inside of it, is pretty bad practice :wink:

components/loading/LoadingBar.vue
MrTimscampi: After testing it, I think the sidebar shouldn't be hiding the loading bar (in desktop mode, at least).

The sidebar isn't closable on desktop (it was before, but it's not intended to be in the current design), so it looks a bit weird there. And on mobile, you'll never be able to navigate while the navigation bar is open (It should close when you click on a link).

As for the error bar, I think filling it completely would be better.  
I expect that version to barely be shown, since we should handle errors in a more verbose way. We have an error page callable from everywhere, which should be used when API calls fail. For server disconnections, I'm planning to have an overlay open with a re-connection attempt tracker when the websocket goes down. So it'll only be shown in some very rare cases where we don't handle things properly, but then 30% of the bar filled doesn't make sense. It should be as visible as possible, hence the 100% value. At least imo :smiley:

components/loading/LoadingBar.vue
MrTimscampi, line 18: Looking at it again, I think this should be a `data`.

Being manipulated externally or in other components has no bearing on where it will live, imo. If it's part of what the component needs to work, it should be inside of the components.

components/loading/LoadingBar.vue
```diff
@@ -0,0 +1,59 @@
...
+export default Vue.extend({
+  data() {
+    return {
+      loading: false,
+      barColor: 'primary',
+      indeterminate: true,
```
MrTimscampi, line 24: Regarding this, I looked a bit into the behavior of the normal Nuxt bar and it seems better to me.

For reference:
> Unfortunately, it is not possible for the Loading component to know in advance how long loading a new page will take. Therefore, it is not possible to accurately animate the progress bar to 100% of the loading time.
> Nuxt's loading component partially solves this by letting you set the duration, this should be set to an estimate of how long the loading process will take. Unless you use a custom loading component, the progress bar will always move from 0% to 100% in duration time (regardless of actual progression). When the loading takes longer than duration time, the progress bar will stay at 100% until the loading finishes.

So instead of being indeterminate, we would go from 0 to 100 in the time set in the config, then stay at 100% until loaded.

For this to work, we'll need to disable the nuxt-axios integration into the loading components, by setting `progress` to `false` in the `axios` key of `nuxt.config.ts` (Since that's what is taking over the loading bar).

components/loading/LoadingBar.vue
```diff
@@ -0,0 +1,59 @@
...
+      barValue: 0
+    };
+  },
+  methods: {
+    start() {
+      window.clearTimeout(timeout);
```
MrTimscampi, line 30: This will generate an error if `timeout` isn't initialized, I believe.

components/loading/LoadingBar.vue
```diff
@@ -0,0 +1,59 @@
...
+      this.barColor = 'primary';
+      this.barValue = 0;
+    },
+    finish() {
+      this.indeterminate = true;
+      this.barColor = 'primary';
+      this.barValue = 0;
+      // Ensure that the 'slide up' animation appears after loading finishes. Without this
+      // the animation could be skipped due to DOM blocking.
+      timeout = window.setTimeout(() => (this.loading = false), 50);
```
MrTimscampi, line 42: In addition to hiding the bar, the timeout method should also reset the data to its original state, imo. This allows you to avoid resetting everything in every state.

Moving it to a method would also probably be better.

components/loading/LoadingBar.vue
MrTimscampi, line 47:
```suggestion
      this.barValue = 100;
```

layouts/default.vue
```diff
@@ -147,14 +147,19 @@ export default Vue.extend({
     }
   },
   beforeMount() {
+    this.$nuxt.$loading.start();
```
MrTimscampi, line 150: With the change proposed above (ie going back to Nuxt's original behavior), this is not needed.

pages/library/_viewId.vue
```diff
@@ -103,6 +103,9 @@ export default Vue.extend({
     }
   },
   async beforeMount() {
+    this.$nextTick(() => {
```
MrTimscampi, line 106: Same with this, we should adopt nuxt's original estimation system, which avoids us having to manually set loading on every single route (which is going to get annoying quickly, and makes maintainability worse)

MrTimscampi (changes requested):

components/loading/LoadingBar.vue
ferferga: @MrTimscampi No changes, as soon as I remove all the ``this.$nuxt.$loading.start();`` functions, progress bar doesn't appear.

The [``increase``](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-loading#using-a-custom-loading-component) method doesn't seem to work as well

MrTimscampi (approved):

Issue #145, opened 2020-10-02
Add filters to the library view


MrTimscampi:

AlfHou:
I'll take a look at this

AlfHou:
I'm looking at the network calls the jellyfin-web client makes when selecting filters. My understanding is that the filtering is done serverside. Is that correct?

PR #230, opened 2020-10-24
Feat/library filter

AlfHou:
This pull request implementes #145. 

I have never used typescript before, so please let me know if I've done something unconventional.
image


components/FilterButton.vue
```diff
@@ -0,0 +1,193 @@
...
+        <v-expansion-panel-content>
+          <v-form v-for="(filter, index) in item.items" :key="index">
+            <v-checkbox
+              v-model="filter.selected"
+              class="my-0"
+              :label="filter.label"
+              :value="filter.value"
+              :true-value="true"
+              :false-value="false"
+              @change="$emit('input', filters)"
```
MrTimscampi, line 20:
```suggestion
              @change="$emit('change', filters)"
```
A change event makes more sense, since you're changing the filters, I think.

pages/library/_viewId.vue
```diff
@@ -55,6 +64,12 @@ import Vue from 'vue';
 import { chunk } from 'lodash';
 import { BaseItemDto } from '~/api/api';
 
+interface FilterItem {
```
MrTimscampi, line 67: You should reuse the type from other component. Avoid duplicating types, it leads to errors. Feel free to make a `types` folder in the root of the repository to store them, if needed.

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
...
+            options.includeItemTypes = 'Movie';
+          } else if (collectionInfo.data.Items[0].CollectionType === 'books') {
+            options.includeItemTypes = 'Book';
+          }
+
+          const filterString = this.makeFilterString(
+            this.allFilters.filters.items,
+            ','
+          );
+          options.filters = filterString;
```
MrTimscampi, line 237: I don't get why you're fetching these again. You should pass data around instead of fetching it multiple times (You're already passing all of this in the change event for the filter menu, for example, which is bound to this. So one of your parameters would already contain the list).

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
...
+
+          for (let i = 0; i < this.allFilters.features.items.length; i++) {
```
MrTimscampi, line 239: Use a modern type of `for` to do this:
```suggestion
          for (const feature of this.allFilters.features.items) {
```

They're generally more readable.

MrTimscampi: You can also probably use `.filter()` ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)) instead of a loop, to make it even nicer.

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
     }
   },
   methods: {
+    async filterMedia() {
```
MrTimscampi, line 201: This mostly looks like stuff that could be done inside of the filter menu itself.

It would generate the filter string internally and pass it through the `change` event.

This would have the benefit of avoiding the need to redo this entire method on every page that wants to use the filter menu.

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
...
+          options.genres = genreString;
+
+          const ratingString = this.makeFilterString(
+            this.allFilters.officialRatings.items,
+            '|'
+          );
+          options.officialRatings = ratingString;
+
+          let videoTypeString = '';
+          for (let i = 0; i < this.allFilters.videoTypes.items.length; i++) {
```
MrTimscampi, line 261: Prefer `for...of` loops in most cases.

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
...
+              videoTypeString += videoType.value;
+            }
+          }
+          options.videoTypes = videoTypeString;
+
+          const yearString = this.makeFilterString(
+            this.allFilters.years.items,
+            ','
+          );
+          options.years = yearString;
```
MrTimscampi, line 289:
```suggestion
          options.years = this.makeFilterString(
            this.allFilters.years.items,
            ','
          );
```

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
...
+          userId: this.$auth.user.Id,
+          ids: this.$route.params.viewId
+        });
+
+        if (
+          collectionInfo.data.Items &&
+          (collectionInfo.data.Items[0].Type === 'CollectionFolder' ||
+            collectionInfo.data.Items[0].Type === 'Folder')
+        ) {
+          const options: any = {
```
MrTimscampi, line 214:
```suggestion
          const options: ItemsApiGetItemsRequest = {
```

Avoid using `any` when types are available.

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
...
+        }
+      } catch (error) {
+        // Can't get given library ID
+        this.$nuxt.error({
+          statusCode: 404,
+          message: this.$t('libraryNotFound') as string
+        });
+      }
+    },
+    makeFilterString(filterList: FilterItem[], seperator: string) {
```
MrTimscampi, line 307: Similarly, this should live in the filter menu component, since it's very much tied to that.

pages/library/_viewId.vue
```diff
@@ -157,6 +198,127 @@ export default Vue.extend({
...
+      let filterString = '';
+      for (let i = 0; i < filterList.length; i++) {
```
MrTimscampi, line 309: `for...of` again :smiley:

MrTimscampi (changes requested):
UI-wise, I think it would be beneficial to have the `v-expansion-panel-content` have a maximum height.

These lists can get very long, which makes it not ideal to browse. Making them limited in height and scrollable would be good.

MrTimscampi:
You also seem to have some weirdness in your history, with the merge commit. Could you rebase and clean it up, please?

components/FilterButton.vue
```diff
@@ -0,0 +1,193 @@
...
+      type: Object,
+      required: true
+    }
+  },
+  data() {
+    return {
+      selectedFilters: this.value,
+      filters: {
+        filters: {
+          header: this.$t('filters'),
```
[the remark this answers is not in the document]
AlfHou: What part would you like to be status? The "filters" on line 50 or 49? Also if you want the line 50 to be changed, I assume the header should still be 'filters'?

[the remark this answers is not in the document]
MrTimscampi: 50 and the accompanying header :)

pages/library/_viewId.vue
AlfHou: When I chang eit to `ItemsApiGetItemsRequest` I get errors regarding `Cannot assign to 'filters' because it is a read-only property`.

AlfHou:
> You also seem to have some weirdness in your history, with the merge commit. Could you rebase and clean it up, please?

What weirdness are you referring to? Do you want the merge commit gone?

ferferga:
@AlfHou Yeah, I think 4c7cfd2 is the one that @MrTimscampi was referring to. 505d62b and ffbef1e are fine, unless we want to change the way we do our history from the old web client.

MrTimscampi:
@ferferga is correct, I was referring to 4c7cfd2 :)

Other merge commits are alright, but I'd like to avoid duplicate commits (And that one is already in the history, with a different hash).

pages/library/_viewId.vue
AlfHou: Would it be okay if I changed some of the `ItemsApiGetItemsRequest` properties from readonly to not readonly?

AlfHou:
All the changes requested should be fixed now

MrTimscampi (dismissed):
LGTM for now :)

Sorry for the late re-review and thanks for the contribution !

MrTimscampi (approved):

PR #264, opened 2020-11-09
Add music-centric item details pages

MrTimscampi:
Depends on #263

## Artist
Screenshot_2020-11-09 jellyfin-vue(3)

### Single disc
Screenshot_2020-11-09 jellyfin-vue(1)

### Multi-disc
Screenshot_2020-11-09 jellyfin-vue(2)

## Genres
Screenshot_2020-11-09 jellyfin-vue


locales/en-US.json
```diff
@@ -31,8 +35,10 @@
   "serverVersionTooLow": "Server version needs to be 10.7.0 or higher",
   "settings": "Settings",
   "shows": "Shows",
+  "shuffleAll": "Shuffe all",
```
neopc10, line 38: "shuffe"

neopc10:
i haven't checked, but have you approached music playlists yet? 
It'd be very helpful to have the Album and Album Artist visible for each song. I'd really like to see a (toggleable?) limit on the number of genres displayed on the playlists. Something along the lines of hiding genres that don't appear in over 10-20% of the playlist, perhaps, for those with over 50-100 tracks.
image

h1dden-da3m0n:
Love the design, very Spotify like. Great work :+1:

locales/en-US.json
MrTimscampi: Good catch :yum:

MrTimscampi:
> i haven't checked, but have you approached music playlists yet?

Not at all, really. It's like pretty far down the line, once we have proper support for playing music and a playback manager with support for queuing items (Currently, it's a very rough video player that also plays audio, and can only play one thing).

> Might we be given the option to toggle off the "heart" button as well?

It's planned. I'm mostly going over these screens trying to get everything to look right, then we'll hook them up later on.

> Love the design, very Spotify like. Great work +1

Thanks! That's very much the inspiration for this lol There isn't much you can do with music (everything has been figured out over 25 years of music players) and I figured a familiar UI would be the best for onboarding users.

KristupasSavickas:
Great work!

A few notes/suggestions:
1. Details should be limited to sane length, possibly with a button to "Read more" which would expand everything.
Screenshot_20201110_103303

2. Last.fm likes to add HTML tags to details, not sure if good practice, but they're not supported here:
Screenshot_20201110_103312

3. It would be nice if hovering over a song would display a play symbol (i.e. a triangle), possibly instead of the track number:
Screenshot_20201110_103337

4. Song's with multiple artists don't include the artist:
Screenshot_20201110_103644

My suggestion to solve 3 and 4:
Artist list could be next to/below the song name (maybe in a smaller font/slightly grayed out). Each artist name could be clickable to navigate to the artist page. The song would only start playing by clicking the play symbol suggested in (3), not by clicking anywhere on the row.

ferferga:
@KristupasSavickas I agree with you on everything except the play part only on the button.

That's not how Spotify works and we shouldn't artificially force users in my opinion. Spotify lets you do a single click on the play button while hovering but also a double quick everywhere the item is highlighted.

KristupasSavickas:
Regarding the play button, I suggested it because clickable artist names could be problematic, i.e. you'd have to click exactly on the name, otherwise you might start playback.

MrTimscampi:
> I suggested it because clickable artist names could be problematic otherwise

We can stop event propagation to deal with that.

With 155cd4d, the play action is now bound to `dblclick` on the entire row, along with a hover-only play button which is enabled with a single click.

MrTimscampi:
FYI, a7c40b9 adds an HTML sanitation mixin, using [DOMPurify](https://github.com/cure53/DOMPurify), as well as support for multi-line overviews and HTML-tags in overviews.

I'm moving this PR back to draft, this requires us to re-thing the position of overviews a bit in every details page, as well as provide a "read more" thing. See the screenshot below for an example of the issue.

image

MrTimscampi:
Small update for the overview text redesign.

Screenshot_2020-11-11 jellyfin-vue

**(Note: the black part in the screenshot below is due to the way the background and navigation bar on the left are set to follow the scroll. The move along with the page, hence the black background in this screenshot)**
Screenshot_2020-11-11 jellyfin-vue

Screenshot_2020-11-11 jellyfin-vue(1)

KristupasSavickas:
The new changes looks really good, thanks!
A couple of things I noticed:
- HTML tags in overview still do not work:
Screenshot_20201112_105726
- The "About" section in artist page solves issues with long overviews, but it's still an issue in album pages:
Screenshot_20201112_105825

MrTimscampi:
> HTML tags in overview still do not work

That's weird, since I'm pretty much outputing exactly what the server sends me, with `v-html`. Adding repositories is currently broken on master, so I'm afraid this will have to wait until someone working on the server fixes it (I have notified the rest of the team internally about the regression).

> The "About" section in artist page solves issues with long overviews, but it's still an issue in album pages

Indeed. I was honestly considering just not showing them, because I didn't even think albums would have an overview (It's kind of useless, really). But I guess we can use the same tab system as the artists, though it looks pretty weird there...

KristupasSavickas:
>Indeed. I was honestly considering just not showing them, because I didn't even think albums would have an overview (It's kind of useless, really). But I guess we can use the same tab system as the artists, though it looks pretty weird there...

components/RelatedItems.vue
```diff
@@ -119,19 +127,33 @@ export default Vue.extend({
...
-      console.error('Unable to get related items:', error);
+        this.relatedItems = response.data.Items;
+      }
+
+      this.loading = false;
+    },
+    watch: {
+      item() {
+        this.refreshItems();
+      }
```
ferferga, line 156:
```suggestion
        this.relatedItems = response.data.Items;
      }

      this.loading = false;
    },
}
watch: {
	item() {
		this.refreshItems();
	}
```

pages/item/_itemId/index.vue
```diff
@@ -54,12 +71,29 @@
...
+                "
               >
                 <v-col cols="2" class="d-flex align-center pa-0">
                   <label class="text--secondary">Genres</label>
                 </v-col>
-                <v-col cols="7">{{ item.Genres.join(', ') }}</v-col>
+                <v-col cols="7">
+                  <v-chip
+                    v-for="genre in item.GenreItems"
+                    :key="genre"
```
ferferga, line 87:
```suggestion
                    v-for="(genre, index) in item.GenreItems"
                    :key="index"
```

You treat ``genre`` as an string, but it isn't, so Vue outputs an error in console

ferferga (changes requested):

pages/item/_itemId/index.vue
MrTimscampi: Used genre.Id in order to account for items updating

KristupasSavickas:
>But I guess we can use the same tab system as the artists, though it looks pretty weird there...
My proposition would be to limit the number of lines shown in album view, with ability to see everything by pressing "Read more...", or something along the lines, which would open a modal with full overview.

pages/item/_itemId/index.vue
ferferga: @MrTimscampi Good point!

ferferga (dismissed):

ferferga:
@MrTimscampi This is good to go now as soon as merge conflicts are fixed! 👏🏻

MrTimscampi:
I disabled the stale thing for next time :eyes: I had forgotten I had that enabled, and it's annoying for everybody involved.

PR #265, opened 2020-11-09
Disable nuxt telemetry

ferferga:


MrTimscampi (commented):
Seems like you need to run Prettier on the config file (PS: If you use VSCode, install the Prettier extension and this should happen automatically on save, with the config we ship with the project).

ferferga:
@MrTimscampi Lint errors seems to be caused by other files. I already ran prettier and comment's tabulation are in line with the other ones 🤔.

MrTimscampi (approved):

MrTimscampi:
@ferferga Seems like the comment block had one too many spaces:
image

I fixed it myself, since I had the branch checked out to run the linter on it.

ferferga:
@MrTimscampi Oops!

Thank you very much! <3

PR #267, opened 2020-11-10
Rework and cleanup the library pages

### LINKED ISSUE #270: Typescript type error in FilterButton component
I get an error regarding the `filters` field in the `ItemsApiGetItemsRequest` used in the `FilterButton` component
```ERROR in components/FilterButton.vue:251:13                                                                                            
TS2322: Type 'string' is not assignable to type 'ItemFilter[] | undefined'.
    249 |             includeItemTypes: this.itemType,
    250 |             recursive: true,
  > 251 |             filters: statusString,
        |             ^^^^^^^
    252 |             genres: genreString,
    253 |             officialRatings: ratingString,
    254 |             videoTypes: videoTypeString,
```
it seems like the type was changed in commit 7585079a1479256916c2d58b0dfb1eee7b75d7bf

MrTimscampi:
* Completely rewrites the sort and filter buttons, by simplifying the way they send data back to the parent component. Also making a better use of Vuetify components and Vue features (models) to reduce the amount of code in the Filter button.
* Prepares the library page for moving data fetching/refreshes to Vuex, by decoupling the item refresh from the rest of the component.
* Adds a "view type" button, which allows the user to view different kinds of objects for different library types.
* Cleans up the design of the library page (in both desktop and mobile sizes)
* Moves the part of the library page to its own component named "Item Grid", which contains everything needed to render a grid of items in the client
* Adds some Card stuff for a few more item types related to the added views
* Properly truncate the text for cards
* Contains the appbar opacity changes from #264
* Fixes a typo on master that prevents the virtual scroller from working

Fixes #270 

image


neopc10:
I'm guessing that what is screenshotted is the core album view for the music library. If that's the case, the album artist would be more useful than the date. Down the line when the database will take musicbrainz' "originaldate" and not "date," that would be a great thing to feature here. 

I'd also reduce the fontsize of the subtitle. better yet, might that belong on a mouseover of the album art?

Has there been any thought of somehow shoehorning coverflow here?

Being able to "zoom in" or "zoom out" without resizing the window could be pretty cool here . . .

I'm not sure if this belongs here but a rather killer feature of Blackplayer on Android is this "hide small artists" and "hide small albums." That would be an absolutely wonderful thing to have available for Jellyfin users, especially for those with larger libraries that can't retrieve artist images or would rather not have 1-2 track "single" albums displayed here. 
Screenshot_20201109-153549

MrTimscampi:
> I'm guessing that what is screenshotted is the core album view for the music library. If that's the case, the album artist would be more useful than the date.

This PR is for cleaning up code and the filtering/sorting UI of libraries :)

> I'd also reduce the fontsize of the subtitle. better yet, might that belong on a mouseover of the album art?

Reducing the font size, I can see (It's 16px now, which is the Vuetify default). Mouseover would be bad UX though, since you can't find something quickly.

> Has there been any thought of somehow shoehorning coverflow here?

I'm not a fan of coverflow in general, and I think it doesn't have any place on a page whose purpose is to find stuff quickly. There can always be contributions by other folks down the line, when we have settings though. Keep in mind this client is super early and we working in passes, most of the time, doing the bare minimum and slowly adding to it.

> Being able to "zoom in" or "zoom out" without resizing the window could be pretty cool here . . .

There are eventual plans for a "card size" slider in that top bar. Overriding the browser's zoom is bad though, imo, as it reduces accessibility. So it'll never be bound to the mouse wheel.

> I'm not sure if this belongs here but a rather killer feature of Blackplayer on Android is this "hide small artists" and "hide small albums." That would be an absolutely wonderful thing to have available for Jellyfin users, especially for those with larger libraries that can't retrieve artist images or would rather not have 1-2 track "single" albums displayed here.

That would primarily be a server thing. Do keep in mind that, if your library is properly tagged, we have an "Artist" and "Album artist" field. The album artist one takes the main artist for an album, not the individual track ones.

KristupasSavickas:
I know this isn't final, but IMO it would look better to have the search bar and filter bar in a single row, as there's a bunch of unused space right now.

KristupasSavickas:
Also, the bar is a bit broken on narrow screens:
recording-2020-11-12_11 06 54

MrTimscampi:
> Also, the bar is a bit broken on narrow screens:

That's know, the Genres bar in the music PR is also broken on smaller screens. It needs some per-breakpoint margins, which I haven't gotten to yet since I focus on desktop for now (though I should)

components/ItemGrid.vue
```diff
@@ -0,0 +1,125 @@
...
+      type: Boolean,
+      required: true
+    }
+  },
+  computed: {
+    itemsChunks: {
+      get() {
+        let cardsPerLine = 8;
+
+        if (this.$vuetify.breakpoint.smAndDown) {
```
MrTimscampi, line 62: This seem to trigger a type error, but I'm not sure why (especially since it works properly).

Aside from this, this PR is ready to go.

components/FilterButton.vue
```diff
@@ -1,26 +1,196 @@
...
+    <template #activator="{ on, attrs }">
+      <v-btn
+        v-if="!$vuetify.breakpoint.smAndDown"
+        class="ma-2"
+        text
+        rounded
+        v-bind="attrs"
+        v-on="on"
+      >
+        Filter
```
Camc314, line 12:
```suggestion
        {{ $t('filter') }}

File: components/FilterButton.vue [hunk 3]
```diff
@@ -1,26 +1,196 @@
...
+                      <v-checkbox :input-value="active"></v-checkbox>
+                    </v-list-item-action>
+                  </template>
+                </v-list-item>
+              </template>
+            </v-list-item-group>
+          </v-list>
+        </v-expansion-panel-content>
+      </v-expansion-panel>
+      <v-expansion-panel>
```
Camc314, line 84:
```suggestion
      <v-expansion-panel v-if="genreFilters">

File: components/FilterButton.vue [hunk 4]
```diff
@@ -1,26 +1,196 @@
...
```
Camc314, line 110:
```suggestion
      <v-expansion-panel v-if="ratingFilters">

File: components/FilterButton.vue [hunk 5]
```diff
@@ -1,26 +1,196 @@
...
```
Camc314, line 168:
```suggestion
      <v-expansion-panel v-if="yearFilters">

File: components/ItemGrid.vue (hunk 1 above)
Comment (Camc314) [reply components/ItemGrid.vue:62]: Shall we add Typescript ignore comments so it can still build?

File: pages/library/_viewId.vue [hunk 6]
```diff
@@ -90,196 +64,256 @@ export default Vue.extend({
...
+          itemsResponse = (
+            await this.$api.studios.getStudios({
+              userId: this.$auth.user.Id,
+              parentId: this.$route.params.viewId
+            })
+          ).data;
+          break;
+        default:
+          itemsResponse = (
+            await this.$api.items.getItems({
```
Camc314, line 247: Put this in a ```try``` block?

pages/library/_viewId.vue
```diff
@@ -90,196 +64,256 @@ export default Vue.extend({
...
-          (a[this.orderMethod as keyof BaseItemDto] || '') <
-          (b[this.orderMethod as keyof BaseItemDto] || '')
-            ? 1
-            : -1
-        );
+        this.filterIs3d = undefined;
       }
+
+      console.dir(filter);
+
```
Camc314, line 198:
```suggestion

```

components/ItemGrid.vue
MrTimscampi: Yeah, I think it's okay for now. I'd prefer to properly fix it, but I have no idea why it does this...

ferferga: @MrTimscampi You don't specify the return type of the function inside the computed properties, so you're not referring to the Vue instance on this, rather on the ComputedProperties instance.

This is something I've seen we miss in this codebase and we should do a pass to verify that each function has it's return type defined correctly because it won't be long until they hit us like crazy with things like this.

I pushed a commit for fixing this, I hope you don't mind.

components/Card.vue
```diff
@@ -121,18 +127,21 @@ export default Vue.extend({
...
           case 'PhotoAlbum':
           case 'Playlist':
           case 'Video':
             return 'square-card';
           case 'Episode':
+          case 'Studio':
             return 'thumb-card';
           case 'Book':
           case 'BoxSet':
           case 'Movie':
```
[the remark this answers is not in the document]
MrTimscampi: Added above Movie (To keep a somewhat alphabetical order)

locales/cs.json
```diff
@@ -1,60 +1,60 @@
 {
-    "alphabetically": "Abecedně",
```
Camc314, line 2: Indents?

pages/library/_viewId.vue
```diff
@@ -181,105 +149,174 @@ export default Vue.extend({
...
+        this.itemsCount = itemsResponse.TotalRecordCount;
       } catch (error) {
-        // Can't get given library ID
-        this.$nuxt.error({
-          statusCode: 404,
-          message: this.$t('libraryNotFound') as string
-        });
+        console.error('Unable to refresh items:', error);
+        this.items = [];
+        this.itemsCount = 0;
```
Camc314, line 279: Use the snackbar?

pages/library/_viewId.vue
```diff
@@ -1,87 +1,61 @@
...
+    </v-app-bar>
+    <v-container class="after-second-toolbar">
+      <item-grid v-if="items.length" :loading="loading" :items="items">
         <h1 class="text-h5">
           {{ $t('libraryEmpty') }}
         </h1>
-      </div>
-    </v-row>
-  </v-container>
+      </item-grid>
```
Camc314, line 28:
```suggestion
      <item-grid v-if="items.length" :loading="loading" :items="items" />
```
It shows an empty library error in the itemGrid?

MrTimscampi: The condition is already inside of ItemGrid, so I removed it from here.

pages/library/_viewId.vue
MrTimscampi: Done

locales/cs.json
MrTimscampi: I reset the file to master, since there's no change in it (We should really get Weblate to respect indent size, though...)

components/ItemGrid.vue
```diff
@@ -0,0 +1,134 @@
...
+    loading: {
+      type: Boolean,
+      required: true
+    }
+  },
+  computed: {
+    itemsChunks(): Array<{ [id: number]: BaseItemDto }> {
+      let cardsPerLine = 8;
+
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
ferferga, line 62: Why are we still having this after my fix? I've got no linter errors back when I committed and everything worked flawlessly.

I can no longer go back for it because you force pushed.

MrTimscampi: Indeed, I forgot to re-remove them when porting over your change (The commit you had pushed undid some changes I had made before, so I re-applied the changes on top of the cleaned-up branch).

I just pushed a commit that removes these `ts-ignore`.

components/Card.vue
```diff
@@ -173,12 +184,8 @@ export default Vue.extend({
      * @returns {string} Either an empty string, or a string representing the production year(s) for the current item or the relevant episode number of the item.
      */
     cardSubtitle(): string {
-      if (this.episode) {
-        if (this.item.IndexNumber) {
-          return this.$t('episodeNumber', {
-            episodeNumber: this.item.IndexNumber
-          }).toString();
-        }
```
Camc314, line 181: Probably out of the scope for this, but the title and subtitle for cards (especially for the up next/continue watching) need to be fixed

Camc314 (approved):
LGTM

components/ItemGrid.vue
```diff
@@ -1,84 +1,124 @@
...
+        </slot>
+      </div>
+    </v-row>
+  </div>
 </template>
 
 <script lang="ts">
+import { chunk } from 'lodash';
 import Vue from 'vue';
+import { BaseItemDto } from '~/api';
```
Camc314, line 45:
```suggestion
```

components/ItemGrid.vue
```diff
@@ -1,84 +1,124 @@
...
+      const chunks = chunk(this.items, cardsPerLine);
+
+      const keyedChunks = chunks.map((itemChunk, index) => {
+        return {
+          id: index,
+          chunk: itemChunk
+        };
+      });
+
+      return keyedChunks;
```
Camc314, line 85:
```suggestion
      return chunks.map((itemChunk, index) => {
        return {
          id: index,
          chunk: itemChunk
        };
      });

Camc314 (commented):
Ready once the code smells are gone

components/Card.vue
Camc314: Improved in #304

components/ItemGrid.vue
MrTimscampi: It's used on line 59.

PR #293, opened 2020-11-26
feat(login): add public users

### LINKED ISSUE #138: Add cards for the visible users on the login screen

MrTimscampi:
When connecting to a server, this will fetch a list of public users. If users are found, it'll show a list instead of showing the login form.

image

Upon selecting a user, it's set as the currently active user, which changes the login form slightly.

image

Selecting "Login manually" will show the previous username/password entry form.

Clicking "Connect" on a user without a password will login without going through the form, as there is no password to enter.

Fixes #138


components/LoginForm.vue
```diff
@@ -67,10 +81,17 @@ export default Vue.extend({
     ...mapActions('deviceProfile', ['setDeviceProfile']),
     ...mapActions('snackbar', ['pushSnackbarMessage']),
     userLogin() {
+      if (!isEmpty(this.user)) {
+        // If we have a user from the public user selector, set it as login
+        this.login.username = this.user.Name || '';
```
ferferga, line 86: Reviewing from phone so perhaps I'm missing something obvious, but what's the purpose of this line? Why are we calling it "Name" when what we use is "Username"? I am confused.

components/LoginForm.vue
```diff
@@ -67,10 +81,17 @@ export default Vue.extend({
...
+      }
       this.loading = true;
       this.setDeviceProfile();
       this.$auth.loginWith('jellyfin', this.login);
       this.loading = false;
+    },
+    isEmpty(value: Record<any, any>) {
```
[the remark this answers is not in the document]
MrTimscampi: It's mainly to make it generic. The idea is that we need to wrap Lodash's `isEmpty` in a method in order to be able to use it in templates, and that doesn't care about whatever is passed to it, as long as it's a valid object.

components/LoginForm.vue
MrTimscampi: `this.login.username` is what is passed to the Nuxt Auth scheme.

`this.user` is a `UserDto` object, which has the user's name as a "Name" property.

So essentially, we're skipping having to manually enter the username, by using the name already provided to us by the API.

ferferga: @MrTimscampi Nice, thanks for the clarification. Shouldn't we use something different than an empty string though? Null or undefined for instance.

MrTimscampi: It's mainly to prevent a Typescript error. In practice, you'll never reach this without a valid UserDto object, which always has a name.

pages/login.vue
```diff
@@ -24,8 +65,36 @@ export default Vue.extend({
...
+  },
   methods: {
-    ...mapActions('page', ['setPageTitle'])
+    ...mapActions('page', ['setPageTitle']),
+    ...mapActions('deviceProfile', ['setDeviceProfile']),
+    isEmpty(value: Record<any, any>) {
+      return isEmpty(value);
+    },
+    setCurrentUser(user: UserDto) {
+      console.dir(user);
```
Camc314, line 82:
```suggestion
```

Camc314 (commented):
image

Needs some work for mobile devices

MrTimscampi:
Made some adjustments to make these responsive.

Screenshot_2020-12-01 jellyfin-vue
Screenshot_2020-12-01 jellyfin-vue(1)
Screenshot_2020-12-01 jellyfin-vue(2)

ferferga (changes requested):
@MrTimscampi Can you add the same padding between that exists between the cards and the "Select server" title to the buttons below?

So it looks symmetrical.

ferferga:
I would also like to raise that I think that we should us back arrows near the titles of the pages instead of buttons for going back to previous views. IIRC the startup wizard has arrows and, generally, I find them more common in UI design

MrTimscampi:
> I would also like to raise that I think that we should us back arrows near the titles of the pages instead of buttons for going back to previous views. IIRC the startup wizard (#284) has arrows and, generally, I find them more common in UI design

Well, the issue here is that it's not really a flow per se. Technically, you can come in at any point (Select server, select user or login form). So you need an explicit indication of which actions you can take.

ferferga (approved):

PR #295, opened 2020-11-27
fix(login): show correct error message

bbogdanov:
Had some trouble running the app with 10.6 and spend like 2h to understand that the server version is what I was having wrong in my dev configuration. 

Really tricky one to find when the error is suppressed like that. Had to read almost the entire code base to find it.


MrTimscampi (approved):

MrTimscampi:
> Really tricky one to find when the error is suppressed like that. Had to read almost the entire code base to find it.

Sorry for that, it was a known issue that some messages were suppressed, but nobody had looked into it yet.

Thanks for the fix :)

PR #299, opened 2020-11-28
fix: silence erroneous typescript errors

MrTimscampi:


ferferga:
@MrTimscampi Can we block this so I can check it next weekend? Maybe its a tiny thing like in the library one but I'm too busy this weekend.

MrTimscampi:
> @MrTimscampi Can we block this so I can check it next weekend? Maybe its a tiny thing like in the library one but I'm too busy this weekend.

Well, I'd like to get CI to be green again. You can certainly come back to it at a later date, though.

Camc314 (approved):

PR #300, opened 2020-11-28
refactor(login): move login and server request to vuex

bbogdanov:
### PR summary 

1. Login and Server request moved to vuex
2. Removing possibility to duplicate a server
3. Adding store value about current server used
4. Extracting some errors to reusable actions

### Additional notes:

It will affect #293


MrTimscampi (approved):
This is pretty much exactly what I had in mind for cleaning this up, thanks.

MrTimscampi:
Hm, is it possible to fix these Typescript issues though?

bbogdanov:
@MrTimscampi What do you think about moving to `rebase` instead of `merge`. The git history will become cleaner and easier to trace.

MrTimscampi:
> What do you think about moving to `rebase` instead of `merge`. The git history will become cleaner and easier to trace.

Indeed, I don't know why we didn't go that route. We tried squash initially, but that lost author information, then fell back to the regular merge like the other repos. But rebase does make sense.

MrTimscampi (approved):

bbogdanov:
#165

PR #302, opened 2020-11-29
refactor(navigation): rework navigation sidebar

MrTimscampi:
Reworks the navigation drawer a bit to look nicer and removes the duplicate user icon (Note: The screenshots were taken before adding the translation key for the header).

Screenshot_2020-11-29 jellyfin-vue Screenshot_2020-11-29 jellyfin-vue(1) Screenshot_2020-11-29 jellyfin-vue(2)

Edit: Added a chevron next to the user name, to indicate the menu
image

### DIFF
--- a/components/UserButton.vue
+++ b/components/UserButton.vue
@@ -1,23 +1,26 @@
 <template>
   <v-menu offset-y>
     <template #activator="{ on, attrs }">
-      <v-btn icon v-bind="attrs" v-on="on">
-        <v-avatar>
-          <v-img
-            v-if="userImage"
-            :src="userImage"
-            :alt="$auth.user.Name"
-          ></v-img>
-          <v-icon v-else dark>mdi-account</v-icon>
-        </v-avatar>
-      </v-btn>
+      <div class="d-flex" v-bind="attrs" v-on="on">
+        <div class="mr-4">
+          <v-btn icon>
+            <v-avatar>
+              <v-img
+                v-if="userImage"
+                :src="userImage"
+                :alt="$auth.user.Name"
+              ></v-img>
+              <v-icon v-else dark>mdi-account</v-icon>
+            </v-avatar>
+          </v-btn>
+        </div>
+        <h1 class="font-weight-light">
+          {{ $auth.user.Name }}
+          <v-icon>mdi-chevron-down</v-icon>
+        </h1>
+      </div>
     </template>
-    <v-list>
-      <v-list-item @click="switchColodScheme">
-        <v-switch>
-          <template #label>Toggle dark mode</template>
-        </v-switch>
-      </v-list-item>
+    <v-list dense>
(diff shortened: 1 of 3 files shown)


components/UserButton.vue
```diff
@@ -1,23 +1,26 @@
...
-          <v-icon v-else dark>mdi-account</v-icon>
-        </v-avatar>
-      </v-btn>
+      <div class="d-flex" v-bind="attrs" v-on="on">
+        <div class="mr-4">
+          <v-btn icon>
+            <v-avatar>
+              <v-img
+                v-if="userImage"
+                :src="userImage"
```
Camc314, line 10: Probably better to use the `imageHelper` mixin

Camc314 (commented):
LGTM.

On another note, the translation globe now looks a bit lonely on its own in the top bar. I guess this is slightly longer term, but should we use the user settings instead?

MrTimscampi:
> On another note, the translation globe now looks a bit lonely on its own in the top bar. I guess this is slightly longer term, but should we use the user settings instead?

That's the plan. Though, in the long term, the cast and SyncPlay icons are going to live there as well.

Extarys:
When using a freshly opened window, the dark mode toggle is defaulted to false, even though the dark mode is on.
image

Could this be fixed in this PR?

components/UserButton.vue
MrTimscampi: The main issue (for now) is that the current imageHelper mixing handles items (and a user isn't an item :/ )

I want to refactor it to clean it up and make it more modular, eventually, but I think that's outside the scope of this PR.

Camc314: Ok, sounds good.

components/UserButton.vue
```diff
@@ -1,23 +1,26 @@
...
-            :src="userImage"
-            :alt="$auth.user.Name"
-          ></v-img>
-          <v-icon v-else dark>mdi-account</v-icon>
-        </v-avatar>
-      </v-btn>
+      <div class="d-flex" v-bind="attrs" v-on="on">
+        <div class="mr-4">
+          <v-btn icon>
+            <v-avatar>
```
Camc314, line 7: Not essential, but I think it looks slightly better with a background color on the avatar icon

Extarys:
Tested it, looks neat. I love it.

dkanada:
What did the previous interface look like?

Camc314:
image
image
image

Something like this

components/UserButton.vue
MrTimscampi: Indeed, I think it was removed when moving stuff around between the two user picture components to make a single one. I didn't notice it due to having an avatar set at the time.

It's fixed in 53015bd :sweat_smile:

Camc314 (approved):

PR #312, opened 2020-12-01
Fix library empty text, add function to check for filters

Extarys:
(No idea why title is not what I set with npx cz, sorry)

When library is empty the text was at the bottom of the skeletons. CSS was misplaced.

Added a function to check if there are filters being applied to show proper message, increases UX.

Might also be useful in the future.

Please tell me if you can think of a better way to check for filters.
Screenshot from 2020-12-01 11-13-40


pages/library/_viewId.vue
```diff
@@ -81,6 +81,32 @@ export default Vue.extend({
...
+        this.ratingsFilter.length ||
+        this.filterHasSubtitles ||
+        this.filterHasTrailer ||
+        this.filterHasSpecialFeature ||
+        this.filterHasThemeSong ||
+        this.filterHasThemeVideo ||
+        this.filterIsHd ||
+        this.filterIs4k ||
+        this.filterIs3d ||
+        !defaultViews.includes(this.viewType)
```
ferferga, line 104: Reviewing on phone so I might miss something but why do we need this?

Extarys: The default message is "The library is empty", even if it's not when applying filters.
This allows to display a better message when there is no result "No results found", which is way better.

Although we could write a poem:

> Your library's empty
> And your filters are cloudy,
> Thy shall review their search,
> So thee day can be sunny.

Although now I'm thinking we could save the inital number of results when they load then compare that with the current number of results.

When movies load set variable to 10, refine search no results, we compare 10 != 0, so it's a filter not a library... Could be better and we could display [2/31 movies] when selecting a genre for example

ferferga: But I don't get why we need to check which kind of item the grid is displaying (Music, Books...) which you defined in the const above

Extarys: @ferferga My bad, it's because we can 'sort' by collection, artist, etc. (Technically a _view_) If there is no collection in Movies it doesn't mean there is no movies, it means there are no collection.

Way later we could do something like this:
Screenshot from 2020-12-01 13-16-11
Make the message more friendly with `There is no collection`, etc with a link that could show tips.

ferferga: @Extarys Been a long time without running master so I missed that stuff, so thank you for explaining!

ferferga (approved):

pages/library/_viewId.vue
Extarys: Np, I thought you asked why we needed the function - I don't use git much (therefore not familiar with how PR works) I was confused.

Hope you laughed at the _poem_ ;)

Extarys: Also, I gave some thoughts at how I could avoid a function with the variable that keep the total number of movies in memory while applying filters, but when websocket comes or if we come to that page with filters pre-applied this will break so a function is a little bit future-proof.

MrTimscampi: Some thoughts on this massive condition:

Wouldn't it be better to change the current message to "There is nothing here." (Effectively eliminating the need to check for the type of view we're in, thereby simplifying this a bit) as well as to move the check for filters to `onChangeFilter` ?

So we'd have a `data` named "hasFilters" or something, which would be set with the following (semi pseudo-code):
```typescript
onChangeFilter(filter: Record<string, any>) {
  this.hasFilters = Object.values(filter).every((value) => {
    return value.length > 0 || value !== false;
  })

  if (!hasFilters) return; // We don't have any filters, so no need to waste time doing more checks
}
```

This effectively remove the maintenance burden of having to change all the conditions manually if we ever add/remove filters in the future, while allowing us to keep the specific text for filters.

Extarys:
Sorry I messed the **en-US.json** file, I added a string to it but when I pushed, new lines appeared or  something.
Please be careful when merging or explain to me how I can fix it if's this is not ok. No idea how/why this happened.

MrTimscampi:
I fixed the en-US file (checked out the master one and re-applied the relevant changes), as well as rebased the PR (just for safety) and fixed a small issue in the code I had suggested.

This is good to go now :) Thanks.

MrTimscampi (approved):

Extarys:
I see, I didn't notice the check was wrong as I didn't experience weird behaviour with it.

Thanks mate!

PR #325, opened 2020-12-02
feat(validation): support for vee-validate

Extarys:
Initial support for vee-validate.

Validation messages have their translations in the `validation` object in the locales files.

A user suggested that the following was too repetitive (Form fields had their translations in the `fields` object. )
image

So I removed the field name from the translation.

Rules must be imported in `plugin/veeValidate.ts` or customized in the script section.

Import ValidationObserver, ValidationProvider component:
```
import { ValidationObserver, ValidationProvider } from 'vee-validate';
export default Vue.extend({
  components: {
    ValidationObserver,
    ValidationProvider
  },
...
```

Wrap the form and fields:
```
<validation-observer v-slot="{ invalid }">
      <v-form ref="form" :disabled="loading" @submit.prevent="userLogin">
            <validation-provider
              v-slot="{ errors }"
              name="username"
              :rules="rules.username"
            >
              <v-text-field
                v-model="login.username"
                outlined
                :label="$t('username')"
                :error-messages="errors"
              ></v-text-field>
            </validation-provider>
        <v-btn
              :disabled="invalid"
              color="primary"
              type="submit"
              >{{ $t('signIn') }}</v-btn>
```

Define rules:
```
rules: {
        username: {
          required: true
        }
}
```

For more info on how to use vee-validate, [refer to the official documentation](https://vee-validate.logaretm.com/v3/guide/forms.html#basic-example).

### DIFF
--- a/components/AddServerForm.vue
+++ b/components/AddServerForm.vue
@@ -40,13 +40,8 @@
 <script lang="ts">
 import Vue from 'vue';
 import { mapActions } from 'vuex';
-import { ValidationObserver, ValidationProvider } from 'vee-validate';
 
 export default Vue.extend({
-  components: {
-    ValidationObserver,
-    ValidationProvider
-  },
   data() {
     return {
       serverUrl: '',
--- a/components/AddServerForm.vue
+++ b/components/AddServerForm.vue
@@ -6,14 +6,20 @@
       :disabled="loading"
       @submit.prevent="connectToServer"
     >
-      <v-text-field
-        v-model="serverUrl"
-        outlined
-        :label="$t('serverAddress')"
-        type="url"
-        :rules="rules.serverUrlTest"
-        required
-      ></v-text-field>
+      <validation-provider
+        v-slot="{ errors }"
+        name="serverUrl"
+        :rules="rules.serverUrl"
+      >
+        <v-text-field
+          v-model="serverUrl"
+          outlined
+          :label="$t('serverAddress')"
+          type="url"
+          :error-messages="errors"
+          required
+        ></v-text-field>
+      </validation-provider>
       <v-row align="center" no-gutters>
         <v-col class="mr-2">
           <v-btn
@@ -37,20 +43,30 @@
 <script lang="ts">
 import Vue from 'vue';
 import { mapActions } from 'vuex';
+import { ValidationProvider } from 'vee-validate';
 
 export default Vue.extend({
+  components: {
+    ValidationProvider
+  },
   data() {
     return {
       serverUrl: '',
       validInputs: false,
       loading: false,
       rules: {
-        serverUrlTest: [
-          (v: string) => !!v || this.$t('serverAddressRequired'),
-          (v: string) =>
-            /^https?:\/\/.+/.test(v) || this.$t('serverAddressMustBeUrl')
-        ]
+        serverUrl: {
+          required: true,
+          mustBeUrl: true
+        }
       }
+      // rules: {
+      //   serverUrlTest: [
+      //     (v: string) => !!v || this.$t('serverAddressRequired'),
+      //     (v: string) =>
+      //       /^https?:\/\/.+/.test(v) || this.$t('serverAddressMustBeUrl')
+      //   ]
+      // }
     };
   },
   methods: {
--- a/components/AddServerForm.vue
+++ b/components/AddServerForm.vue
@@ -1,58 +1,55 @@
 <template>
   <div>
-    <v-form
-      ref="form"
-      v-model="validInputs"
-      :disabled="loading"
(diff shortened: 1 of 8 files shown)


Extarys:
Removed field name from strings:
image

Not sure who to tag but @MrTimscampi  if you could review this so we don't have to convert 20 forms to use vee-validate later :laughing: .

plugins/veeValidate.ts
```diff
@@ -0,0 +1,21 @@
...
+extend('mustBeUrl', (value: string): boolean => {
+  return /^https?:\/\/.+/.test(value);
+});
+
+const veeValidate: Plugin = ({ app }) => {
+  configure({
+    // FIXME: Ts doesn't like the returned value to defaultMessage
+    defaultMessage: (_field, values) => {
+      // values._field_ = app.i18n.t(`fields.${field}`);
+      return app.i18n.t(`validation.${values._rule_}`, values);
```
Camc314, line 17:
```suggestion
      return app.i18n.t(`validation.${values._rule_}`, values).toString();
```
This should fix the TS errors

components/AddServerForm.vue
```diff
@@ -1,55 +1,61 @@
...
   </div>
 </template>
 
 <script lang="ts">
 import Vue from 'vue';
 import { mapActions } from 'vuex';
+import { ValidationObserver, ValidationProvider } from 'vee-validate';
 
 export default Vue.extend({
+  components: {
```
MrTimscampi, line 46: Generally, we're heading more towards making components globally available instead of importing them locally

You can take a look at the files in `~/plugins/components/*.ts` to see how to do that.

components/LoginForm.vue
```diff
@@ -1,62 +1,69 @@
...
 
 <script lang="ts">
 import { isEmpty } from 'lodash';
 import Vue from 'vue';
 import { mapActions } from 'vuex';
+import { ValidationObserver, ValidationProvider } from 'vee-validate';
 import { UserDto } from '~/api';
 
 export default Vue.extend({
+  components: {
```
MrTimscampi, line 63: Same here

MrTimscampi (changes requested):
Really small stuff, otherwise it looks good

components/AddServerForm.vue
Extarys: My bad, I'll fix

plugins/veeValidate.ts
```diff
@@ -0,0 +1,21 @@
+import { Plugin } from '@nuxt/types/app';
+import { configure, extend } from 'vee-validate';
+import { required } from 'vee-validate/dist/rules';
+
+// Rules
+extend('required', required);
+
+extend('mustBeUrl', (value: string): boolean => {
```
MrTimscampi, line 8: Small question here. Any reason to use this instead of the already provided [url validator](https://vee-validate.logaretm.com/v2/guide/rules.html#url) ?

components/AddServerForm.vue
Extarys: Do you prefer for the component to be in the components folder or can I smash everything in the plugin/veeValidate.ts file?

plugins/veeValidate.ts
MrTimscampi: The rule provided by vee-validate seems to be base on [this](https://github.com/validatorjs/validator.js), and has a few options to have it handle URLs properly while not having to handle weird cases ourselves in the future.

Extarys: I'll switch - Since you were using a regex I wanted to have the exact same behavior and I forgot to investigate further.

Extarys: Looks like this rule is not available in v3: https://vee-validate.logaretm.com/v3/guide/rules.html#rules

MrTimscampi: Indeed, for some reason, the docs I had was V2, not V3. We can look into moving the rule to something better at some point down the line, no big deal :)

Extarys: Do you mean a better regex check or something more flexible?

I can put:
```js
return /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/.test(
    value
  );
```
```
http://something -> invalid
http://127.0.0.1 -> valid
https://server.local ->valid
```

Camc314: What about using `isUrl` from https://github.com/validatorjs/validator.js

MrTimscampi (approved):
LGTM.

This will serve as a test to see if Weblate messes up the structure of nested translation keys, as well :)

PR #329, opened 2020-12-03
feat(home): add scroller for latest items

MrTimscampi:
Basically jellyfin/jellyfin-web#1489 but with a sane framwork behind it.

*TV*
image
*Movie*
image
*Music*
image


dkanada:
You did the same weird side gradient padding for the normal web client :smile: It looks a lot better with no padding in my opinion.

MrTimscampi:
> You did the same weird side gradient padding for the normal web client :smile: It looks a lot better with no padding in my opinion.

We need a gradient of some sort to hit readability scores. If we do a gradient without this padding, half of the backdrop is behind the gradient, which is less nice.

This does essentially the same thing Netflix does, by offsetting the image a bit. Also I'd like to avoid cropping the backdrop, so unless I make it take the entire height (which looks too overwhelming), It has to have some padding :/

components/HomeHeader.vue
```diff
@@ -0,0 +1,160 @@
...
+              </h2>
+              <media-info :item="item" year runtime rating class="mt-2" />
+              <p class="mt-2" v-html="getOverview(item)" />
+              <v-btn
+                class="mr-2"
+                color="primary"
+                min-width="8em"
+                depressed
+                rounded
+                :to="`./${item.Id}/play`"
```
Extarys, line 47: This redirects to `http://127.0.0.1:3000/4f036c78-e69a-7350-74e6-cd5a387c6433/play` 
`This page could not be found`

It probably should be `http://127.0.0.1:3000/item/4f036c78-e69a-7350-74e6-cd5a387c6433/play`

Camc314 (approved):
Looks Great.

I've just added `@types/swiper` to `package.json`

components/HomeHeader.vue
```diff
@@ -0,0 +1,192 @@
...
+  background-size: contain;
+  background-repeat: no-repeat;
+  box-sizing: border-box;
+  mask-image: linear-gradient(
+      180deg,
+      rgba(18, 18, 18, 1) 60%,
+      rgba(18, 18, 18, 0) 100%
+    ),
+    linear-gradient(90deg, rgba(18, 18, 18, 1) 20%, rgba(18, 18, 18, 0) 70%);
+  mask-composite: subtract;
```
Camc314, line 168: This doesn't seem to work on chrome.



It looks like the bottom gradient is there, but not the horizontal

Camc314 (changes requested):

components/HomeHeader.vue
Camc314, line 168:
```suggestion
  mask-composite: subtract;
  -webkit-mask-composite: source-out;
```

Should fix it.

See [here](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-mask-composite)

MrTimscampi: Hm, we should have autoprefixer, which should do this automatically. I'll investigate.

Camc314: The auto prefixed is working but it is outputting '-webkit-mask-composite: subtract' which isn't a valid propert

MrTimscampi: Oh, okay, I get it. Autoprefixer handles it, but it doesn't convert between options. Since subtract doesn't exist for the webkit version, it doesn't work.

MrTimscampi: Fixed in 4ced6d0 :tada:

Camc314: Ah yeah that's right. 🎉

components/HomeHeader.vue
```diff
@@ -0,0 +1,193 @@
...
+  -webkit-mask-composite: source-out; // This is needed due to autoprefixed not converting subtract to the proper webkit equivalent
+  z-index: 1;
+  animation: backdrop-fadein 800ms ease-in normal both;
```
Camc314, line 171: This doesn't seem right

pages/index.vue
```diff
@@ -138,3 +153,21 @@ export default Vue.extend({
...
+@import '~vuetify/src/styles/styles.sass';
+.home-header-margin {
+  margin-top: -56px;
+}
+
+@media #{map-get($display-breakpoints, 'md-and-up')} {
+  .home-header-margin {
+    margin-top: -64px;
+  }
+}
```
Camc314, line 167: I think it'd be better to move this to the `HomeHeader.vue`. Because currently, if no items are returned, the searchbar and Libraries text will overlap

components/HomeHeader.vue
```diff
@@ -0,0 +1,193 @@
+<template>
+  <swiper v-show="items.length > 0" class="swiper" :options="swiperOptions">
```
Camc314, line 2: Would a v-if be better here?

components/HomeHeader.vue
MrTimscampi: Indeed, it's missing the keyframes, but it's not needed anyway.

components/HomeHeader.vue
MrTimscampi: Hm, yeah, no need to keep the swiper in DOM.

Camc314 (approved):

Nickbert7:
I've got an issue with an album with a long description.
The other items are looking fine so far.
image

PR #345, opened 2020-12-06
feat(search): add search

MrTimscampi:
It searches.

image


MrTimscampi:
Converted back to draft, as it needs additional UI design, since I'm not happy with the way this looks.

ferferga:
This probably should be in it's own discussion thread but:

What about renaming "People" tab to whatever category those "People" fall in? Just like current web, where they appear as "Artists" and "People".

Imo, we should have three categories:

* Artists
* Cast & Crew
* People

Although we don't have this feature yet, we shouldn't mix people's faces from photo libraries with the ones of movies/music/TV Shows. 

I don't know how this is currently implemented but we don't have to add tabs when there are no items that match the query, so this shouldn't increase the UI's weight.

Again, this is completely unnecessary to do right now, but I think it's better if we do it this way since the beginning, instead of getting users "accostumed" to it later.

Nickbert7:
I don't know if it is difficult to add, but would it be possible also include tags into the search?

pages/search.vue
```diff
@@ -0,0 +1,162 @@
...
+    <v-app-bar fixed flat dense class="second-toolbar">
+      <v-tabs v-model="searchTab" centered>
+        <v-tab :key="0">Top results</v-tab>
+        <v-tab :key="1">Movies</v-tab>
+        <v-tab :key="2">Shows</v-tab>
+        <v-tab :key="3">Albums</v-tab>
+        <v-tab :key="4">Tracks</v-tab>
+        <v-tab :key="5">Books</v-tab>
+        <v-tab :key="6">People</v-tab>
+        <v-tab :key="7">Artists</v-tab>
```
ThibaultNocchi, line 12: I think we can put those and the actual tabs in a v-for no?

store/search.ts
```diff
@@ -0,0 +1,25 @@
+import { ActionTree, MutationTree } from 'vuex';
+
+export interface SearchState {
+  query: string;
+}
+
+export const state = (): SearchState => ({
+  query: ''
+});
```
ThibaultNocchi, line 9:
```suggestion
const defaultState = (): SearchState => ({
  query: ''
});
export const state = defaultState;
```

And please add RESET_SEARCH commit and resetSearch action to call `Object.assign(state, defaultState())`

ThibaultNocchi (changes requested):

pages/search.vue
MrTimscampi: They need translating too.

These we could probably do. The pages, the idea is to be able to easily use a component inside of a page if we want later on.

ferferga:
Maybe we should replace the skeleton loader while searching with a spinning circle? There's a lot of jumpiness when searching for music, for instance.

MrTimscampi:
> Maybe we should replace the skeleton loader while searching with a spinning circle? There's a lot of jumpiness when searching for music, for instance.

I'm thinking about getting rid of the loader entirely, actually.

I think we should be able to combine `asyncData` and `fetch` to handle fetching information, as well as debouncing the search input to prevent it from triggering a search request on each input change.

This should make the experience a little better.

store/search.ts
ThibaultNocchi: @MrTimscampi If you wish I can add it for you :)

ferferga:
@Camc314 I remember mrTimscampi mentioned that he wanted to follow this design for this: https://github.com/jellyfin/jellyfin-vue/discussions/522

Camc314:
> @Camc314 In case you want to take over this PR, I remember MrTimscampi mentioned that he wanted to follow this design for this: https://github.com/jellyfin/jellyfin-vue/discussions/522
> 
> 
> 
> If not, I can take a look at it this weekend maybe.

Ah that design there looks good. All I have done, is rebased and moved some of the type fixings to a separate PR

store/search.ts
Camc314: Added

Camc314:
@ferferga I've got it to a good enough state to be reviewed, any improvements based on the discussion can be done at a later date imo.

ThibaultNocchi:
@Camc314 I got errors loading the home page, I'm trying to see what it is

```
TypeError: String.prototype.search called on incompatible undefined
    _render vue-i18n.esm.js:1605
    _interpolate vue-i18n.esm.js:1511
    _translate vue-i18n.esm.js:1731
    _t vue-i18n.esm.js:1756
    $t vue-i18n.esm.js:233
    render default.vue:188
    VueJS 30
    NuxtJS 4
    Babel 18
    NuxtJS 2
    Babel 12
    NuxtJS 11
client.js:97
    _callee$ NuxtJS
    Babel 8
    VueJS 32
    NuxtJS 4
    Babel 18
    NuxtJS 2
    Babel 12
    NuxtJS 11
```

```
TypeError: String.prototype.search called on incompatible undefined
    _render vue-i18n.esm.js:1605
    _interpolate vue-i18n.esm.js:1511
    _translate vue-i18n.esm.js:1731
    _t vue-i18n.esm.js:1756
    $t vue-i18n.esm.js:233
    render default.vue:188
    VueJS 13
    SET_USER_VIEWS userViews.ts:36
    wrappedMutationHandler vuex.esm.js:844
    commitIterator vuex.esm.js:466
    commit vuex.esm.js:465
    _withCommit vuex.esm.js:624
    commit vuex.esm.js:464
    boundCommit vuex.esm.js:409
    commit vuex.esm.js:796
    _callee$ userViews.ts:52
    Babel 10
    wrappedActionHandler vuex.esm.js:851
    dispatch vuex.esm.js:516
    boundDispatch vuex.esm.js:406
    dispatch vuex.esm.js:779
    mappedAction vuex.esm.js:1064
    beforeMount default.vue:183
    VueJS 18
client.js:97
```

locales/en-US.json
```diff
@@ -208,7 +209,10 @@
   "role": "Role",
   "save": "Save",
   "saved": "Saved",
-  "search": "Search",
+  "search": {
+    "search": "search",
```
ThibaultNocchi, line 213:
```suggestion
    "search": "Search",
```

IMO, it's weird not having this capitalized

ThibaultNocchi:
So I pushed the needed changes for the locales, we'll have to check before merging if no other new "search" strings were added and add them to our PR if it is

Camc314:
I don't think we should be changing strings in other locale files. It should be done on weblate. In prod, it shouldn't cause an issue, just produce an error in the console

ThibaultNocchi:
Well, this is not the case for me in production (static). It crashes the same way, the placeholder doesn't like an undefined value. Although I don't understand why i18n crashes when accessing the `search` subproperty.

Besides, it'd be a PITA to move the current search strings to the new object with weblate. :/

layouts/default.vue
```diff
@@ -161,6 +170,12 @@ export default Vue.extend({
       } else if (!this.$store.state.page.showNavDrawer) {
         this.showNavDrawer({ showNavDrawer: true });
       }
+    },
+    searchQuery(newQuery: string, oldQuery: string): void {
+      // If neither the old nor the new query are empty, we don't want to move.
+      if (newQuery !== '' && oldQuery === '') {
```
ferferga, line 176: We might want to trim this so writing spaces doesn't trigger the push either?

ThibaultNocchi: I agree

Camc314: Added

pages/search.vue
```diff
@@ -0,0 +1,167 @@
...
+      trackSearchResults: [] as BaseItemDto[],
+      bookSearchResults: [] as BaseItemDto[],
+      personSearchResults: [] as BaseItemDto[],
+      artistSearchResults: [] as BaseItemDto[]
+    };
+  },
+  computed: {
+    searchQuery(): string {
+      return this.$store.state.search.query;
+    }
```
ThibaultNocchi, line 73: Couldn't you just map the state? :)

pages/search.vue
```diff
@@ -0,0 +1,167 @@
...
+  },
+  beforeMount() {
+    if (this.searchQuery === '') {
+      this.$router.back();
+    }
+  },
+  destroyed() {
+    this.setSearchQuery({ query: '' });
+  },
```
ThibaultNocchi, line 91: I'd avoid a store for this unique goal, and just pass it from `default` as a prop and use an event or something to clear it in the parent

pages/search.vue
Camc314: Mmm maybe, but because there is both a get and a set, I don't think so

pages/search.vue
Camc314: We need a store to be able to pass the query from `default` layout to search page

ThibaultNocchi: Why though, couldn't a prop do it?

pages/search.vue
ThibaultNocchi: It'll depend on how we resolve our other point ^^

pages/search.vue
Camc314: Because we don't use the search page as a component, rather as a page, so in the code there is never <search-page />, it lives inside <nuxt> inside the default layout

ThibaultNocchi: Oh yeah that's right, it didn't tick. Couldn't then the query be put into the URL?

Camc314: That's not a bad shout

Camc314: I've updated it to use query params. I haven't squashed it, so once it is all good and approved, then I'll squash...

pages/search.vue
```diff
@@ -0,0 +1,162 @@
...
+  computed: {
+    searchQuery(): string {
+      return this.$route.query?.q || '';
+    }
+  },
+  watch: {
+    searchQuery(newQuery: string, _oldQuery: string): void {
+      if (newQuery === '') {
+        this.$router.back();
```
ThibaultNocchi, line 77: Won't that be interfering with the setter in default.vue?

pages/search.vue
```diff
@@ -0,0 +1,162 @@
...
+      } else {
+        this.performSearch();
+      }
+    }
+  },
+  beforeMount() {
+    if (this.searchQuery === '') {
+      this.$router.back();
```
ThibaultNocchi, line 85: Same here?

Camc314: I think here we still want to go to back if something has gone wrong, and the query is ''

pages/search.vue
Camc314: yeah, I've removed the `back()` here

pages/search.vue
ThibaultNocchi: Oh yeah didn't think about the `beforeMount`, np then :)

ThibaultNocchi: Okay yeah I see it, it's good then :)

ThibaultNocchi:
I think we need debouncing / throttling also

ferferga:
@jellyfin rebase

ThibaultNocchi:
@jellyfin rebase

ThibaultNocchi (approved):

neopc10:
Artists results seem not to populate at all and People do not appear in the Top Results, even if it contains the only result - example: Sean Connery.

At 100% zoom at half-screen width on a 1080 display, the search bar could also be wider when expanded. It's currently green and could be brought out to brown at minimum. It's already been widened out on full-screen width.

image

image

ThibaultNocchi:
@neopc10 The search bar should now expend on smaller sizes if you wanna check ;)

hamburglar2160:
If the general design for this is moving forward, I'd eventually like to see categories of results that have zero results greyed out/disappeared like for artists/people details. An item count, maybe next to each category tab, would be useful here too.

MrTimscampi:
@ferferga When you have some free time, could you give this a review, so we can move forward with improving it, please?

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

client/pages/search.vue
```diff
@@ -0,0 +1,167 @@
...
+
+      this.loading = false;
+    }
+  }
+});
+</script>
+
+<style lang="scss" scoped>
+@import '~vuetify/src/styles/styles.sass';
+.second-toolbar {
```
dkanada, line 148:
```suggestion

.second-toolbar {
```

ferferga (changes requested):
Search doesn't work outside the main page, so you can't search if you're browsing a library, for instance.

client/layouts/default.vue
```diff
@@ -161,6 +162,20 @@ export default Vue.extend({
     ...mapState('page', ['opaqueAppBar', 'navDrawer']),
     ...mapState('user', ['accessToken']),
     ...mapState('deviceProfile', ['deviceId']),
+    searchQuery: {
+      get(): string {
+        return this.$route.query.q?.toString();
+      },
+      set(value: string): void {
+        if (value === '' || !value) {
+          this.$router.go(-1);
```
dkanada, line 171:
```suggestion
          this.$router.back();
```

dkanada: pretty sure this exists and is more convenient

MrTimscampi:
@jellyfin-bot rebase

PR #346, opened 2020-12-06
Feat/up next

AlfHou:
Implementing the up-next component. Related pull request from [jellyfin-web](https://github.com/jellyfin/jellyfin-web/pull/1825)

### DIFF
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -1,14 +1,14 @@
 <template>
   <v-fade-transition>
-    <v-card class="container pa-4">
+    <v-card class="container white--text pa-4">
       <div class="d-flex flex-column flex-grow-1">
-        <v-card-title class="countdown-header pa-0">
+        <v-card-title class="text-h6 pa-0 my-1 mx-0">
           {{ $t('dialog.upNext.nextEpisodePlayingIn') }}
           <span class="primary--text darken-2">
             &ensp;{{ timeLeft }} {{ $t('seconds').toLowerCase() }}
           </span>
         </v-card-title>
-        <v-card-subtitle class="title subtitle-1 pa-0">
+        <v-card-subtitle class="mt-1 mx-0 mb-2 text-truncate subtitle-1 pa-0">
           {{ getNextItem.SeriesName }} -
           {{
             $t('tvShowAbbrev', {
@@ -71,8 +71,7 @@ export default Vue.extend({
   bottom: 0;
   width: 100%;
   will-change: transform, opacity;
-  background-color: rgba(0, 0, 0, 0.7);
-  color: #fff;
+  background-color: rgba(map-get($shades, 'black'), 0.7);
   user-select: none;
   z-index: 6;
   -webkit-touch-callout: none;
@@ -83,16 +82,5 @@ export default Vue.extend({
     width: 30em;
     margin: 0 2em 6em 0;
   }
-}
-.countdown-header {
-  margin: 0.25em 0;
-  font-weight: 500;
-}
-.title {
-  width: 100%;
-  margin: 0.25em 0 0.5em;
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
 }
 </style>
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -16,7 +16,7 @@
               episodeNumber: nextEpisodeNumber
             })
           }}
-          - <span v-if="$vuetify.breakpoint.xsOnly"> <br /> </span>
+          <span v-if="$vuetify.breakpoint.smAndUp"> - </span> <br v-else />
           {{ nextName }}
         </v-card-subtitle>
         <div>
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -9,19 +9,21 @@
           </span>
         </v-card-title>
         <v-card-subtitle class="title subtitle-1 pa-0">
-          {{ nextSeriesName }} -
+          {{ getNextItem.SeriesName }} -
           {{
             $t('tvShowAbbrev', {
-              seasonNumber: nextSeasonNumber,
-              episodeNumber: nextEpisodeNumber
+              seasonNumber: getNextItem.ParentIndexNumber,
+              episodeNumber: getNextItem.IndexNumber
             })
           }}
           <span v-if="$vuetify.breakpoint.smAndUp"> - </span> <br v-else />
-          {{ nextName }}
+          {{ getNextItem.Name }}
         </v-card-subtitle>
         <div>
-          {{ nextRunTime }}
-          <span class="pl-4">{{ $t('endsAt', { time: nextEndsAt }) }} </span>
+          {{ getRuntimeTime(getNextItem.RunTimeTicks) }}
+          <span class="pl-4"
+            >{{ $t('endsAt', { time: nextEndsAt(getNextItem.RunTimeTicks) }) }}
+          </span>
         </div>
         <v-card-actions>
           <v-spacer />
@@ -49,27 +51,13 @@ export default Vue.extend({
     }
   },
   computed: {
-    ...mapGetters('playbackManager', ['getNextItem']),
-    nextSeriesName(): string {
-      return this.getNextItem?.SeriesName;
-    },
-    nextEpisodeNumber(): number {
-      return this.getNextItem?.IndexNumber;
-    },
-    nextSeasonNumber(): number {
-      return this.getNextItem?.ParentIndexNumber;
-    },
-    nextRunTime(): string {
-      return this.getRuntimeTime(this.getNextItem?.RunTimeTicks);
-    },
-    nextEndsAt(): string {
-      const seconds =
-        this.ticksToMs(this.getNextItem?.RunTimeTicks) + this.timeLeft * 1000;
+    ...mapGetters('playbackManager', ['getNextItem'])
+  },
+  methods: {
+    nextEndsAt(runtimeTicks: number): string {
+      const seconds = this.ticksToMs(runtimeTicks) + this.timeLeft * 1000;
 
       return this.$dateFns.format(Date.now() + seconds, 'p');
-    },
-    nextName(): string {
-      return this.getNextItem?.Name;
     }
   }
 });
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -37,7 +37,7 @@
 
 <script lang="ts">
 import Vue from 'vue';
-import { mapState, mapGetters, mapActions } from 'vuex';
+import { mapGetters } from 'vuex';
 import timeUtils from '~/mixins/timeUtils';
 
 export default Vue.extend({
@@ -49,7 +49,6 @@ export default Vue.extend({
     }
   },
   computed: {
-    ...mapState('playbackManager', ['queue', 'currentItemIndex']),
     ...mapGetters('playbackManager', ['getNextItem']),
     nextSeriesName(): string {
       return this.getNextItem?.SeriesName;
@@ -72,9 +71,6 @@ export default Vue.extend({
     nextName(): string {
       return this.getNextItem?.Name;
     }
-  },
-  methods: {
-    ...mapActions('playbackManager', ['setNextTrack'])
   }
 });
 </script>
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -11,7 +11,7 @@
         <v-card-subtitle class="mt-1 mx-0 mb-2 text-truncate subtitle-1 pa-0">
           {{ getNextItem.SeriesName }} -
           {{
-            $t('tvShowAbbrev', {
+            $t('seasonEpisodeAbbrev', {
               seasonNumber: getNextItem.ParentIndexNumber,
               episodeNumber: getNextItem.IndexNumber
             })
@@ -80,7 +80,7 @@ export default Vue.extend({
 @media #{map-get($display-breakpoints, 'md-and-up')} {
   .container {
     width: 30em;
-    margin: 0 2em 6em 0;
+    margin: 0 2em 10em 0;
   }
 }
 </style>
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -3,12 +3,15 @@
     <v-card class="container white--text pa-4">
       <div class="d-flex flex-column flex-grow-1">
         <v-card-title class="text-h6 pa-0 my-1 mx-0">
-          {{ $t('dialog.upNext.nextEpisodePlayingIn') }}
+          {{ $t('dialog.upNext.nextItemPlayingIn') }}
           <span class="primary--text darken-2">
             &ensp;{{ timeLeft }} {{ $t('seconds').toLowerCase() }}
           </span>
         </v-card-title>
-        <v-card-subtitle class="mt-1 mx-0 mb-2 text-truncate subtitle-1 pa-0">
+        <v-card-subtitle
+          v-if="getCurrentItem.Type === 'Episode'"
+          class="mt-1 mx-0 mb-2 text-truncate subtitle-1 pa-0"
+        >
           {{ getNextItem.SeriesName }} -
           {{
             $t('seasonEpisodeAbbrev', {
@@ -17,6 +20,12 @@
             })
           }}
           <span v-if="$vuetify.breakpoint.smAndUp"> - </span> <br v-else />
+          {{ getNextItem.Name }}
+        </v-card-subtitle>
+        <v-card-subtitle
+          v-if="getCurrentItem.Type === 'Movie'"
+          class="mt-1 mx-0 mb-2 text-truncate subtitle-1 pa-0"
+        >
           {{ getNextItem.Name }}
         </v-card-subtitle>
         <div>
@@ -51,7 +60,7 @@ export default Vue.extend({
     }
   },
   computed: {
-    ...mapGetters('playbackManager', ['getNextItem'])
+    ...mapGetters('playbackManager', ['getNextItem', 'getCurrentItem'])
   },
   methods: {
     nextEndsAt(runtimeTicks: number): string {
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -66,6 +66,7 @@ export default Vue.extend({
     nextEndsAt(): string {
       const seconds =
         this.ticksToMs(this.getNextItem?.RunTimeTicks) + this.timeLeft * 1000;
+
       return this.$dateFns.format(Date.now() + seconds, 'p');
     },
     nextName(): string {
--- a/client/components/Item/UpNext.vue
+++ b/client/components/Item/UpNext.vue
@@ -1,72 +1,140 @@
 <template>
-  <v-fade-transition>
-    <v-card class="container white--text pa-4">
-      <div class="d-flex flex-column flex-grow-1">
-        <v-card-title class="text-h6 pa-0 my-1 mx-0">
-          {{ $t('dialog.upNext.nextItemPlayingIn') }}
-          <span class="primary--text darken-2">
-            &ensp;{{ timeLeft }} {{ $t('seconds').toLowerCase() }}
-          </span>
-        </v-card-title>
-        <v-card-subtitle
-          v-if="getCurrentItem.Type === 'Episode'"
-          class="mt-1 mx-0 mb-2 text-truncate subtitle-1 pa-0"
-        >
-          {{ getNextItem.SeriesName }} -
-          {{
-            $t('seasonEpisodeAbbrev', {
-              seasonNumber: getNextItem.ParentIndexNumber,
-              episodeNumber: getNextItem.IndexNumber
-            })
-          }}
-          <span v-if="$vuetify.breakpoint.smAndUp"> - </span> <br v-else />
-          {{ getNextItem.Name }}
-        </v-card-subtitle>
-        <v-card-subtitle
-          v-if="getCurrentItem.Type === 'Movie'"
-          class="mt-1 mx-0 mb-2 text-truncate subtitle-1 pa-0"
-        >
-          {{ getNextItem.Name }}
-        </v-card-subtitle>
-        <div>
-          {{ getRuntimeTime(getNextItem.RunTimeTicks) }}
-          <span class="pl-4"
-            >{{ $t('endsAt', { time: nextEndsAt(getNextItem.RunTimeTicks) }) }}
-          </span>
-        </div>
-        <v-card-actions>
-          <v-spacer />
-          <v-btn class="primary darken-2" @click="$emit('startNext')">
-            {{ $t('dialog.upNext.startNow') }}
-          </v-btn>
-          <v-btn @click="$emit('hide')"> {{ $t('dialog.upNext.hide') }}</v-btn>
-        </v-card-actions>
-      </div>
-    </v-card>
-  </v-fade-transition>
+  <v-container
+    v-if="visible"
+    class="up-next-dialog pointer-events-none pa-lg-6"
+  >
+    <v-row>
+      <v-col
+        cols="12"
+        offset-md="6"
+        md="6"
+        offset-lg="8"
+        lg="4"
+        offset-xl="9"
+        xl="3"
+      >
+        <v-card class="pointer-events-all">
+          <v-card-title class="text-h6">
+            <i18n path="dialog.upNext.nextItemPlayingIn" tag="span">
+              <template #time>
+                <span class="primary--text darken-2">
+                  {{ $tc('units.time.seconds', currentItemTimeLeft) }}
+                </span>
+              </template>
+            </i18n>
+          </v-card-title>
+          <v-card-subtitle class="text-truncate subtitle-1">
+            <span v-if="getCurrentItem.Type === 'Episode'">
+              {{ getNextItem.SeriesName }} -
+              {{
+                $t('seasonEpisodeAbbrev', {
+                  seasonNumber: getNextItem.ParentIndexNumber,
+                  episodeNumber: getNextItem.IndexNumber
+                })
+              }}
+              <span v-if="$vuetify.breakpoint.smAndUp"> - </span> <br v-else />
+              {{ getNextItem.Name }}
+            </span>
+            <span v-if="getCurrentItem.Type === 'Movie'">
+              {{ getNextItem.Name }}
+            </span>
+          </v-card-subtitle>
+          <v-card-text>
+            <span>
+              {{ getRuntimeTime(getNextItem.RunTimeTicks) }}
+              <span class="pl-4"
+                >{{
+                  $t('endsAt', {
+                    time: getEndsAtTime(getNextItem.RunTimeTicks)
+                  })
+                }}
+              </span>
+            </span>
+          </v-card-text>
+          <v-card-actions>
+            <v-spacer />
+            <v-btn class="primary darken-2" depressed @click="setNextTrack()">
+              {{ $t('dialog.upNext.startNow') }}
+            </v-btn>
+            <v-btn depressed outlined @click="isHiddenByUser = true">
+              {{ $t('dialog.upNext.hide') }}
+            </v-btn>
+          </v-card-actions>
+        </v-card>
+      </v-col>
+    </v-row>
+  </v-container>
 </template>
 
 <script lang="ts">
 import Vue from 'vue';
-import { mapGetters } from 'vuex';
+import { mapGetters, mapState } from 'vuex';
 import timeUtils from '~/mixins/timeUtils';
 
 export default Vue.extend({
   mixins: [timeUtils],
-  props: {
-    timeLeft: {
-      type: Number,
-      required: true
-    }
+  data() {
+    return {
+      isHiddenByUser: false
+    };
   },
   computed: {
-    ...mapGetters('playbackManager', ['getNextItem', 'getCurrentItem'])
-  },
-  methods: {
-    nextEndsAt(runtimeTicks: number): string {
-      const seconds = this.ticksToMs(runtimeTicks) + this.timeLeft * 1000;
+    ...mapState('playbackManager', [
+      'currentItemIndex',
+      'currentTime',
+      'isMinimized'
+    ]),
+    ...mapGetters('playbackManager', [
+      'getCurrentlyPlayingMediaType',
+      'getNextItem',
+      'getCurrentItem',
+      'setNextTrack'
+    ]),
+    currentItemDuration(): number {
+      return this.ticksToMs(this.getCurrentItem?.RunTimeTicks) / 1000;
+    },
+    currentItemTimeLeft(): number {
+      return Math.round(this.currentItemDuration - this.currentTime);
+    },
+    visible(): boolean {
+      if (
+        this.isMinimized ||
+        this.isHiddenByUser ||
+        this.getCurrentlyPlayingMediaType !== 'Video' ||
+        !this.getNextItem
+      ) {
+        return false;
+      }
 
-      return this.$dateFns.format(Date.now() + seconds, 'p');
+      if (this.currentItemTimeLeft <= this.nextUpDuration) {
+        return true;
+      }
+
+      return false;
+    },
+    nextUpDuration(): number {
+      // If longer than 5 hours, set the duration to 9 minutes
+      if (this.currentItemDuration >= 5 * 60 * 60) {
+        return 540;
+      }
+      // If longer than 2 hours, set the duration to 3.5 minutes
+      else if (this.currentItemDuration >= 2 * 60 * 60) {
+        return 210;
+      }
+      // If longer than 45 minutes, set the duration to 2 minutes
+      else if (this.currentItemDuration >= 45 * 60) {
+        return 120;
+      }
+
+      return 45;
+    }
+  },
+  watch: {
+    currentItemIndex(): void {
+      this.isHiddenByUser = false;
+    },
+    visible(): void {
+      this.$emit('change', this.visible);
     }
   }
 });
@@ -74,22 +142,10 @@ export default Vue.extend({
 <style lang="scss" scoped>
 @import '~vuetify/src/styles/styles.sass';
 
-.container {
-  position: fixed;
-  right: 0;
+.up-next-dialog {
+  position: absolute;
   bottom: 0;
-  width: 100%;
-  will-change: transform, opacity;
-  background-color: rgba(map-get($shades, 'black'), 0.7);
-  user-select: none;
-  z-index: 6;
-  -webkit-touch-callout: none;
-}
-
-@media #{map-get($display-breakpoints, 'md-and-up')} {
-  .container {
-    width: 30em;
-    margin: 0 2em 10em 0;
-  }
+  right: 0;
+  z-index: 9999;
 }
 </style>
--- a/client/components/Players/PlayerManager.vue
+++ b/client/components/Players/PlayerManager.vue
@@ -25,6 +25,7 @@
         />
         <v-hover v-slot="{ hover }">
           <v-card class="player-card" width="100%">
+<<<<<<< HEAD
             <v-container fill-height fluid class="pa-0 justify-center">
               <shaka-player ref="videoPlayer" />
             </v-container>
@@ -83,6 +84,56 @@
                     >
                       <v-icon size="32">mdi-skip-next</v-icon>
                     </v-btn>
+=======
+            <video-player ref="videoPlayer" />
+          <up-next
+            v-if="showUpNext"
+            :time-left="timeLeft"
+            @hide="upNextUserHidden = true"
+          />
+          <!-- Mini Player Overlay -->
+          <v-fade-transition>
+            <v-overlay v-show="hover && isMinimized" absolute>
+              <div class="d-flex flex-column player-overlay">
+                <div class="d-flex flex-row">
+                  <v-btn icon @click="toggleMinimized">
+                    <v-icon>mdi-arrow-expand-all</v-icon>
+                  </v-btn>
+                  <v-spacer />
+                  <v-btn icon @click="stopPlayback">
+                    <v-icon>mdi-close</v-icon>
+                  </v-btn>
+                </div>
+                <div
+                  class="absolute d-flex flex-row justify-center align-center"
+                >
+                  <v-btn
+                    class="all-pointer-events"
+                    icon
+                    large
+                    @click="setPreviousTrack"
+                  >
+                    <v-icon size="32">mdi-skip-previous</v-icon>
+                  </v-btn>
+                  <v-btn
+                    class="all-pointer-events"
+                    icon
+                    x-large
+                    @click="playPause"
+                  >
+                    <v-icon size="48">
+                      {{ isPaused ? 'mdi-play' : 'mdi-pause' }}
+                    </v-icon>
+                  </v-btn>
+                  <v-btn
+                    class="all-pointer-events"
+                    icon
+                    large
+                    @click="setNextTrack"
+                  >
+                    <v-icon size="32">mdi-skip-next</v-icon>
+                  </v-btn>
+>>>>>>> 8c0ac403 (feat(up-next): use v-card as upNext container)
                   </div>
                 </div>
               </v-overlay>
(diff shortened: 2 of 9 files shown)


ferferga:
@AlfHou how is this going? It's ready to be reviewed?

AlfHou:
@ferferga not yet. I was away last week, so haven’t gotten as much done as I would’ve hoped. I saw that there was a OSD added, so I’ll try to integrate with that maybe?

If you feel like I am blocking your progress, feel free to complete it before me btw. Don’t want to hold your progress back 😬

ferferga:
@AlfHou Oh, not at all. Have your own place, this is a hobby after all.

Just asking because I saw from time to time notifications about your pushes here, so I thought that you were endlessly fixing conflicts and expecting us to review it when it was marked as a draft. 

So TLDR, remember to mark this as ready as soon as we can chime in, otherwise you will be stuck with this for a long time :).

AlfHou:
It took a while, but I believe this is ready to be reviewed now.
image

Camc314:
Thanks AlfHou, looks good, Would it be possible to squash some of your commits so they make a bit more sense?

AlfHou:
@Camc314 Sure! Better now? :)

Camc314:
> @Camc314 Sure! Better now? :)
> 
> 

Yep thanks, I'll take a more detailed look either tonight or tomorrow morning.

locales/nb_NO.json
```diff
@@ -1,375 +1,379 @@
...
-        "serverAddressMustBeUrl": "Serveradressen må være en gyldig URL",
-        "serverAddress": "Serveradresse",
-        "serverVersionTooLow": "Serveren må være versjon 10.7.0 eller nyere",
-        "selectUser": "Velg en bruker",
-        "rememberMe": "Husk meg",
-        "loginAs": "Logg inn som {name}",
-        "addServer": "Legg til server",
-        "selectServer": "Velg server",
-        "login": "Logg inn"
+  "3DFormat": "3D-format",
```
Camc314, line 2: We leave all locale files apart from en-us unchanged. Other languages are translated on [Weblate](https://translate.jellyfin.org/projects/jellyfin/jellyfin-vue/)

components/Item/UpNext.vue
```diff
@@ -0,0 +1,126 @@
...
+        <span class="primary--text darken-2"
+          >{{ timeLeft }} {{ $t('seconds') }}</span
+        >
+      </h2>
+      <h3 class="title subtitle-1">
+        {{ nextSeriesName }} - S{{ nextSeasonNumber }}:E{{
+          nextEpisodeNumber
+        }}
+        - {{ nextName }}
+      </h3>
```
Camc314, line 15: This should be translated. See [here](https://github.com/jellyfin/jellyfin-vue/blob/master/components/Item/Card/Card.vue#L222-L225) and [here](https://github.com/jellyfin/jellyfin-vue/blob/master/locales/en-US.json#L364)

components/Item/UpNext.vue
```diff
@@ -0,0 +1,126 @@
+<template>
+  <div class="container">
+    <div class="d-flex flex-column flex-grow-1">
```
Camc314, line 3: we could use a v-card instead of a div here

components/Item/UpNext.vue
AlfHou: Ah, neat. I'm assuming you meant to link to line 344 and not `"version": "Version",` in the second link?

ferferga: Yes

components/Players/PlayerManager.vue
```diff
@@ -193,12 +206,42 @@ export default Vue.extend({
...
+      }
+      return true;
     }
   },
   created() {
     this.$store.subscribe((mutation, state: AppState) => {
       switch (mutation.type) {
         case 'playbackManager/INCREASE_QUEUE_INDEX':
+          this.upNextUserHidden = false;
+          break;
```
ferferga, line 244: What if I go back to the previous item while the popup is shown? Will the warning be persistent, or it will be removed because the percentage completed is less than 97%?

locales/en-US.json
```diff
@@ -1,5 +1,6 @@
 {
```
ferferga, line 1: Why this file has changes that are unrelated to this feature? ``noMediaSourcesAvailable`` is a duplicate, was moved to the bottom.

locales/en-US.json
```diff
@@ -152,18 +154,18 @@
   "movies": "Movies",
   "name": "Name",
   "networks": "Networks",
+  "nextEpisodePlayingIn": "Next Episode Playing in",
```
ferferga, line 157:
```suggestion
  "nextEpisodePlayingIn": "Next episode playing in {time} seconds",
```

And remember to change the occurrences in the code.

ferferga (changes requested):

components/Players/PlayerManager.vue
```diff
@@ -20,9 +20,15 @@
...
+          <up-next
+            v-if="showUpNext"
+            :time-left="timeLeft"
+            @hide="upNextUserHidden = true"
+          />
           <!-- Mini Player Overlay -->
           <v-fade-transition>
-            <v-overlay v-show="hover && isMinimized" absolute>
+            <!-- z-index: 5 is the default, but hardcoded in case of changes -->
+            <v-overlay v-show="hover && isMinimized" absolute z-index="5">
```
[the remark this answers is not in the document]
AlfHou: Mostly to make sure that line 100 in `Items/UpNext.vue` which is: `z-index: 6;` is above the overlay, but it might be a bit redundant as we would probably look for visual errors when updating vuetify. I'll remove it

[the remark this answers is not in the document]
ferferga: Now that you're using a card component this will probably be fine.

locales/en-US.json
AlfHou: I don't think this would work as we want the `{time} seconds` part to be in a different color than the rest

locales/en-US.json
AlfHou: I used the i18n ally vs-code extension to add them which seems to have sorted them alphabetically. Removed the unrelated changes.

components/Players/PlayerManager.vue
AlfHou: If you go back it would've stayed hidden even at the last 97%. Fixed this.

ferferga:
@AlfHou One thing I forgot in my initial review, sorry 😅

Can you place the relevant strings in a subkey in the locale file?

Something like ``dialogs.upNext`` would be great, that way we have the strings more organised and are easily recognisable by translators.

Ping me if you need an example or something.

locales/en-US.json
ferferga: Oh, true. Fair enough.

AlfHou:
@ferferga done :) I left the string "seconds" out as it might be reusable in other cases, let me know if I should move it.

locales/en-US.json
```diff
@@ -30,6 +30,13 @@
   "darkModeToggle": "Toggle dark mode",
   "dateAdded": "Date added",
   "details": "Details",
+  "dialog": {
+    "upNext": {
+      "nextEpisodePlayingIn": "Next Episode Playing in",
```
ferferga, line 35:
```suggestion
      "nextEpisodePlayingIn": "Next episode playing in",
```

ferferga (changes requested):
With this and the changes to the relevant card components for titles, subtitles and such, I think we're good to go.

Also, how's this looking in mobile?

AlfHou:
@ferferga Hmm, I think it looks alright on mobile in landscape[0], but it doesn't look quite right in portrait[1]. I'll try a fix for that a little later today.
[0]: image
[1]: 
image

Extarys:
@AlfHou 
I think on landscape (and portrait) the card should be lower. [0] is ok but some like me sometimes watch the outro to find a specific name and this position could prevent that. Making it lower would also obstruct the view a little but the main content would be more visible.

Also some animes/animation show have splitted episodes so this popup could appear while there is still some action going on screen if there is no outro.

It's a small detail that could also be addressed later.

AlfHou:
@Extarys @ferferga I think it looks better on mobile now! Please let me know what you think 😄 
image

image

Perseusdehond:
If you don't mind me giving some feedback too than I think that the up next window in portrait mode looks way better, but in landscape mode there is just too much space left unused (to the right) and the popup window takes almost half of the screen space. I would suggest making the text space longer in landscape mode so it takes up less of the bottom part of the screen, because this looks a bit weird. You could even let the text be shown in the same row (to the left) as where the buttons are rn, because there is also too much unused space imo. 

Overall I think the up next window looks a bit basic, but nothing wrong with that necessarily. I think it looks good enough for the most part. But I would like to see things like showing pictures of the next episodes.

Maybe this functionality can also be expanded to movies/end of a series etc. where it shows you 1 to 3 movies you would possibly want to watch. Like the way how Netflix (shows 3) and Disney+ (shows 1) do this. I know that that is way beyond the scope of this PR, but something I would like to see maybe (although we already have such a feature, but not represent in a way D+/Netflix do this totally)

Sorry for that whole idea, but good job so far. Its already great that this feature is almost done :)

AlfHou:
@Perseusdehond 
Thank you for the feedback. The more feedback the better 😃 

>[...] in landscape mode there is just too much space left unused (to the right) and the popup window takes almost half of the screen space.

I agree. I may have been a little fast on the trigger with pushing it. I'll try to change it!


> Overall I think the up next window looks a bit basic, but nothing wrong with that necessarily. I think it looks good enough for the most part. But I would like to see things like showing pictures of the next episodes.

Hmm, I see what you mean, however I took quite a lot of inspiration from the new up next component in [jellyfin-web 10.7](https://github.com/jellyfin/jellyfin-web/pull/1825). There was also some talk about having a setting for making the component even smaller  #158. I think it is fine for now as a starting point.

> Maybe this functionality can also be expanded to movies/end of a series etc. where it shows you 1 to 3 movies you would possibly want to watch. Like the way how Netflix (shows 3) and Disney+ (shows 1) do this. I know that that is way beyond the scope of this PR, but something I would like to see maybe (although we already have such a feature, but not represent in a way D+/Netflix do this totally)

I think it is a nice idea, however as you said, it is a little bit out of scope for this PR I think. Feel free to start a discussion for it tho 😄

Perseusdehond:
I've made a redesign of how I would like this feature to look. You can find it in this discussion https://github.com/jellyfin/jellyfin-vue/discussions/731 . I made this by no means of hating on what you've already did, but maybe you can get some inspiration out of it to make it look more modern :)

Camc314:
@AlfHou Let us know, do you want to work on it a bit more and incorporate Perseusdehond's design, or have it merged in the next few days after some reviews?

AlfHou:
Hmm, no I'd rather have it merged first and then I can iterate on it

components/Players/PlayerManager.vue
```diff
@@ -193,13 +200,45 @@ export default Vue.extend({
...
   },
   created() {
     this.$store.subscribe((mutation, state: AppState) => {
       switch (mutation.type) {
         case 'playbackManager/INCREASE_QUEUE_INDEX':
+          this.upNextUserHidden = false;
+          break;
         case 'playbackManager/DECREASE_QUEUE_INDEX':
+          this.upNextUserHidden = false;
+          break;
```
Camc314, line 241: I don't think we should break here?

components/Players/PlayerManager.vue
```diff
@@ -20,6 +20,11 @@
       <v-hover v-slot="{ hover }">
         <v-card class="player-card" width="100%">
           <video-player />
+          <up-next
+            v-if="showUpNext"
+            :time-left="timeLeft"
+            @hide="upNextUserHidden = true"
```
Camc314, line 26: Why not use
`v-is="showUpNext && !upNextUserHidden"`

Camc314: You would have to use $emit in `UpNext` component to tell the playerManager to hide

components/Item/UpNext.vue
```diff
@@ -0,0 +1,131 @@
...
+  position: fixed;
+  right: 0;
+  bottom: 0;
+  width: 100%;
+  padding: 1em;
+  margin: 0;
+  display: flex;
+  flex-direction: column;
+  will-change: transform, opacity;
+  transition: opacity 300ms ease-out;
```
Camc314, line 100: You could use v-fade-transition instead?

components/Item/UpNext.vue
```diff
@@ -0,0 +1,131 @@
...
+
+.container {
+  position: fixed;
+  right: 0;
+  bottom: 0;
+  width: 100%;
+  padding: 1em;
+  margin: 0;
+  display: flex;
+  flex-direction: column;
```
Camc314, line 98: try to use vuetify display styles instead

https://vuetifyjs.com/en/styles/display/

components/Item/UpNext.vue
```diff
@@ -0,0 +1,131 @@
...
+  -webkit-touch-callout: none;
+}
+
+@media #{map-get($display-breakpoints, 'md-and-up')} {
+  .container {
+    width: 30em;
+    margin: 0 2em 6em 0;
+  }
+}
+.countdownHeader {
```
Camc314, line 114:
```suggestion
.countdown-header {
```

components/Item/UpNext.vue
```diff
@@ -0,0 +1,131 @@
...
+    timeLeft: {
+      type: Number,
+      required: true
+    }
+  },
+  data() {
+    return {
+      source: ''
+    };
+  },
```
Camc314, line 53:
```suggestion
```

components/Players/PlayerManager.vue
AlfHou: Wouldn't it fall through if not? ESLint complains if i remove the break: `Expected a 'break' statement before 'case'.`

components/Players/PlayerManager.vue
AlfHou: `@hide` is an event emitted from the component which sets the `upNextUserHidden` variable to true, which basically means that the user want the component to be hidden until a new episode is shown. Might be a little confusing variable name tho, I'll try to thing of something better

AlfHou (commented):
Can't delete. But this was related to emitting a `startNext` event instead of setting the next track in the `UpNext` component. It can be resolved/approved

components/Players/PlayerManager.vue
AlfHou: Haven't gotten any comments on this. Marking as resolved

components/Players/PlayerManager.vue
AlfHou: Haven't gotten any comments on this. Marking as resolved.

components/Item/UpNext.vue
```diff
@@ -0,0 +1,118 @@
...
+        </v-card-title>
+        <v-card-subtitle class="title subtitle-1 pa-0">
+          {{ nextSeriesName }} -
+          {{
+            $t('tvShowAbbrev', {
+              seasonNumber: nextSeasonNumber,
+              episodeNumber: nextEpisodeNumber
+            })
+          }}
+          - <span v-if="$vuetify.breakpoint.xsOnly"> <br /> </span>
```
MrTimscampi, line 19: This seems a bit messy. For one, you could probably move the v-if to the `<br />` tag, but also with this line like this, the mobile view would have a dangling `-` at the end of the line.

components/Item/UpNext.vue
```diff
@@ -0,0 +1,118 @@
...
+      return this.getRuntimeTime(this.getNextItem?.RunTimeTicks);
+    },
+    nextEndsAt(): string {
+      const seconds =
+        this.ticksToMs(this.getNextItem?.RunTimeTicks) + this.timeLeft * 1000;
+      return this.$dateFns.format(Date.now() + seconds, 'p');
+    },
+    nextName(): string {
+      return this.getNextItem?.Name;
+    }
```
MrTimscampi, line 78: This entire section is a bit unclean, imo.

It'd be better (and probably a bit more optimized, since each compute creates a new reactive thing) to just use getNextItem with the property right after it.

store/playbackManager.ts
```diff
@@ -120,6 +120,13 @@ export const getters: GetterTree<PlaybackManagerState, PlaybackManagerState> = {
       return state.queue?.[state.currentItemIndex]?.MediaType;
     }
     return null;
+  },
+  getCurrentTime: (state) => {
+    if (state.currentTime !== null) {
+      return state.currentTime;
+    } else {
+      return null;
+    }
```
ThibaultNocchi, line 129: Why that if? :)

locales/en-US.json
```diff
@@ -210,6 +217,7 @@
   "save": "Save",
   "saved": "Saved",
   "search": "Search",
+  "seconds": "seconds",
```
ThibaultNocchi, line 220: Shouldn't we keep it as "Seconds" and use a `toLowerCase` in the sentence?

locales/en-US.json
ThibaultNocchi: What's the reason for keeping everything capitalized? :)

components/Item/UpNext.vue
```diff
@@ -0,0 +1,118 @@
...
+  margin: 0.25em 0;
+  font-weight: 500;
+}
+.title {
+  width: 100%;
+  margin: 0.25em 0 0.5em;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
```
MrTimscampi, line 117: `.countdown-header` and `.title` should preferably use the utility classes provided by Vuetify. Both width, margins, font-weight and ellipsis are provided.

components/Item/UpNext.vue
```diff
@@ -0,0 +1,118 @@
...
+@import '~vuetify/src/styles/styles.sass';
+
+.container {
+  position: fixed;
+  right: 0;
+  bottom: 0;
+  width: 100%;
+  will-change: transform, opacity;
+  background-color: rgba(0, 0, 0, 0.7);
+  color: #fff;
```
MrTimscampi, line 95: These should preferably be using Vuetify classes as well, or at least be tied to Vuetify variables for future theming.

components/Players/PlayerManager.vue
ThibaultNocchi: It's definitely not the same logic as before the PR though. Currently both `INCREASE_QUEUE_INDEX` and `DECREASE_QUEUE_INDEX` finally executed the code of the case `SET_CURRENT_ITEM_INDEX`, and now it won't. And yeah, it's good practice to keep the break when you have logic in your cases. So I'd say the best thing is to factorize somewhere the code of `SET_CURRENT_ITEM_INDEX` and call it in those 3 cases. :)

components/Players/PlayerManager.vue
```diff
@@ -195,6 +203,34 @@ export default Vue.extend({
         default:
           return this.getCurrentItem.Name;
       }
+    },
+    timeLeft(): number {
+      return parseInt((this.mediaDuration - this.getCurrentTime).toFixed());
+    },
+    mediaDuration(): number {
```
MrTimscampi, line 210: This should probably be a getter on the playbackManager store instead, since it's something we're bound to re-use (In fact, we already have something similar for the duration in the music controls, only with added formatting).

components/Players/PlayerManager.vue
```diff
@@ -195,6 +203,34 @@ export default Vue.extend({
         default:
           return this.getCurrentItem.Name;
       }
+    },
+    timeLeft(): number {
```
MrTimscampi, line 207: Similarly to the item duration, this should be a getter, since we're bound to re-use it at some point (Example: if we want to add a preference to show remaining time instead of total duration, like some players do).

components/Players/PlayerManager.vue
```diff
@@ -195,6 +203,34 @@ export default Vue.extend({
...
+    },
+    mediaDuration(): number {
+      // In seconds
+      return this.ticksToMs(this.getCurrentItem.RunTimeTicks) / 1000;
+    },
+    showUpNext(): boolean {
+      if (
+        this.isMinimized ||
+        this.upNextUserHidden ||
+        this.getCurrentItem.Type !== 'Episode' ||
```
MrTimscampi, line 218: What if I'm playing multiple movies (For example, if I'm watching Back to the Future or Lord of the Rings) ? It'd be good for this to be modular enough to support all media types, just like the audio player isn't inherently tied to music.

components/Item/UpNext.vue
```diff
@@ -0,0 +1,118 @@
...
+      const seconds =
+        this.ticksToMs(this.getNextItem?.RunTimeTicks) + this.timeLeft * 1000;
+      return this.$dateFns.format(Date.now() + seconds, 'p');
+    },
+    nextName(): string {
+      return this.getNextItem?.Name;
+    }
+  },
+  methods: {
+    ...mapActions('playbackManager', ['setNextTrack'])
```
ThibaultNocchi, line 81: There's no call of this, maybe it can be removed? :)

components/Item/UpNext.vue
```diff
@@ -0,0 +1,118 @@
...
+      required: true
+    }
+  },
+  data() {
+    return {
+      source: ''
+    };
+  },
+  computed: {
+    ...mapState('playbackManager', ['queue', 'currentItemIndex']),
```
ThibaultNocchi, line 57: I can't find calls to this state, maybe it isn't used?

components/Item/UpNext.vue
ThibaultNocchi: I agree, is it necessary? :)

ThibaultNocchi (changes requested):

components/Players/PlayerManager.vue
```diff
@@ -195,6 +203,34 @@ export default Vue.extend({
...
+        !this.getNextItem
+      ) {
+        return false;
+      }
+      // How much of the video to go through before showing
+      // the 'up-next' component
+      const showAtProgressPercentage = 0.97;
```
MrTimscampi, line 225: Percentages for this kind of thing don't really work well. It should use a fixed duration.

Some examples:

* For a 2h video, this would be ~3min30
* For a 10min show, this would be ~18 seconds
* For a 5h video, this would be ~9min

While not as extreme as the same issues we've had on in general with Audiobooks progress using percentages to mark things as played, this doesn't scale very well, imo.

We could probably use like 30 seconds for now, then once we have proper intro/outro detection, rework this to also take that into account.

components/Players/PlayerManager.vue
MrTimscampi: This indeed breaks playback reporting when changing tracks.

locales/en-US.json
MrTimscampi: > What's the reason for keeping everything capitalized? :)

Weird English conventions that are not even really standard (It depends on the style guide used).

locales/en-US.json
MrTimscampi: Agreed, no need to keep another instance of an identical string just for capitalization.

store/playbackManager.ts
```diff
@@ -120,6 +120,13 @@ export const getters: GetterTree<PlaybackManagerState, PlaybackManagerState> = {
       return state.queue?.[state.currentItemIndex]?.MediaType;
     }
     return null;
+  },
+  getCurrentTime: (state) => {
```
MrTimscampi, line 124: Getters should perform some kind of processing on the state to return a different value (For example filtering data or merging two stores' data).

This just returns the state as-is, so it's not necessary. You can just access the state directly in these cases.

MrTimscampi (changes requested):
Need a functional test as well, but here are some thoughts about the implementation for now.

components/Players/PlayerManager.vue
ThibaultNocchi: Do we have queues for movies?

components/Players/PlayerManager.vue
ThibaultNocchi: To me 3min30 in a 2h media doesn't seem too far fetched, as it's generally a movie with credits.

components/Players/PlayerManager.vue
MrTimscampi: Currently, everything has a queue.

In jf-web, you couldn't really make a queue with movies, I think (short of making a playlist), but with the floating player in this client, it's perfectly possible to launch playback, then queue other movies after it.

components/Players/PlayerManager.vue
MrTimscampi: For a miniseries (Battlestar Galactica for example) you'd probably be eating a few of the final minutes though.

components/Players/PlayerManager.vue
ThibaultNocchi: Yeah ofc, I should've known better how I've played with it haha

components/Players/PlayerManager.vue
ThibaultNocchi: Imo there's no universal solution until we get the nice intro/outro detection system. So I'd say we shouldn't worry too much, especially when the "next up" prompt is discardable. So either a fixed timing or a percentage is fine to me

Extarys: The Kodi Next up plugin have configurable options. It shows the popup 45 seconds before the end on a 45 minute video for example. there is also the option for <=15 minute videos and 1h30+ videos (not sure but you get the idea) so it's more tunable depending on the "general context".

MrTimscampi: That seems like a good compromise, yeah.

We can set defaults based on these lengths now, then provide configuration when we get to config pages.

components/Item/UpNext.vue
AlfHou: Got rid of the dangling `-` for the mobile view, however it still looks just as messy: ` <span v-if="$vuetify.breakpoint.smAndUp"> - </span> <br v-else />`

Do you have a suggestion for a cleaner way to write it?

locales/en-US.json
AlfHou: > Weird English conventions that are not even really standard (It depends on the style guide used).
So is the consensus that it should be lowercased?

client/locales/en-US.json
```diff
@@ -31,6 +31,13 @@
   "darkModeToggle": "Toggle dark mode",
   "dateAdded": "Date added",
   "details": "Details",
+  "dialog": {
+    "upNext": {
+      "nextEpisodePlayingIn": "Next Episode Playing in",
```
neopc10, line 36: I recommend "Next episode plays in"

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,86 @@
...
+.container {
+  position: fixed;
+  right: 0;
+  bottom: 0;
+  width: 100%;
+  will-change: transform, opacity;
+  background-color: rgba(0, 0, 0, 0.7);
+  user-select: none;
+  z-index: 6;
+  -webkit-touch-callout: none;
```
dkanada, line 77: Don't we want to avoid browser-specific CSS classes in this codebase? Thought there was a library to handle them automatically.

client/locales/en-US.json
AlfHou: Good point, I think I can use `getCurrentItem.Type` to change the text depending on media type

components/Players/PlayerManager.vue
AlfHou: > Currently, everything has a queue.
Should we just show the dialog every time?

components/Players/PlayerManager.vue
AlfHou: Sure, would it be best to return it in milliseconds or ticks?

client/components/Item/UpNext.vue
MrTimscampi: This property is different, it's not standard and isn't on a standards track, but is sort of required to make iOS Safari looks nicer in some cases.

MrTimscampi: Otherwise, yes, we have autoprefixer by default, so anything that has a standard property will get vendor prefixes if necessary.

dkanada:
@AlfHou any news on this one? I only see three unresolved comments and some merge conflicts, but this pull request looks about ready to merge otherwise :)

AlfHou:
@dkanada Hi, thanks for following up :) 
I'm not sure how to mark the comments as resolved. I have been a little busy with school lately, but I think this is more or less ready to be merged (I just need to resolve the merge conflicts).

The only part I'm a little unsure about how to resolve is @MrTimscampi comment:

> What if I'm playing multiple movies (For example, if I'm watching Back to the Future or Lord of the Rings) ? It'd be good for this to be modular enough to support all media types, just like the audio player isn't inherently tied to music.

Do you want me to change it so the dialog is shown for all media types? Maybe with a condition that: as long as there is some media left in the playlist. show the dialog, disregarding whether it's an episode or a movie.

MrTimscampi:
> Do you want me to change it so the dialog is shown for all media types?

Probably only show it for videos (You can look at the player to see how to determine this). Aside from that, we can probably merge it as-is (after it's rebased and the conflicts are fixed)

AlfHou:
I believe this is ready to be merged now :)

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
...
+<style lang="scss" scoped>
+@import '~vuetify/src/styles/styles.sass';
+
+.container {
+  position: fixed;
+  right: 0;
+  bottom: 0;
+  width: 100%;
+  will-change: transform, opacity;
+  background-color: rgba(map-get($shades, 'black'), 0.7);
```
MrTimscampi, line 83: This is far from ideal. Not only is it using a fixed color (not used for any of the other OSD elements) but it also makes one of the buttons nearly invisible ("Hide"):

image

`v-card` also had a `color` attribute to do this, so using CSS isn't needed. Imo, simply using the `dark` prop is fine here (And even then, it's debatable if the player should be forced to dark mode or not)..

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
...
+  user-select: none;
+  z-index: 6;
+  -webkit-touch-callout: none;
+}
+
+@media #{map-get($display-breakpoints, 'md-and-up')} {
+  .container {
```
MrTimscampi, line 90: Not sure about this, especially when there is a grid system that can allow you to position this properly on all screen sizes, with consistent margins.

The playback information popup uses an absolutely positioned `v-container` (To act as an overlay), with disabled pointer events (To allow interaction with the OSD) and positions the `v-card` using the grid. That's a better, more consistent solution, I think.

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
...
+    timeLeft: {
+      type: Number,
+      required: true
+    }
+  },
+  computed: {
+    ...mapGetters('playbackManager', ['getNextItem', 'getCurrentItem'])
+  },
+  methods: {
+    nextEndsAt(runtimeTicks: number): string {
```
MrTimscampi, line 66: Use `getEndsAtTime` in the `timeUtils` mixin. It already takes the number of ticks.

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
...
+
+<script lang="ts">
+import Vue from 'vue';
+import { mapGetters } from 'vuex';
+import timeUtils from '~/mixins/timeUtils';
+
+export default Vue.extend({
+  mixins: [timeUtils],
+  props: {
+    timeLeft: {
```
MrTimscampi, line 57: Why bring it as a prop when you can map the state here?

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+  <v-fade-transition>
+    <v-card class="container white--text pa-4">
+      <div class="d-flex flex-column flex-grow-1">
+        <v-card-title class="text-h6 pa-0 my-1 mx-0">
+          {{ $t('dialog.upNext.nextItemPlayingIn') }}
+          <span class="primary--text darken-2">
+            &ensp;{{ timeLeft }} {{ $t('seconds').toLowerCase() }}
```
MrTimscampi, line 8: This isn't properly localized.

Use something like `$tc('units.time.seconds', { value: timeLeft })`

Not only does this allow you to use [pluralization](https://kazupon.github.io/vue-i18n/guide/pluralization.html) properly, it also allows translator to move the unit if they need to (Currently, you force the unit to be after the number, which may not always be the case).

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+  <v-fade-transition>
+    <v-card class="container white--text pa-4">
+      <div class="d-flex flex-column flex-grow-1">
+        <v-card-title class="text-h6 pa-0 my-1 mx-0">
+          {{ $t('dialog.upNext.nextItemPlayingIn') }}
```
MrTimscampi, line 6: I'm not sure this is properly translated either.

Different languages have different orders for words. So while English may be "The next item is playing in 15 seconds", another language might be equivalent to "15 seconds until the next item".

A good way to circumvent this while still using a separate "units" translation is to chain translations:
```typescript
$t('dialog.upNext.nextItemPlayingIn', { seconds: $tc('units.time.seconds', { value: timeLeft }))
```

There might be a nicer way to do this, but this avoids duplicating a translation for units.

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
...
+});
+</script>
+<style lang="scss" scoped>
+@import '~vuetify/src/styles/styles.sass';
+
+.container {
+  position: fixed;
+  right: 0;
+  bottom: 0;
+  width: 100%;
```
MrTimscampi, line 81: There's a prop on `v-card` for this.

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
...
+            $t('seasonEpisodeAbbrev', {
+              seasonNumber: getNextItem.ParentIndexNumber,
+              episodeNumber: getNextItem.IndexNumber
+            })
+          }}
+          <span v-if="$vuetify.breakpoint.smAndUp"> - </span> <br v-else />
+          {{ getNextItem.Name }}
+        </v-card-subtitle>
+        <v-card-subtitle
+          v-if="getCurrentItem.Type === 'Movie'"
```
MrTimscampi, line 26: Could probably be a `v-else`? This way videos in Photo libraries also work.

client/components/Item/UpNext.vue
```diff
@@ -0,0 +1,95 @@
...
+        </v-card-subtitle>
+        <div>
+          {{ getRuntimeTime(getNextItem.RunTimeTicks) }}
+          <span class="pl-4"
+            >{{ $t('endsAt', { time: nextEndsAt(getNextItem.RunTimeTicks) }) }}
+          </span>
+        </div>
+        <v-card-actions>
+          <v-spacer />
+          <v-btn class="primary darken-2" @click="$emit('startNext')">
```
MrTimscampi, line 39: Prefer kebab-case for event names (It's the recommended convention for Vue 2).

But also, why send an event upwards? You could simply map the action too and call it directly.

client/components/Players/PlayerManager.vue
```diff
@@ -295,6 +340,125 @@ export default Vue.extend({
             window.addEventListener('keydown', this.handleKeyPress);
           }
 
+          break;
+        case 'playbackManager/INCREASE_QUEUE_INDEX':
+        case 'playbackManager/DECREASE_QUEUE_INDEX':
+        case 'playbackManager/SET_CURRENT_ITEM_INDEX':
+          this.upNextUserHidden = false;
+
```
MrTimscampi, line 348: Anything past this line for the subscription was moved to a Vuex plugin, so it shouldn't be there.

client/components/Players/PlayerManager.vue
```diff
@@ -295,6 +340,125 @@ export default Vue.extend({
             window.addEventListener('keydown', this.handleKeyPress);
           }
 
+          break;
+        case 'playbackManager/INCREASE_QUEUE_INDEX':
+        case 'playbackManager/DECREASE_QUEUE_INDEX':
+        case 'playbackManager/SET_CURRENT_ITEM_INDEX':
+          this.upNextUserHidden = false;
```
MrTimscampi, line 347: No need to subscribe for this, you can simply watch the current item index. If it changes, reset this to false.

client/components/Players/ShakaPlayer.vue
```diff
@@ -40,7 +40,8 @@ export default Vue.extend({
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       player: null as any,
       // eslint-disable-next-line @typescript-eslint/no-empty-function
-      unsubscribe(): void {}
+      unsubscribe(): void {},
+      mediaDuration: 0
```
MrTimscampi, line 44: This seems unused.

client/components/Players/ShakaPlayer.vue
```diff
@@ -51,7 +52,8 @@ export default Vue.extend({
     ...mapState('playbackManager', [
       'lastProgressUpdate',
       'currentTime',
-      'currentVolume'
+      'currentVolume',
+      'isMinimized'
```
MrTimscampi, line 56: Same here.

client/locales/en-US.json
```diff
@@ -32,6 +32,13 @@
   "darkModeToggle": "Toggle dark mode",
   "dateAdded": "Date added",
   "details": "Details",
+  "dialog": {
+    "upNext": {
+      "nextItemPlayingIn": "Next Item Playing in",
```
MrTimscampi, line 37: Avoid Title Case.

"Next item" is also kind of bad. The sentence should indicate *which* kind of item is playing. So something like "Next episode in", "Next movie in", "Next video in".  
We know it'll be playing, so I don't think "playing" is needed.

client/locales/en-US.json
```diff
@@ -357,6 +365,7 @@
   "sortTitle": "Sort title",
   "specialFeatures": "Special Features",
   "speed": "Speed",
+  "startNow": "Start now",
```
MrTimscampi, line 368: "Play now" would be a better way of saying this, I think.

client/store/playbackManager.ts
```diff
@@ -140,6 +141,32 @@ export const getters: GetterTree<PlaybackManagerState, RootState> = {
         return stream.Type === 'Subtitle';
       });
     }
+  },
+  getCurrentTime: (state) => {
+    return state.currentTime;
```
MrTimscampi, line 146: No need for a getter that does no processing. Simply mapState for `currentTime`.

client/store/playbackManager.ts
```diff
@@ -140,6 +141,32 @@ export const getters: GetterTree<PlaybackManagerState, RootState> = {
...
+  },
+  getDuration: (state, _getters, _rootState, rootGetters) => {
```
MrTimscampi, line 148: This shouldn't be a thing, imo.

You can already get the current item pretty easily, there's no need for getters to get properties of the current item, it's only adding more API surface for no real benefit.

client/store/playbackManager.ts
```diff
@@ -140,6 +141,32 @@ export const getters: GetterTree<PlaybackManagerState, RootState> = {
...
+      const currentItem = rootGetters['items/getItem'](
+        state.queue?.[state.currentItemIndex]
+      );
+
+      return ticksToMs(currentItem?.RunTimeTicks) / 1000;
+    }
+
+    return null;
+  },
+  getTimeLeft: (state, _getters, _rootState, rootGetters) => {
```
MrTimscampi, line 159: Not sure about this either. It feels more like a computed property where you'd use it than something in the state itself.

MrTimscampi (changes requested):
Sorry for asking for more changes :/

locales/en-US.json
neopc10: "Next episode plays in..."

client/components/Item/UpNext.vue
AlfHou: `PlaybackManager` uses is to decided whether to show the component or not.

MrTimscampi (approved):

MrTimscampi:
@AlfHou I ended up going in and fixing most of the points myself, since I'd like to get it merged :)

Thanks for the PR and sorry for the time it took to get it in.

AlfHou:
@MrTimscampi Yeah, sorry I have been pretty inactive lately... Hopefully I will be a better contributor next time.

Thank you for all the valuable feedback :)

Extarys:
@AlfHou You have no obligation what so ever. you do your best and that's enough :heart: Thanks for this.

PR #351, opened 2020-12-07
feat(activity.vue): activity logs page

sarab97:
Activity Logs vue page
image


MrTimscampi:
Quick thoughts from the screenshot: the page seems a bit empty. I think we could likely merge it with the Logs pages into a "Logs & Activity".

The layout could be as such:

**Desktop**
```
| Logs   | Activity  |
| Log1   | Activity1 |
| Log2   | Activity2 |
| Log3   | Activity3 |
```
**Mobile**
```
| Logs      |
| Logs1     |
| Logs2     |
| Logs3     |
| Activity  |
| Activity1 |
| Activity2 |
| Activity3 |
```

Essentially echoing the layout of the main settings page, while making the desktop version more full.

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
...
+                <v-list-item-subtitle v-text="activity.ShortOverview" />
+              </v-list-item-content>
+              <v-list-item-action>
+                <v-list-item-subtitle v-text="activity.Date" />
+              </v-list-item-action>
+            </v-list-item>
+          </v-list-item-group>
+        </v-list>
+        <!-- Mobile links -->
+        <v-list v-if="$vuetify.breakpoint.mobile">
```
MrTimscampi, line 28: This isn't needed here :p

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <v-container>
+    <v-row>
+      <v-col cols="12" :offset-md="1" md="5" class="pt-0 pb-4">
+        <!-- Administrator settings -->
```
MrTimscampi, line 5: You can remove the comments here :)

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
...
+        <v-list
+          v-if="$auth.user.Policy.IsAdministrator"
```
MrTimscampi, line 7: This should probably be handled via a [router middleware](https://nuxtjs.org/examples/middleware-router) instead. It'd check this same policy, and redirect to the home page if the user is trying to access an admin page without being an admin.

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
...
+    return {
+      activityList: {} as SystemInfo
+    };
+  },
+  async beforeMount() {
+    this.setAppBarOpacity({ opaqueAppBar: true });
+    this.setPageTitle({ title: this.$t('settings') });
+
+    if (this.$auth.user.Policy.IsAdministrator) {
+      this.activityList = (await this.$api.system.getActivityLogs()).data.Items;
```
MrTimscampi, line 71: I'm unable to find `this.$api.system.getActivityLogs()` in the API. The proper function seems to be `this.$api.activityLog.getLogEntries()`.

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
...
+    }
+  },
+  methods: {
+    ...mapActions('page', ['setPageTitle', 'setAppBarOpacity']),
```
MrTimscampi, line 75:
```suggestion
    ...mapActions('page', ['setPageTitle']),
```

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
...
+    isEmpty(object: any): boolean {
```
MrTimscampi, line 76: You don't need this :)

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
...
+import Vue from 'vue';
+import { mapActions } from 'vuex';
+import { SystemInfo } from '~/api';
+import htmlHelper from '~/mixins/htmlHelper';
+
+export default Vue.extend({
+  mixins: [htmlHelper],
+  data() {
+    return {
+      activityList: {} as SystemInfo
```
MrTimscampi, line 63:
```suggestion
      activityList: [] as ActivityLogEntry[]
```

Your type here is `ActivityLogEntry` and the API returns an array of them.

pages/settings/activity.vue
```diff
@@ -0,0 +1,81 @@
...
+        <v-list
+          v-if="$auth.user.Policy.IsAdministrator"
+          two-line
+          class="mb-4"
+          disabled
+        >
+          <v-list-item-group>
+            <v-list-item v-for="activity in activityList" :key="activity.Id">
+              <v-list-item-avatar>
+                <v-icon>mdi-shield-account-outline</v-icon>
```
MrTimscampi, line 15: This should probably vary depending on `activity.Type` and `activity.Severity` (Change the icon depending on the type of activity and have it in a `v-avatar` or something, that changes color based on severity)

MrTimscampi (commented):
Some more early comments :)

pages/settings/activity.vue
sarab97: this was still in initial stage so a lot of extras and wrongs are there :P. I just opened PR to reserve this feature. 
I should have focused more on code to remove all these.

pages/settings/activity.vue
sarab97: that was just test from my side. Should have been more careful before committing. 
I seem to always have commit issue :p

pages/settings/activity.vue
sarab97: removing mapActions give error.
And another thing is this page title where does it go. To me it doesnt seem to work

pages/settings/activity.vue
sarab97: what are probable values for severity and type? I can see severity has 'Trace' and 'Error'. And for type there are many.
How to manage all those ?
should conditions be managed there in html part only or add complex ones in vue code?

pages/settings/activity.vue
sarab97: middleware created :)

sarab97:
> Quick thoughts from the screenshot: the page seems a bit empty. I think we could likely merge it with the Logs pages into a "Logs & Activity".
> 
> The layout could be as such:
> 
> **Desktop**
> 
> ```
> | Logs   | Activity  |
> | Log1   | Activity1 |
> | Log2   | Activity2 |
> | Log3   | Activity3 |
> ```
> 
> **Mobile**
> 
> ```
> | Logs      |
> | Logs1     |
> | Logs2     |
> | Logs3     |
> |-----------|
> | Activity  |
> | Activity1 |
> | Activity2 |
> | Activity3 |
> ```
> 
> Essentially echoing the layout of the main settings page, while making the desktop version more full.


This is what i created and with smaller size log list stays there and activity log goes down just as you mentioned.

image
What you think ?

pages/settings/activity.vue
MrTimscampi: I meant the isEmpty :)

It's basically there to allow us to use Lodash's isEmpty inside of the template, but you're not using it here.

Though you can remove the `setAppBarOpacity` from the line above.

sarab97: done :)

pages/settings/activity.vue
MrTimscampi: For the Type, currently it's just a string and according to server folks, it's non trivial to change, so we can probably forget that for now :)

For the level, you have a type available as part of the API (`LogLevel`), which is the following:
```typescript
export enum LogLevel {
    Trace = 'Trace',
    Debug = 'Debug',
    Information = 'Information',
    Warning = 'Warning',
    Error = 'Error',
    Critical = 'Critical',
    None = 'None'
}
```

You can use a method in your vue component that returns the color from a given severity level. You'd call it like:
```html
<v-avatar
  :color="getColorFromSeverity(activity.Severity)"
>
      <!-- your icon here -->
</v-avatar>
```

Then have a method in the component named like this:
```typescript
getColorFromSeverity(severity: LogLevel): string
```
With a basic switch..case inside of it :)

MrTimscampi:
> What you think ?

Looks good :)

locales/en-US.json
```diff
@@ -1,97 +1,97 @@
 {
-  "actors": "Actors",
```
Camc314, line 2: Please leave the JSON indentation unchanged ;)

pages/settings/activity.vue
```diff
@@ -0,0 +1,155 @@
...
+      </v-col>
+    </v-row>
+  </v-container>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import { mapActions } from 'vuex';
+import colors from 'vuetify/lib/util/colors';
+import { ActivityLogEntry, LogFile, LogLevel } from '~/api';
```
Camc314, line 60:
```suggestion
import { ActivityLogEntry, LogFile, LogLevel } from '@jellyfin/client-axios';

File: pages/settings/activity.vue [hunk 11]
```diff
@@ -0,0 +1,155 @@
...
+    <v-row>
+      <v-col cols="12" :offset-md="1" md="5" class="pt-0 pb-4">
+        <v-list two-line class="mb-2">
+          <v-list-item-group>
+            <v-list-item
+              v-for="file in logFiles"
+              :key="file.Name"
+              :href="getLogFileLink(file.Name)"
+              target="_blank"
+            >
```
Camc314, line 12:
```suggestion
            <v-list-item
              v-for="file in logFiles"
              :key="file.Name"
              :href="getLogFileLink(file.Name)"
              target="_blank"
              rel="noopener"
            >
```

https://web.dev/external-anchors-use-rel-noopener/

pages/settings/activity.vue
```diff
@@ -0,0 +1,155 @@
...
+  },
+  head() {
+    return {
+      title: this.$store.state.page.title
+    };
+  },
+  async beforeMount() {
+    this.setPageTitle({ title: this.$t('activityLogs') });
+    const minDate = new Date();
+    minDate.setTime(minDate.getTime() - 7 * 24 * 60 * 60 * 1000);
```
Camc314, line 80: Probably best to add a comment mentioning what this is calculating

pages/settings/activity.vue
```diff
@@ -0,0 +1,155 @@
...
+    this.activityList =
+      (await this.$api.activityLog.getLogEntries({ minDate })).data.Items || [];
+    this.logFiles = (await this.$api.system.getServerLogs()).data;
+  },
+  methods: {
+    ...mapActions('page', ['setPageTitle']),
+    getColorFromSeverity(severity: LogLevel): string {
```
Camc314, line 87: This is a good case for a computed property

pages/settings/activity.vue
```diff
@@ -0,0 +1,155 @@
...
+      const today = this.$dateFns.endOfToday();
+      const day = this.$dateFns.formatRelative(
+        this.$dateFns.parseJSON(date),
+        today
+      );
+      return day;
+    },
+    getFormattedLogDate(date: Date): string {
+      const day = this.$dateFns.format(date, 'Ppp');
+      return day;
```
Camc314, line 136:
```suggestion
      return this.$dateFns.format(date, 'Ppp');

**Review by Camc314 — COMMENTED** [review]
Looks good so far, just a few comments

File: pages/settings/activity.vue (hunk 13 above)
Comment (sarab97) [reply pages/settings/activity.vue:87]: didnt get you :confused:

Comment (Camc314) [reply pages/settings/activity.vue:87]: Ah ignore me, I didn't realize this function was taking parameters.

Compared to methods:

> Instead of a computed property, we can define the same function as a method. For the end result, the two approaches are indeed exactly the same. However, the difference is that computed properties are cached based on their reactive dependencies. A computed property will only re-evaluate when some of its reactive dependencies have changed. This means as long as message has not changed, multiple access to the reversedMessage computed property will immediately return the previously computed result without having to run the function again.



https://vuejs.org/v2/guide/computed.html

File: pages/settings/activity.vue [hunk 15]
```diff
@@ -0,0 +1,156 @@
...
+  head() {
+    return {
+      title: this.$store.state.page.title
+    };
+  },
+  async beforeMount() {
+    this.setPageTitle({ title: this.$t('activityLogs') });
+    // calculating date of 7 days ago as min date to get activity logs
+    const minDate = new Date();
+    minDate.setTime(minDate.getTime() - 7 * 24 * 60 * 60 * 1000);
```
MrTimscampi, line 82:
```suggestion
    minDate.setTime(minDate.getDate() - 7);
```

This should get you exactly 7 days ago without looking so complicated.

pages/settings/activity.vue
```diff
@@ -0,0 +1,156 @@
...
+      logFiles: [] as LogFile[]
+    };
+  },
+  head() {
+    return {
+      title: this.$store.state.page.title
+    };
+  },
+  async beforeMount() {
+    this.setPageTitle({ title: this.$t('activityLogs') });
```
MrTimscampi, line 79: The page title should probably be "Logs & Activity"

pages/settings/activity.vue
MrTimscampi: This should also probably be a computer property.

Camc314 (approved):

MrTimscampi (changes requested):
Small changes and then it's good

pages/settings/activity.vue
sarab97: this change would be in translations file

MrTimscampi: It should use the `settingsSections.logs.name` from the settings page PR, really (To avoid duplicated translations)

sarab97: done :)

MrTimscampi (approved):

PR #359, opened 2020-12-08
fix(homesection.vue): fixed dash color not following the white mode

ThibaultNocchi:
The dash before "Next up", "Libraries" etc. on the index was always dark and wasn't visible on white mode. It now follows the current vuetify theme.


MrTimscampi:
I think we should use one of the theme's SCSS variables for this, so we can theme it later on: https://vuetifyjs.com/en/features/sass-variables

Using the text-color variable seems to be the best solution, I think (So it'll be the same color as the text right next to it)

MrTimscampi (approved):

PR #364, opened 2020-12-09
feat(api): move to the API client package

MrTimscampi:


Camc314 (approved):
Assuming it builds, it should be fine.

PR #366, opened 2020-12-09
fix(personlist): duplicated key

ferferga:
Cherry-picked from #275, partially fixed on #362 

This fixes errors on movies where the same person plays more than one role

### DIFF
--- a/components/PersonList.vue
+++ b/components/PersonList.vue
@@ -2,8 +2,8 @@
   <v-list color="transparent" two-line>
     <div v-if="items.length > 0">
       <v-list-item
-        v-for="item in items"
-        :key="item.Id"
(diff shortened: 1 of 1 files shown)


components/PersonList.vue
```diff
@@ -2,8 +2,8 @@
   <v-list color="transparent" two-line>
     <div v-if="items.length > 0">
       <v-list-item
-        v-for="item in items"
-        :key="item.Id"
+        v-for="(item, index) in items"
+        :key="index"
```
MrTimscampi, line 6: Could we maybe combine the two approaches?

```suggestion
        :key="`${item.Id}-${index}`"
```

MrTimscampi (approved):

PR #367, opened 2020-12-09
fix(card): make button nuxt-link aware

ferferga:
Cherry-picked from #275

### DIFF
--- a/components/Card.vue
+++ b/components/Card.vue
@@ -53,7 +53,7 @@
           v-if="overlay"
           class="card-overlay d-flex justify-center align-center"
         >
-          <v-btn fab color="primary" :to="`/item/${item.Id}/play`">
+          <v-btn fab color="primary" nuxt :to="`/item/${item.Id}/play`">
             <v-icon size="36">mdi-play</v-icon>
           </v-btn>
         </div>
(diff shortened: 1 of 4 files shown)


components/Card.vue
```diff
@@ -52,6 +52,7 @@
         <div
           v-if="overlay"
           class="card-overlay d-flex justify-center align-center"
+          nuxt
         >
           <v-btn fab color="primary" :to="`/item/${item.Id}/play`">
```
MrTimscampi, line 57:
```suggestion
        >
          <v-btn fab color="primary" nuxt :to="`/item/${item.Id}/play`">
```

`div` doesn't have a nuxt attribute, but v-btn does: https://vuetifyjs.com/en/api/v-btn/#props :)

MrTimscampi (changes requested):

components/Card.vue
ferferga: @mrtimscampi Oops that's what happens when you cherry pick and doesn't pay much attention to the conflicts 😋.

I'll also take some time to check each v-btn and add the nuxt property to the links, as probably there are more than this one.

MrTimscampi (approved):

PR #373, opened 2020-12-10
build(nuxt.config.ts): add options to improve build perfs

ThibaultNocchi:


MrTimscampi (approved):
Note for future us: these options are experimental

PR #377, opened 2020-12-11
feat: display preferences store and dark mode/locale syncing

ThibaultNocchi:
* Dark mode toggle is into its own component
* Added a store to manage the display preferences with default values for dark mode and locale from `nuxt.config.ts`
* Removed displayPreferences from `user.ts` and now using the `CustomPrefs` from the new store
* Removed id from `user.ts` and now using `$auth.user.id`
* When the app has successfully authenticated the user, it retrieves its display preferences
* When logging out, displayPreferences store is reset to default (so dark mode is reset to true)
* Dark mode is now synced to the server under the custom preference 'darkMode'
* Locale is now synced under the custom preference 'locale' and i18n is no longer saving it into a cookie
* If displayPreferences can't be retrieved or updated, a snackbar is displayed
* Fixed snackbar to make it display a previously stored message when first created
* Moved initialization logic of snackbar from `created` to `mounted`
* Fixed logout label in `userbutton` not being reactive to locale change due to the list being defined in `data` and not `computed`

The `updateMethods` object is used by the store to set custom preferences callbacks for each prop. It's usefulness if in the case of properties where other functions need to be called to apply the new state. Dark mode or locale update for instance need this feature as the new value has to be passed to `$vuetify` or `$i18n` to be taken into account.


store/displayPreferences.ts
```diff
@@ -0,0 +1,197 @@
...
+> = {
+  /**
+   * Fetches display preferences, stores them and updates Vuetify dark mode
+   *
+   * @param {any} param0 Vuex
+   * @param {any} param0.commit Vuex commit
+   * @param {any} param0.dispatch Vuex dispatch
+   */
+  async initState({ commit, dispatch }) {
+    const response = await this.$api.displayPreferences.getDisplayPreferences({
```
bbogdanov, line 105: You can try to catch this `api` call cause it can fail and we won't understand it until we open the console.

https://github.com/jellyfin/jellyfin-vue/blob/master/store/user.ts#L43 `loginRequest` is an example of how you can proceed further

ThibaultNocchi: Yeah that's fair enough. What should I do in case of failure? Single snackbar? For now it's not a problem as only the dark mode is used, I could say "Can't load dark mode preferences, applying the default one". But what will it be in the future if more settings are used?

MrTimscampi: Snackbar is probably enough, for now.

Something like "Failed to load display preferences". We should have a copy of the object this call returns as the default state, so we have sensible defaults to play with in case of a failure.

ThibaultNocchi: I like that, I'll do that

components/DarkModeToggle.vue
```diff
@@ -0,0 +1,22 @@
...
+import Vue from 'vue';
+
+export default Vue.extend({
+  computed: {
+    darkMode: {
+      get() {
+        return this.$store.getters['displayPreferences/getDarkMode'];
+      },
+      set(value: boolean) {
+        this.$store.dispatch('displayPreferences/setDarkMode', {
```
bbogdanov, line 15: This one can be set in `mapActions`

components/DarkModeToggle.vue
```diff
@@ -0,0 +1,22 @@
...
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend({
+  computed: {
+    darkMode: {
+      get() {
+        return this.$store.getters['displayPreferences/getDarkMode'];
```
bbogdanov, line 12: This one can be set in `mapGetters`

store/displayPreferences.ts
ThibaultNocchi: Done, there's a default value for dark mode and a check for the status of the answer

bbogdanov: I would encourage you to set a 'throw Error()' inside the 'getDisplayPreferences' method and observe what the action code will do.

ThibaultNocchi: I'd prefer avoiding meddling with the API files as they are not in the codebase anymore. But I can wrap my call and code check in the try

ThibaultNocchi: How do you like what I've done?

store/displayPreferences.ts
```diff
@@ -0,0 +1,224 @@
...
+   */
+  EDIT_CUSTOM_PREF(
+    state: DisplayPreferencesState,
+    { pref }: { pref: SingleCustomPrefMutationPayload }
+  ) {
+    state.CustomPrefs[pref.key] = pref.value;
+    if (pref.key in updateMethods) updateMethods[pref.key](pref.value);
+  },
+
+  CALL_ALL_CALLBACKS(state: DisplayPreferencesState) {
```
bbogdanov, line 101: This one should be an `action` that triggers a mutation multiple times

store/displayPreferences.ts
bbogdanov: Much better! Now we will catch an error response from the server. Can we proceed even further with following the pattern that the other stores used for the API calls? I mean, having a `success` action and `failure` action for the calls. `loginRequest` is still an example of what I am referring to.

components/Snackbar.vue
```diff
@@ -16,13 +16,26 @@ export default Vue.extend({
     };
   },
   created() {
+    // Empties potential already set buffer
+    this.displaySnackbar(
+      this.$store.state.snackbar.message,
```
bbogdanov, line 21: `computed: mapState<AppState>({
   snackBar: (state) => state.snackBar
})`

components/Snackbar.vue
```diff
@@ -16,13 +16,26 @@ export default Vue.extend({
...
       }
     });
+  },
+  methods: {
+    displaySnackbar(message: string, color: string) {
+      if (message !== '') {
+        this.message = message;
+        this.color = color;
+        this.model = true;
+        this.$store.commit('snackbar/RESET_MESSAGE');
```
bbogdanov, line 36: This should be triggered by an `action`.

layouts/default.vue
```diff
@@ -147,6 +144,7 @@ export default Vue.extend({
     }
   },
   beforeMount() {
+    this.$store.commit('displayPreferences/CALL_ALL_CALLBACKS');
```
bbogdanov, line 147: This should be triggered by an `action`

pages/index.vue
```diff
@@ -43,7 +43,7 @@ export default Vue.extend({
 
     // Filter for valid sections in Jellyfin Vue
     let homeSectionsArray = pickBy(
-      this.$store.state.user.displayPreferences,
+      this.$store.getters['displayPreferences/getCustomPrefs'],
```
bbogdanov, line 46: `mapGetters`

store/displayPreferences.ts
```diff
@@ -0,0 +1,224 @@
+import { DisplayPreferencesDto } from '@jellyfin/client-axios';
+import { MutationTree, ActionTree, GetterTree } from 'vuex';
+import nuxtConfig from '~/nuxt.config';
+
+const stringToBoolean = (value: string) => value === 'True';
```
bbogdanov, line 5: Can you please explain what this function aims to solve?

store/displayPreferences.ts
```diff
@@ -0,0 +1,224 @@
...
+    nuxtConfig.vuetify?.theme?.default === 'dark' || false
+  ),
+  locale: nuxtConfig.i18n?.defaultLocale || 'en'
+};
+
+const defaultState = (): DisplayPreferencesState => ({
+  CustomPrefs: defaultCustomPrefs
+});
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
bbogdanov, line 27: Can you please explain why would you need `any`?

components/Snackbar.vue
bbogdanov: `this.snackBar.message, this.snackBar.color`

store/displayPreferences.ts
```diff
@@ -0,0 +1,224 @@
...
+          client: 'vue',
+          displayPreferencesDto: state
+        }
+      );
+      if (response.status !== 204)
+        throw new Error(
+          'set display preferences status response = ' + response.status
+        );
+    } catch (error) {
+      // eslint-disable-next-line no-console
```
bbogdanov, line 170: The `initState` has the same error handling as this one. We can extract this in action and just trigger it in both places.

bbogdanov (changes requested):

store/displayPreferences.ts
ThibaultNocchi: stringToBoolean? The `CustomPrefs` property of the `displayPreferences` as defined by the server is an object of keys to string value. The boolean values are also noted as `True` and `False` (although we could use whatever we want I guess) but in the end it's still an object with string values. This small function is just there to avoid repetition and be easier to edit the logic if we ever change the keyword for `true` and `false` bools.

store/displayPreferences.ts
ThibaultNocchi: I just don't know the type for functions, what could I use?

components/Snackbar.vue
ThibaultNocchi: Done

pages/index.vue
ThibaultNocchi: Done

store/displayPreferences.ts
ThibaultNocchi: Thinking of it, it doesn't even have to be a commit as it doesn't edit the state. Maybe only a dispatch `callAllCallbacks` doing this same loop could do it?

store/displayPreferences.ts
bbogdanov: Ok, so the server returns boolean as a string? This c#.... always something has to be out of the standards :D

components/Snackbar.vue
ThibaultNocchi: `computed: mapState<AppState>({
    snackBar: (state: AppState) => state.snackBar
  })` doesn't work, the `snackBar` computed prop in the component is undefined

store/displayPreferences.ts
bbogdanov: `() => void` for functions that do not return a value 

`() => unknown` for functions that we don't know what they would return

store/displayPreferences.ts
bbogdanov: I believe so. Try it out

components/Snackbar.vue
bbogdanov: Do we actually need that exactly in the `created` lifecycle hook?

store/displayPreferences.ts
ThibaultNocchi: Done

store/displayPreferences.ts
ThibaultNocchi: The `displayPreferences` endpoint returns an object with a mix of different types, but the `CustomPrefs` object in it can only have string values. I think that's because in `CustomPrefs` no key is necessary, and maybe C# doesn't handle non specified objects with multiple types? I don't know in the end haha. Here is an exemple by the doc:

```json
{
  "Id": "string",
  "ViewType": "string",
  "SortBy": "string",
  "IndexBy": "string",
  "RememberIndexing": true,
  "PrimaryImageHeight": 0,
  "PrimaryImageWidth": 0,
  "CustomPrefs": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  },
  "ScrollDirection": "Horizontal",
  "ShowBackdrop": true,
  "RememberSorting": true,
  "SortOrder": "Ascending",
  "ShowSidebar": true,
  "Client": "string"
}
```

components/Snackbar.vue
ThibaultNocchi: Maybe in `mounted` we could do it, I should try.

store/displayPreferences.ts
ThibaultNocchi: Done!

layouts/default.vue
ThibaultNocchi: Done

store/displayPreferences.ts
ThibaultNocchi: I'll just resolve this conversation :)

components/Snackbar.vue
ThibaultNocchi: Done, it works in mounted

store/displayPreferences.ts
ThibaultNocchi: Done

store/displayPreferences.ts
ThibaultNocchi: I don't know about that, they do not act the same. One is for getting data, the other one is just to push. I've put the error handling in its own action, is it enough? :)

components/Snackbar.vue
bbogdanov: You can use ‘mapState’ I guess now?

store/displayPreferences.ts
bbogdanov: What do you mean by ‘they do mot act the same’? One makes the request, one handles the success of this request and one handles the error of it. The main idea is that at some point you would want to know when the preferences are loaded. You can do that by subscribing to the store and listen for the success action instead of ‘try catch’ the ‘initState’

ThibaultNocchi: Ok even though I do feel it's kinda bloaty, I agree with the ease of use to subscribe to the methods. So I've done it with the two methods which query the server. Is it how you were thinking?

bbogdanov (approved):

store/displayPreferences.ts
```diff
@@ -0,0 +1,294 @@
...
+const defaultState = (): DisplayPreferencesState => ({
+  CustomPrefs: defaultCustomPrefs
+});
+
+const updateMethods: { [key: string]: (value: string) => void } = {
+  darkMode: (value: string) => {
+    window.$nuxt.$vuetify.theme.dark = stringToBoolean(value);
+  },
+  locale: (value: string) => {
+    window.$nuxt.$i18n.setLocale(value);
```
Camc314, line 32: Initially, when loading the page, it is calling this before `window.$nuxt` is defined. Probably best to check if `$nuxt` exists before trying to set the relevant mode

locales/en-US.json
```diff
@@ -20,6 +20,8 @@
   "endDate": "End date",
   "endsAt": "Ends at {time}",
   "episodeNumber": "Episode {episodeNumber}",
+  "failedRetrievingDisplayPreferences": "Failed retrieving display preferences, using default ones",
+  "failedSettingDisplayPreferences": "Failed updating display preferences",
```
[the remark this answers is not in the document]
ThibaultNocchi: Done

store/displayPreferences.ts
ThibaultNocchi: Yeah that's right, I put the try catch in the `callAllCallbacks` method but that's stupid if we wanna call them one by one like in `EDIT_PREF`

Camc314: Ah ok, I missed that, that makes sense.

ThibaultNocchi: No worry, I just put individual try catch in the locale and darkMode methods. This way even when we call them individually it catches the error. And this allows us to use "property methods" without a try catch.

components/DarkModeToggle.vue
```diff
@@ -0,0 +1,25 @@
...
+
+export default Vue.extend({
+  computed: {
+    ...mapGetters('displayPreferences', ['getDarkMode']),
+    darkMode: {
+      get() {
+        return this.getDarkMode;
+      },
+      set(value: boolean) {
+        this.setDarkMode({ darkMode: value });
```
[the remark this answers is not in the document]
ThibaultNocchi: You're right, I used this as booleans are stored in `CustomPrefs` in the server are strings, but I didn't want to parse the boolean to string in the component. But I'm making a more general `editBooleanCustomPref` action which then passes the parsed bool to the standard `editCustomPref`.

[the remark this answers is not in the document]
ThibaultNocchi: Ok it's done in the latest commit, there's a wrapper `editBooleanCustomPref` and a getter `getBooleanCustomPref` which returns an anonymous function which you give the key you want parsed to boolean.: )

store/user.ts
```diff
@@ -1,55 +1,30 @@
 import { ActionTree, MutationTree } from 'vuex';
 
 export interface UserState {
-  id: string;
   accessToken: string;
```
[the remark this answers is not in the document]
MrTimscampi: Not really, because the way Nuxt Auth checks for login is by using that $auth store (fetchUser in the scheme stores the UserDto there, which then means we are logged in).

The ideal would be to get rid of the user store completely and get the access token (which is the only thing we use it for, now) as part of the UserDto.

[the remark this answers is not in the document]
Camc314: Ah ok. My only thought was we might get type safety if it was stored in the user store instead of $auth

MrTimscampi (changes requested):
I'm getting two of these every refresh:
image

Should probably add a condition or something.

Otherwise, works well and the code looks good, aside from that one comment.

ThibaultNocchi:
@MrTimscampi Yeah that's because at start when the display preferences are loaded, it tries to update the theme but ofc $nuxt is undefined. In my try-catch I still log the error, but maybe I shouldn't cause the only time it should fail is when the object is undefined. You'll see it here: https://github.com/jellyfin/jellyfin-vue/pull/377/files#diff-81691a4aed89daee2eaa756abc8200bf494d8dae6c6703d9e61c7b0e5e883f86R33

What should I do?

store/displayPreferences.ts
```diff
@@ -0,0 +1,312 @@
...
+  DisplayPreferencesState,
+  DisplayPreferencesState
+> = {
+  /**
+   * Returns the CustomPrefs object
+   *
+   * @param {DisplayPreferencesState} state Current state
+   * @returns {object} CustomPrefs object
+   */
+  getCustomPrefs: (state: DisplayPreferencesState) => state.CustomPrefs,
```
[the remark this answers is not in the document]
ThibaultNocchi: Done, I removed it

store/displayPreferences.ts
```diff
@@ -0,0 +1,304 @@
...
+ * The try-catch are there to silence those errors.
+ */
+const updateMethods: { [key: string]: (value: string) => void } = {
+  darkMode: (value: string) => {
+    try {
+      window.$nuxt.$vuetify.theme.dark = stringToBoolean(value);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
+    }
```
MrTimscampi, line 40:
```suggestion
    if (window.$nuxt) {
      window.$nuxt.$vuetify.theme.dark = stringToBoolean(value);
    }
```

Something like this should fix the issue.

ThibaultNocchi: Done!

MrTimscampi (approved):

PR #382, opened 2020-12-14
Refactor imageHelper mixin and use it everywhere

ferferga:
Made the mixin more modular and reusable.


mixins/imageHelper.ts
```diff
@@ -9,87 +9,166 @@ import { BaseItemDto, ImageType } from '@jellyfin/client-axios';
...
+      type: ImageType,
+      item?: BaseItemDto,
+      element?: HTMLElement,
+      tag?: string,
+      itemId?: string,
+      maxWidth?: number,
+      maxHeight?: number,
+      quality?: number,
+      limitByWidth?: boolean
+    ) => string;
```
Camc314, line 22: Probably better to pass this as an object

ferferga: How exactly I should expose this so suggestions appear in VSCode? I can't get that to work

Camc314:
This should work

```
    getImageUrlForElement({
      type,
      item,
      element,
      tag,
      maxWidth,
      maxHeight,
      quality,
      limitByWidth
    }: {
      type: ImageType;
      item?: BaseItemDto;
      element?: HTMLElement;
      tag?: string;
      maxWidth?: number;
      maxHeight?: number;
      quality?: number;
      limitByWidth?: boolean;
    }): string {
```

image

mixins/imageHelper.ts
Camc314: See comment below

ferferga:
@Camc314 Not sure if JSDoc is in the best format for this kind of argument, but I couldn't find any best practice for writing JSDoc in functions that use dictionaries for its parameters.

Also, I don't know why Sonarcloud is complaining about duplication now, I guess it's caused on the lines where the function is exposed to the interfaces. If someone has an idea for simplifying this, it would be much appreciated.

Camc314:
I don't know whether this would fix the sonarcloud errors, but you could create a type 
```
type imageparams {
      type: ImageType;
      item?: BaseItemDto;
      element?: HTMLElement;
      tag?: string;
      maxWidth?: number;
      maxHeight?: number;
      quality?: number;
      limitByWidth?: boolean;
    }

to reduce some of the duplication

**Comment by ferferga** [comment]
@Camc314 That did the job! Thanks!

File: mixins/imageHelper.ts [hunk 2]
```diff
@@ -7,106 +7,178 @@ import Vue from 'vue';
...
-        } else {
-          type = ImageType.Backdrop;
+    }: getImageUrlForElementParams): string {
+      if (item) {
+        if (!item.ImageTags) {
+          throw new TypeError(
+            'item.ImageTags must not be null or undefined when an item object is passed'
+          );
+        } else if (!tag) {
+          tag = item.ImageTags[type];
```
MrTimscampi, line 124: Why do this instead of having `item?.ImageTags?.[type]` as the default value for the `tag` parameter?

mixins/imageHelper.ts
```diff
@@ -7,106 +7,178 @@ import Vue from 'vue';
...
       return url.toString();
+    },
+    /**
+     * Shorthand for the 'getImageUrlForElement' function when using an itemId only.
+     *
+     * @param {ImageType} type - The type of the image to fetch.
+     * @param {string} id - iitemId to get the image from.
+     * @returns {string} The URL for the image, with the base URL set and the options provided.
+     */
+    getImageUrlById(type: ImageType, id: string): string {
```
MrTimscampi, line 158: This feels superfluous. Let's avoid this kind of aliases and just call the original method directly, please.

mixins/imageHelper.ts
```diff
@@ -7,106 +7,178 @@ import Vue from 'vue';
...
+     * @param {string} id - itemId to get the image from.
+     * @param {HTMLElement} element - The DOM element which size will be used for the image's maximum width or height (optional).
+     * @returns {string} The URL for the image, with the base URL set and the options provided.
+     */
+    getSizedImageUrlById(
+      type: ImageType,
+      id: string,
+      element: HTMLElement
+    ): string {
+      return this.getImageUrlForElement({
```
MrTimscampi, line 177: Same for this, please just call the original one instead of making multiple aliases that all go to the same thing.

MrTimscampi (changes requested):

mixins/imageHelper.ts
ferferga: Fair enough, did these when there were no options array, so most of the calls had a lot of ``undefined`` as passed variables.

mixins/imageHelper.ts
ferferga: @MrTimscampi You just solved a thing that was driving me crazy (literally spend hours understanding why) in this PR and in #394!

I was doing ``tag = item?.ImageTags?[type]``

Thank you very much!

MrTimscampi (approved):

PR #385, opened 2020-12-14
feat(userbutton): image size linked to icon size + default icon in img placeholder

ThibaultNocchi:
Putting the icon in the v-img placeholder makes it appear while the user image is loading (or not at
all) instead of the plain blue circle

Also idk about you, but I got this blue halo around the pic due to the background, I'll try to look into it
print_screen_2020-12-14-10-11-02

### DIFF
--- a/components/UserButton.vue
+++ b/components/UserButton.vue
@@ -2,13 +2,12 @@
   <v-menu offset-y>
     <template #activator="{ on, attrs }">
       <div class="d-flex align-center" v-bind="attrs" v-on="on">
-        <v-avatar size="48" color="primary" class="mr-4">
-          <v-img
-            v-if="userImage"
-            :src="userImage"
-            :alt="$auth.user.Name"
-          ></v-img>
-          <v-icon v-else dark>mdi-account</v-icon>
+        <v-avatar :size="avatarSize" color="primary" class="mr-4">
+          <v-img :src="userImage" :alt="$auth.user.Name">
+            <template #placeholder>
+              <v-icon dark>mdi-account</v-icon>
+            </template>
+          </v-img>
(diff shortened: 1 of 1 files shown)


components/UserButton.vue
```diff
@@ -2,13 +2,12 @@
...
-          <v-img
-            v-if="userImage"
-            :src="userImage"
-            :alt="$auth.user.Name"
-          ></v-img>
-          <v-icon v-else dark>mdi-account</v-icon>
+        <v-avatar :size="avatarSize" color="primary" class="mr-4">
+          <v-img :src="userImage" :alt="$auth.user.Name">
+            <template #placeholder>
+              <v-icon dark>mdi-account</v-icon>
```
Camc314, line 8: Can you confirm that changing from between light/dark theme keeps the icon the same color (white)?

ThibaultNocchi: Yup, the `dark` prop is static and not linked to the `$vuetify.theme.dark` property I've noticed. Here are the pics:
print_screen_2020-12-14-12-02-46
print_screen_2020-12-14-12-02-41

Camc314: Cool Thanks for checking

Camc314 (approved):

components/UserButton.vue
```diff
@@ -51,7 +51,10 @@ export default Vue.extend({
     userImage: {
       get() {
```
MrTimscampi, line 52:
```suggestion
    userImage(): string {
```

This should help with the TypeScript error.

ThibaultNocchi: You're right, I didn't even think of it haha. Thanks!

PR #386, opened 2020-12-14
feat(swiper): story-like progress bars for swiper

ferferga:
* Adds an Spotify's storylines and Instagram Stories look for Swiper

* Reactive to touch event, halting the progress bar when the user is holding or clicking the slide.

gif

## Issues

* Swiper behaviour is somewhat strange at moments: sometimes it pauses the autoplay on interaction and sometimes it doesn't. Also, sometimes it resumes the autoplay where it was left off and sometimes it starts from 0. I'm investigating those but they're not big enough for this to be considered WIP imo.

* I don't understand why the linter is complaining so much :(


Shadowghost:
Really like the idea, but can we add some arrows on the left and right edge of the swiper element? Imho on desktop there should be a clickable element instead of drag and hold to swipe? It's great for mobile devices with a touch screen but not common behaviour on desktop.

ferferga:
Latest commits change a bit the design (as I tested without #381 in): **I moved the progress to the top and I undarkened the center-right part of the images (darker colors, like in my photo, looked washed out)**:

### Before
image
### After
image

However, as you can see, some images really break the visibility of it. I darkened the colors already (screenshot with the new colors) but I would like some input in how to solve this (because maybe somebody has a better idea for positioning it at the bottom, where there is already a gradient). I also tried adding another gradient to the top, but I'm really bad at it and I can't figure out how to make it work in webkit browsers :P.

Dark screenshot for reference:

image

ferferga:
@Shadowghost That's what I'm currently investigating :P. But probably for another PR.

I thought about placing the arrows where the localization icon is located right now, so the image is as clear as possible. 

Doing the arrows on the side might also conflict with the text in the left side.

I also want to implement a11y to swiper, so accessibility is still granted. Would be great if someone who actually uses accessibility stuff can give some input.

components/SwiperProgressBar.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+  <div class="progress-container">
+    <div v-for="i in pages" :key="i" ref="progress" class="progress"></div>
```
Camc314, line 3:
```suggestion
    <div v-for="i in pages" :key="`progress-key-${i}" ref="progress" class="progress"></div>
```

Reduce the chance of conflicting keys

Camc314 (commented):
Looks really nice!

ferferga:
The ``SwiperProgresBar`` component now emits events, which helped me fix the swiper autoplay issue (giving us maximum control of the appearance):

**Check how no matter how much I click to pause the progress, the slide will go in perfect sync, even in the edge, without ``setTimeout`` and the previous dirty hacks:**

gif

**Now, bars are also clickable to move to another slide, which is specially useful for desktop usage:**

gif2

ThibaultNocchi:
Although I don't know how I feel about having this instead of classical dots or small lines below swiper, I do find it elegant and different. For the desktop arrows in the top right, I do agree with it. I just don't know about the potential confusion of these with history navigation arrows.  Maybe we could use backward / forward icons instead of typical arrows?

MrTimscampi (approved):

MrTimscampi:
For a further PR, the filled bars aren't very visible on a the light theme: 
image

PR #394, opened 2020-12-15
feat(swiper): blurhash support for swiper

ferferga:
* Refactored BlurhashImage so it works with any type of image
* Added blurhash to backdrops and logos in Swiper.

gif


components/BlurhashImage.vue
```diff
@@ -59,13 +60,63 @@ export default Vue.extend({
...
+            this.item?.ImageTags,
+            this.type
+          )) ||
+        (this.type === ImageType.Backdrop &&
+          this.item?.BackdropImageTags &&
+          this.item?.BackdropImageTags.length > 0)
+      ) {
+        return true;
+      } else {
+        console.error('Provided tag does not exist in the item');
```
ferferga, line 80: I know we shouldn't use console, but throw halts the client altogether. I don't think silencing them is a good idea either, so what do you think?

MrTimscampi:
I think logos shouldn't have blurhashes, since they always have transparency (Which is known to cause issues with blurhash, since that doesn't support transparency)

ferferga:
@MrTimscampi API-wise they do

components/HomeHeader.vue
```diff
@@ -1,24 +1,24 @@
...
+        :item="getRelatedItem(item)"
+        :type="'Backdrop'"
       />
       <div class="slide-backdrop-overlay" />
       <div class="slide-content">
         <v-container class="mx-10 mt-5">
           <v-row>
             <v-col cols="5">
-              <v-img
+              <blurhash-image
```
MrTimscampi, line 18: Mentioned in my logo PR, but logos shouldn't use blurhash, since they have transparency.

MrTimscampi:
> @MrTimscampi API-wise they do

They do in the API, but that's not a reason to use them :) Transparency with blurhash looks terrible, since it just slaps a white background on things. It's nice for fully opaque images, but we should avoid it in cases where there is transparency, imo.

ferferga:
@MrTimscampi Any specific tag we should ignore aside from Logos so I can add them to my commit?

Camc314:
The only other possible ones are `Art` and `Disk`. I'm not sure where those are used or what they look like so I may be wrong.

components/BlurhashImage.vue
```diff
@@ -59,13 +60,63 @@ export default Vue.extend({
...
+        (this.item?.ImageTags &&
+          Object.prototype.hasOwnProperty.call(
+            this.item?.ImageTags,
+            this.type
+          )) ||
+        (this.type === ImageType.Backdrop &&
+          this.item?.BackdropImageTags &&
+          this.item?.BackdropImageTags.length > 0)
+      ) {
+        return true;
```
Camc314, line 78: This looks very similar to #306. In it the following is addressed:

> Changes
> - Show movie backdrop instead of primary image when in resume home section
> - Show episode primary image when in resume/up next home section
> - Fall back to backdrop if there is no thumbnail

ferferga: Worked on this before knowing anything about #306 

I haven't explained it here so much (mostly in Matrix) but I want to make this component as modular as possible with image types so, whenever we have any item object, we can pass it with an image type confidently, knowing that it will handle all the work for us. We rely too much on the mixin right now (which is still great to have in some cases), but for some basic stuff we can simplify our lives a lot by using a single component.

We can simply create an array of image types that we want to ignore, so we skip the canvas altogether and just keep the image for cases where we pass an image type where we don't want blurhashes.

In a future PR, I'll wrap most images in a LazyImage component that depends on BlurhashImage, so we can have the unloading effect we have in current web without messing too much (see #275). BlurhashImage will be the center of all of it.

Would be great to have all your input on this. cc @MrTimscampi

MrTimscampi:
Trying to test this, getting the following:
image

ferferga:
@MrTimscampi #382 fixes this

components/BlurhashImage.vue
ThibaultNocchi: I've let a few of console.error where it is critical, but @MrTimscampi wanted to look more into it for modal stacktraces. So I think for now console errors are good. You can disable the linter going crazy with `// eslint-disable-next-line no-console` on the line before.

ferferga:
Blocked until #382 is merged

components/HomeHeader.vue
```diff
@@ -124,50 +131,69 @@ export default Vue.extend({
...
-        });
+        continue;
       }
+
+      const itemData = (
+        await this.$api.userLibrary.getItem({
+          userId: this.$auth.user.Id,
+          itemId: id
+        })
+      ).data;
```
MrTimscampi, line 158: We should need this, since we should have everything required in what's returned by the latest endpoint (If we don't, it's better to add the info there (for things like blurhashes and such), imo.

components/HomeHeader.vue
```diff
@@ -1,16 +1,19 @@
...
-        :style="{
-          backgroundImage: `url('${getBackdrop(item)}')`
-        }"
-      />
+      <div class="slide-backdrop">
+        <blurhash-image
+          :key="`${item.Id}-${reloadSentinel}`"
+          :item="getRelatedItem(item)"
+          :type="'Backdrop'"
+        />
```
MrTimscampi, line 15:
```suggestion
        <blurhash-image
          v-if="getRelatedItem(item)"
          :key="`${item.Id}-${reloadSentinel}`"
          :item="getRelatedItem(item)"
          :type="'Backdrop'"
        />
```

This fixes my `this.item cannot be undefined` error, but then some items don't have a backdrop at all:
image

components/HomeHeader.vue
ferferga: @mrtimscampi We don't: what we do here is to get parent items.

This function runs the exact same checks you added for getting the image backdrops: for music, it fetched artist, for series, it fetched series item.

As we now use BlurhashImage, we need to pass the full item object. That's where this function enters the game: it requests the full item object of the items that use the backdrop from the parent.

I store those "Related Items" in a dictionary (key is the index of the element in the this.items array) and value this related item.

components/HomeHeader.vue
ferferga: @Mrtimscampi Now I see the problem.

I only return items that have related items in the function, so items that must display its own background images are completely skipped.

I'm going to push a fix, but it doesn't fully solves the issue, as we don't do any filtering in the initial items to check that everything will display a photo, so items without metadata will still be displayed totally black.

MrTimscampi: >  but it doesn't fully solves the issue, as we don't do any filtering in the initial items to check that everything will display a photo

That's not an issue :) We should show the latest, no matter if it has a backdrop or not. But these had a backdrop and worked before, hence the bug :yum:

components/BlurhashImage.vue
```diff
@@ -57,81 +62,82 @@ export default Vue.extend({
...
+      } else {
+        return ((this.item?.ImageBlurHashes as BaseItemDtoImageBlurHashes)[
+          this.type
+        ] as Record<string, string>)[
+          (this.item?.ImageTags as Record<string, string>)[this.type]
+        ];
+      }
+    },
+    onError(): void {
+      this.pushSnackbarMessage({
```
MrTimscampi, line 136: We shouldn't output a message for every error. Some, like this one, can just deal with an even sent back up.

components/BlurhashImage.vue
```diff
@@ -57,81 +62,82 @@ export default Vue.extend({
...
-          ) {
-            return this.item?.ImageBlurHashes?.Backdrop?.[
-              this.item?.BackdropImageTags?.[0]
-            ];
-          }
-      }
-      return '';
     },
-    checkImage(): boolean {
+    isValidBlurhash(): boolean {
```
MrTimscampi, line 93: This doesn't check if the Blurhash is actually valid, just if it exists in the item.

We should perform a check via Blurhash's `isBlurhashValid` function, too.

components/HomeHeader.vue
```diff
@@ -1,16 +1,19 @@
 <template>
   <swiper
-    v-if="items.length > 0 && !$vuetify.breakpoint.mobile"
+    v-if="items.length > 0 && !$vuetify.breakpoint.mobile && loaded"
```
MrTimscampi, line 3:
```suggestion
    v-if="items.length > 0 && !$vuetify.breakpoint.mobile && !loading"
```

We've been using `loading` everywhere else, it's probably best to keep this consistent.

components/HomeHeader.vue
MrTimscampi: > As we now use BlurhashImage, we need to pass the full item object. That's where this function enters the game: it requests the full item object of the items that use the backdrop from the parent.

I still think we should change the server to include the blurhashes from the parent items as well.

We already get the image tags and item IDs for everything, so it'd make sense to get that as well, to minimize requests.

However, for now, it's necessary. Please add a TODO or HACK mentioning we need to go back to this when the server changes have been made, though.

components/HomeHeader.vue
```diff
@@ -124,50 +131,73 @@ export default Vue.extend({
...
         return '';
       }
+    },
+    getLogo(item: BaseItemDto): string {
+      const relatedItem = this.getRelatedItem(item);
+      return this.getImageUrlForElement(ImageType.Logo, {
+        itemId: relatedItem.Id
+      });
+    },
+    onSlideChange(): void {
```
MrTimscampi, line 185:
```suggestion
    onSlideChange(swiperInstance: Swiper): void {
```

You already get the Swiper instance on all events emitted by Swiper itself, and the Vue library only passes them through. See https://swiperjs.com/api/#events

components/HomeHeader.vue
```diff
@@ -124,50 +131,73 @@ export default Vue.extend({
...
+      const currentIndex = ((this.$refs.homeSwiper as Vue).$swiper as Swiper)
+        .realIndex;
+      if (currentIndex === 0 || currentIndex === this.pages - 1) {
+        this.forceReload();
+      }
+    },
+    // Vue-awesome-swiper seems to have a bug where the components inside of duplicated slides (when loop is enabled,
```
MrTimscampi, line 192:
```suggestion
    // HACK: Vue-awesome-swiper seems to have a bug where the components inside of duplicated slides (when loop is enabled,
```

MrTimscampi: We should also probably report this upstream: https://github.com/surmon-china/vue-awesome-swiper

MrTimscampi (changes requested):

components/BlurhashImage.vue
ferferga: @MrTimscampi That's actually done in the worker with a try/catch (the function performs a halfway decode and it reported false positives for me, so I think doing the decode directly and catching when needing is our best bet)

MrTimscampi: > it reported false positives

This might be a bug in our implementation server-side. I know we at least have a color channel issue, since results are pretty different between our custom C# library and Blurhash's reference one in terms of color.

But that sounds good for now.

components/HomeHeader.vue
ferferga: Sure, will do once this is merged (I think it's better than pointing the original dev to a branch that will be deleted after merging)

components/HomeHeader.vue
ferferga: @MrTimscampi ``swiperInstance`` returns ``undefined``

MrTimscampi (approved):
LGTM :+1:

PR #399, opened 2020-12-16
UI for the music player

MrTimscampi:
This adds a working music UI and a basic audio player using Shaka as a backend (to be replaced with Howler or something else in the future).

Screenshot_2020-12-27 Jellyfin

Screenshot_2020-12-27 Jellyfin(1)


layouts/default.vue
```diff
@@ -106,6 +106,9 @@
     <v-main>
       <nuxt />
     </v-main>
+    <v-footer v-if="false" app>
+      <music-player />
+    </v-footer>
```
Camc314, line 111: This needs a higher z-index, otherwise, home sections overlay ontop of it.

Camc314 (commented):
image

Probably a good idea to hide volume controls on narrower screens

layouts/default.vue
Camc314: a value of 5 works well, and it does not overlay on the nav draw

ThibaultNocchi:
I do agree, especially when narrower screens are tablets or phones having dedicated volume buttons

Camc314:
> I do agree, especially when narrower screens are tablets or phones having dedicated volume buttons. But I think that's what the second screenshot show, doesn't it?

yeah that's right. The second screenshot in the original is a phone. My screenshot is a desktop window at about 50% width

ThibaultNocchi:
Oh yeah you're right, my bad. I don't know if totally hiding is a great idea, maybe only letting the icon and show a vertical slide above when hovering?

MrTimscampi:
Alright, so this evolved a bit, with the recent merge of the playback manager, and now has a basic Shaka-based audio player (basically a copy of VideoPlayer, but with an `<audio>` tag. It shall be replaced with the Howler-based player we have planned in the future) and shows information on currently playing audio, as well as allow some basic control and linking to artist/album.

It also adds a few niceties, like an itemHelper mixin, containing a `canPlay` method, which checks if the current item is playable, and shows the currently playing track in tracklists (I'd like to refine this in the future, so only the source of the currently playing queue highlights this currently playing item, much like Spotify does).

Still no queue button, fullscreen, shuffle or repeat, since that's for Preview Release 2, probably.

KristupasSavickas:
Minor issue I noticed - play button remains in paused state after transitioning to the next playlist item:

output

components/AudioControls.vue
```diff
@@ -0,0 +1,188 @@
...
+              <v-icon>mdi-skip-next</v-icon>
+            </v-btn>
+            <v-btn disabled icon class="mx-1">
+              <v-icon>mdi-repeat-off</v-icon>
+            </v-btn>
+          </div>
+          <v-slider
+            :value="currentPosition"
+            hide-details
+            min="0"
```
ferferga, line 64:
```suggestion
            :min="0"
```

This shouldn't be necessary anyway

ferferga:
@MrTimscampi After a rebase this is good to go imo. I made some quick small improvements to the slider, so it updates more smoothly (right now it makes a huge jump). I also added a nice label with the selected time and now the bar doesn't act like crazy when clicked or moved:

slider

If you want I can make a commit before your rebase so we can get this changes here or I can make a PR with cumulative improvements later, whatever you prefer :)

components/AudioControls.vue
```diff
@@ -0,0 +1,188 @@
...
+        </v-avatar>
+        <div class="d-flex flex-column justify-center ml-4">
+          <span class="font-weight-medium mt-md-n2">
+            <nuxt-link
+              tag="span"
+              class="text-truncate link"
+              :to="`/item/${getCurrentItem.AlbumId}`"
+              >{{ getCurrentItem.Name }}</nuxt-link
+            >
+            <v-btn class="d-none d-md-inline-flex" icon>
```
Camc314, line 20:
```suggestion
            <v-btn class="d-none d-md-inline-flex" icon disabled>

File: components/AudioControls.vue [hunk 4]
```diff
@@ -0,0 +1,188 @@
...
+          </v-img>
+        </v-avatar>
+        <div class="d-flex flex-column justify-center ml-4">
+          <span class="font-weight-medium mt-md-n2">
+            <nuxt-link
+              tag="span"
+              class="text-truncate link"
+              :to="`/item/${getCurrentItem.AlbumId}`"
+              >{{ getCurrentItem.Name }}</nuxt-link
+            >
```
Camc314, line 19:
```suggestion
              class="text-truncate link"
              :to="`/item/${getCurrentItem.AlbumId}`"
            >
              {{ getCurrentItem.Name }}
            </nuxt-link>

File: components/AudioControls.vue [hunk 5]
```diff
@@ -0,0 +1,188 @@
...
+                getCurrentItem.UserData.IsFavorite
+                  ? 'mdi-heart'
+                  : 'mdi-heart-outline'
+              }}</v-icon>
+            </v-btn>
+          </span>
+          <nuxt-link
+            tag="span"
+            class="text--secondary text-caption text-truncate mt-md-n2 link"
+            :to="`/artist/${getCurrentItem.AlbumArtists[0].Id}`"
```
Camc314, line 31: Isn't this making the assumption that there is an album artist.

In the case where there isn't, won't this produce an error

components/AudioPlayer.vue
```diff
@@ -0,0 +1,227 @@
+<template>
+  <audio
+    ref="audioPlayer"
+    :poster="poster"
```
Camc314, line 4:
```suggestion
```
There is nowhere to show the poster, so this is unneeded.

components/AudioPlayer.vue
```diff
@@ -0,0 +1,227 @@
...
+      'currentVolume'
+    ]),
+    poster(): string {
+      return this.getImageUrlForElement(ImageType.Backdrop, {
+        itemId:
+          this.getCurrentItem?.ParentBackdropItemId ||
+          this.getCurrentItem?.SeriesId ||
+          this.getCurrentItem?.Id
+      });
+    }
```
Camc314, line 59:
```suggestion

File: components/TrackList.vue [hunk 8]
```diff
@@ -38,7 +41,13 @@
             </td>
             <td style="width: 3em" class="pr-0 pl-0 text-center">
               <v-btn icon>
```
Camc314, line 43:
```suggestion
              <v-btn icon disabled>

**Review by Camc314 — COMMENTED** [review]
There is an empty file `LikeButton.vue` that can be removed

**Review by Camc314 — COMMENTED** [review]
Settings/metadata is inaccessible on narrow screens:

image

**Comment by neopc10** [comment]
Why the fullscreen button? It seems to me that it's a bit of a waste of space. On Tidal, for instance, clicking anywhere on the minibar will go to the full window player. It seems to me that JF would be well served by this. The add to playlist button, too, to perhaps an overflow menu.
image

**Comment by MrTimscampi** [comment]
@neopc10 We are planning to have this kind of screen for the fuill screen button:
image

While de queue button will navigate to the current queue.

@Camc314 That's z-index stuff with the navigation drawer. Afaik it's fixed on master, but this branch needs to be rebased.

@KristupasSavickas I've seen this happen with videos too, so it's not specific to this branch. Definitely needs to be investigated, though.

File: components/AudioControls.vue [hunk 9]
```diff
@@ -0,0 +1,221 @@
...
+              <v-icon>mdi-repeat-off</v-icon>
+            </v-btn>
+          </div>
+          <v-slider
+            :value="sliderValue"
+            hide-details
+            min="0"
+            :max="runtime"
+            validate-on-blur
+            thumb-label
```
Camc314, line 67:
```suggestion
```
The thumb label shows the incorrect values, since the slider stores value in terms of seconds

ferferga: @Camc314 They're overriden in the template below

hamburglar2160:
I really like what Tidal does here with their UX seems to borrow at least a bit from that. 

Spotify doesn't seem to give two craps about albums in general and I think that's a missed opportunity on their end. I think the Vue client would really benefit from having the album linked in addition to the song title and album artist. as posted above. Tidal actually has a "playing from <album> <playlist> <whatever>" link instead of album and I think that's a really nice touch that would be applicable here in the future.

Tidal also has a fantastic "whole window" player UI, which I don't think is being considered for this round of PRs, but certainly deserves serious consideration for the Vue client: the album cover, player, and metadata on the left 50% of the screen and the queue on the right 50%.  On Tidal's "whole window" UI there's also a button that leads to a "fullscreen" UI that I would argue is inferior to the Spotify one @mrtim linked above. 

Clearly the fullscreen button is placeholding for the Spotify-like fullscreen view. Is there an intention to do a whole window player UI as well? I would argue that the absolute sweet spot for flow in this client would be "play music -> minibar appears -> click anything that's not a button on the minibar to navigate to the tidal-like "whole window" player -> click fullscreen button there to get to a spotify-like fullscreen UI.

I think making the seekbar take up only a fraction of the minibar as opposed to the entire width of the screen is a downgrade from what jf-web has. Having that more granular control I think is useful. I see that Spotify chose to do that though and it certainly works with their fullscreen view. 

I'm bringing this up here and now only because the PR is just slightly less of a placeholder than it might have been an hour ago.

ferferga:
@hamburglar2160 That's for #238

components/AudioControls.vue
Camc314: Ahh I missed that, nice

MrTimscampi:
Indeed, please keep this on topic. PRs are mainly for implementation discussions, not brainstorming future features. We have discussion threads for every major feature specifically for gathering ideas and discussing said features and what we want to do with them.

Also, @hamburglar2160 , please avoid tagging people unrelated to the project. It was probably done accidentally, but it's quite a rude thing to do in general.

components/AudioControls.vue
MrTimscampi: The component accepts `number | string`, so this isn't needed.

components/AudioControls.vue
MrTimscampi: I always have an album artist in my library, even for albums from solo artists.

components/Layout/AudioControls.vue
```diff
@@ -0,0 +1,222 @@
...
+              <v-icon>mdi-skip-next</v-icon>
+            </v-btn>
+            <v-btn disabled icon class="mx-1">
+              <v-icon>mdi-repeat-off</v-icon>
+            </v-btn>
+          </div>
+          <v-slider
+            :value="sliderValue"
+            hide-details
+            min="0"
```
ferferga, line 65:
```suggestion
            :min="0"
```

This shouldn't be needed as it's the default. If we still want to keep it, this needs the ``v-bind`` decorator afaik, as I suggest.

components/Layout/AudioControls.vue
```diff
@@ -0,0 +1,222 @@
...
+    stopPlayback() {
+      this.setLastItemIndex();
+      this.resetCurrentItemIndex();
+      this.setNextTrack();
+    },
+    togglePause() {
+      if (this.isPaused) {
+        this.unpause();
+      } else {
+        this.pause();
```
ferferga, line 217:
```suggestion
    onPositionChange(value: number): void {
      if (!this.clicked) {
        this.changeCurrentTime({ time: value });
      }
    },
    onInputChange(value: number): void {
      this.currentInput = value;
    },
    onClick(): void {
      this.currentInput = this.realPosition;
      this.clicked = !this.clicked;
    },
    stopPlayback(): void {
      this.setLastItemIndex();
      this.resetCurrentItemIndex();
      this.setNextTrack();
    },
    togglePause(): void {
      if (this.isPaused) {
        this.unpause();
      } else {
        this.pause();
```

Sorry, I missed this in my commit and we need it if we want to merge #466 😅

components/Layout/VolumeSlider.vue
```diff
@@ -0,0 +1,39 @@
...
+
+export default Vue.extend({
+  computed: {
+    currentvolume(): number {
+      return this.$store.state.playbackManager.currentVolume;
+    }
+  },
+  methods: {
+    ...mapActions('playbackManager', ['setVolume']),
+    onVolumeChange(value: number) {
```
ferferga, line 28:
```suggestion
    onVolumeChange(value: number): void {
```

Same as above

components/Players/AudioPlayer.vue
```diff
@@ -0,0 +1,218 @@
...
+    getCurrentItem() {
+      this.getPlaybackUrl();
+    },
+    async source(newSource) {
+      if (this.player) {
+        try {
+          await this.player.load(newSource);
+        } catch (e) {
+          // No need to actually process the error here, the error handler will do this for us
+        }
```
ferferga, line 62:
```suggestion
    getCurrentItem(): void {
      this.getPlaybackUrl();
    },
    async source(newSource): Promise<void> {
      if (this.player) {
        try {
          await this.player.load(newSource);
        } catch (e) {
          // No need to actually process the error here, the error handler will do this for us
        }
```
Same as above

components/Players/AudioPlayer.vue
```diff
@@ -0,0 +1,218 @@
...
+              }
+              break;
+            case 'playbackManager/CHANGE_CURRENT_TIME':
+              if (this.$refs.audioPlayer) {
+                (this.$refs
+                  .audioPlayer as HTMLAudioElement).currentTime = this.currentTime;
+              }
+              break;
+            case 'playbackManager/SET_VOLUME':
+              console.warn(this.currentVolume);
```
ferferga, line 101: Not sure if we want this for anything?

components/Players/TrackList.vue
```diff
@@ -92,10 +101,8 @@ export default Vue.extend({
     };
   },
   computed: {
-    tracksPerDisc: {
-      get() {
-        return groupBy(this.$data.tracks.Items, 'ParentIndexNumber');
-      }
+    tracksPerDisc() {
```
ferferga, line 104:
```suggestion
    tracksPerDisc(): BaseItemDto[] {
```

Same as above

ferferga (commented):
Minor stuff

ferferga (approved):

PR #401, opened 2020-12-16
feat(noto): add noto sans packages

ferferga:


assets/global.scss
```diff
@@ -1,7 +1,16 @@
...
+@import "fontsource-noto-sans-kr/index.css";
+@import "fontsource-noto-sans-tc/index.css";
+@import "fontsource-noto-sans-hk/index.css";
+
+
 /* W3C standard (Firefox-only for now) */
 * {
   scrollbar-width: thin;
   scrollbar-color: var(--v-thumb-lighten5) var(--v-track-base);
+  font-family: 'Noto Sans', 'Noto Sans HK', 'Noto Sans JP', 'Noto Sans KR', 'Noto Sans SC', sans-serif;
```
MrTimscampi, line 13: See `variables.scss`, we're already setting it for a bunch of things via `$body-font-family`.

You can likely add `$heading-font-family`, though. Everything else should use these variables (Everything inside v-app already does, so pretty much all of the client uses it by default).

MrTimscampi (changes requested):

assets/global.scss
```diff
@@ -1,3 +1,10 @@
+@import 'fontsource-noto-sans/index.css';
```
ferferga, line 1: Not sure if we should move this to variables as well to keep everything in one place.

MrTimscampi (approved):

PR #407, opened 2020-12-17
feat(images): Create custom component for user image

ferferga:
Cherry-picked from #275


MrTimscampi (changes requested):
This breaks the user icon in the sidebar:

image

Setting an image also completely breaks the client.

ferferga:
@MrTimscampi I tested this before making this PR, even after the cherrypick. However, you're right and for some reason I switched back to the branch and it didn't work.

I merged master and now it works for me, try now please :)

MrTimscampi:
Now it doesn't crash :)

Looks good for the public users, but the main user button inside the app is still really different from what's currently in the client:
image

versus 

image

MrTimscampi (approved):
The space-evenly bit centers everything and looks weird, but we'll fix it in another PR.

PR #409, opened 2020-12-17
build(): disable modern and target static

ferferga:
I don't know if we had modern enabled for an special reason but, with it disabled, I get production and dev builds in half of the time, while maximizing compatibility with older browsers. Usability-wise, I don't find a worse performance as the docs claim (in fact, I think it has better performance while switching between routes)

We were also targeting server for some reason while our client is purely static

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -17,12 +17,12 @@ const config: NuxtConfig = {
    ** Nuxt target
    ** See https://nuxtjs.org/api/configuration-target
    */
-  target: 'server',
+  target: 'static',
   /*
    ** Module loading mode
    ** See https://nuxtjs.org/api/configuration-modern
    */
-  modern: 'client',
+  modern: false,
   /*
    ** Progress bar between routes
    ** See https://nuxtjs.org/api/configuration-loading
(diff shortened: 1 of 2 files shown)


MrTimscampi:
>  enabled for an special reason

We do, it's because the modern build use the `nomodule` attribute for script tags (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script). This removes a bunch of boilerplate code that older browsers need, making the bundles smaller and faster for modern browsers.

It also allows us to have two build environments:

* The modern one targets evergreen browsers, minimizing the amount of polyfills we need and making use of native (ie faster) implementations of things
* The legacy one targets ES5 and bundles more polyfills

It's something we've talked about doing in old web before, but were limited with how the build was made.

As for the `static`, we're not using it because it's not an SPA: it pre-generates the routes your app uses and makes a static site.  
Per the docs:
> When deploying your site in with target:static all your .vue pages will be generated into HTML and JavaScript files. All calls to APIs will be made and cached in a folder called static inside your generated content so that no calls to your API need to be made on client side navigation.
Source: https://nuxtjs.org/docs/2.x/concepts/static-site-generation/#generating-your-site

This is not what we want, as we want the app to stay and SPA and be fully dynamic. The static mode is more for websites like the Jellyfin official site than what we're building here.

PR #418, opened 2020-12-20
fix(activity name escape): escaped activity name of html specials

sarab97:
small fix for escaping html specials chars from activity name

### DIFF
--- a/pages/settings/logsAndActivity.vue
+++ b/pages/settings/logsAndActivity.vue
@@ -47,7 +47,7 @@
                 <v-icon dark v-text="getIconFromType(activity.Type)"></v-icon>
               </v-list-item-avatar>
               <v-list-item-content>
-                <v-list-item-title v-text="activity.Name" />
+                <v-list-item-title v-text="decodeHTML(activity.Name)" />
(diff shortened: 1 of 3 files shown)


pages/settings/logsAndActivity.vue
```diff
@@ -72,6 +72,7 @@ import Vue from 'vue';
 import { mapActions } from 'vuex';
 import colors from 'vuetify/lib/util/colors';
 import { ActivityLogEntry, LogFile, LogLevel } from '@jellyfin/client-axios';
+import { decodeHTML } from 'entities';
```
MrTimscampi, line 75: You need to add the packages to the dependencies with `yarn add entities`.

Currently it's not an explicit dependency of ours. While that's fine for core stuff like Vue and Vuex, this kind of dependency should be explicit.

MrTimscampi (changes requested):

pages/settings/logsAndActivity.vue
sarab97: done

MrTimscampi (approved):

PR #432, opened 2020-12-23
First usable pass of reponsive item page

ThibaultNocchi:
The wide format stays more or less the same, I only changed the narrower ones.

# Widest XS
Screenshot_2020-12-31 Jellyfin

# Lowest and widest SM
Screenshot_2020-12-31 Jellyfin(1)

Screenshot_2020-12-31 Jellyfin(2)

# Lowest and widest MD
Screenshot_2020-12-31 Jellyfin(3)
Screenshot_2020-12-31 Jellyfin(4)

# Lowest and widest LG
Look the same as MD with the navigation drawer added on the left

# Lowest XL
Screenshot_2020-12-31 Jellyfin(5)

***
Also not visible, but the picture is kinda blurred, even though the loaded file by the browser is way clearer. Maybe the blurhash?


ferferga:
@ThibaultNocchi If this is still a draft, I'll mark it as is so you can keep working on this without us chiming in too frequently.

Mark as ready to review as soon as you feel this is finished! :)

ThibaultNocchi:
@ferferga I misnamed my PR, it's not a draft in the usual way. Like I'm kinda happy with how it looks for now, and it seems to be usable on different screens.

I'm just waiting for a few advices or if it suits everyone. :)

MrTimscampi:
The `xs` breakpoint seems wrong:

image

We should likely hide (`v-show` or `v-if`) the right column on the `xs` breakpoint, then add a block like current web with writer and director above the genres on `xs`, and the horizontal version of the actors at the bottom of the page.

`sm` and up look okay (I'm taking an iPad as the smallest `sm` device, which isn't quite right, but eh).

ThibaultNocchi:
@MrTimscampi I've pushed the responsive fix for the crew column on `xs` and added the directors and writers as chips under genres as this resolution. Tell me what you think of it.

Also, with the added computed methods `directors` and `writers`, I get a TS error:

```
ERROR in pages/item/_itemId/index.vue:426:19
TS2339: Property 'crew' does not exist on type 'CombinedVueInstance<Vue, unknown, unknown, unknown, Readonly<Record<never, any>>>'.
```

pages/item/_itemId/index.vue
```diff
@@ -1,14 +1,18 @@
 <template>
-  <v-container class="ml-3 mr-3">
+  <v-container class="pl-6 pr-6">
```
Camc314, line 2:
```suggestion
  <v-container class="px-6">

File: pages/item/_itemId/index.vue [hunk 2]
```diff
@@ -19,13 +23,15 @@
             />
             <h2
               v-if="loaded && item.OriginalTitle"
-              class="text-subtitle-1 text-truncate"
+              class="text-subtitle-1"
+              :class="{ 'text-center': !$vuetify.breakpoint.mdAndUp }"
```
Camc314, line 27: Why not move this class to the parent `v-col` that way we don't have to duplicate it several times

Camc314 (commented):
This looks nice! Few comments, also, for content with no actors, would it be possible grow the content containing the primary image, title e.c.t to contain the full width? (see below)

image

If there is no cast/crew (red square) expand green square to 100%

pages/item/_itemId/index.vue
ThibaultNocchi: Cause the parent `v-col` is also parent to the one L85 and this one doesn't need to be left aligned :)

pages/item/_itemId/index.vue
ThibaultNocchi: Done

ThibaultNocchi:
@Camc314 Although I like this idea, you'll see that L419 there's the `related-items` component in this same column, and I can't know if it's empty. As the related items component doesn't apply to movies, I can hide the column on movies if the cast is empty. But 1) won't movies have related items later on? 2) Movies without the director are rare I'd assume, what do you think?

MrTimscampi:
I think we should have placeholders instead of changing the columns around if things are empty (We risk having the page change layout on the user without warning as data loads or refreshes, which is quite jarring)

ThibaultNocchi:
What are you referring to? Hiding the cast column when there's no one?

ferferga:
@ThibaultNocchi @MrTimscampi Maybe put the tags there when there are no items?

MrTimscampi:
> What are you referring to? Hiding the cast column when there's no one?

Yep

> Maybe put the tags there when there are no items?

We shouldn't change the position of things depending on what is filled or not :/ That's a super confusing thing to do.

Camc314:
> We shouldn't change the position of things depending on what is filled or not :/ That's a super confusing thing to do.

It just looks a bit strange, especially on narrower devices to have a blank column on the right

ThibaultNocchi:
All things aside, it's really easy to hide the right column and let the left one take as much place as it can, when the cast is empty. But I don't see many cases where it happens, and on series and albums there should be the related-items component there (and eventually on movies too I guess).

MrTimscampi:
> > We shouldn't change the position of things depending on what is filled or not :/ That's a super confusing thing to do.
> 
> It just looks a bit strange, especially on narrower devices to have a blank column on the right

We could have placeholders for no information:

```
Crew
Nothing to show

Cast
Nothing to show
```

Or we could rework the design to remove that column entirely and move the cast somewhere else (Using tabs, like for artists bios, maybe).

MrTimscampi:
> All things aside, it's really easy to hide the right column and let the left one take as much place as it can, when the cast is empty. But I don't see many cases where it happens, and on series and albums there should be the related-items component there (and eventually on movies too I guess).

Movies are pretty much expected to have casts and crews, it's the user's problem if they don't have providers, imo.

For every other media type, we'll likely end up making bespoke details page, since they all have slightly different needs.

ferferga:
@MrTimscampi That doesn't provide anything useful, I think it's worst than leaving it blank to be honest.

If moving the tags doesn't convince the majority here, I'd rather have everything centered and blank space at both sides than useless placeholders.

MrTimscampi:
> I think it's worst than leaving it blank to be honest.

But they aren't expected to be blank. Currently, the page is a generic item page, but it's not its final purpose: it's for movies (It was built with them in mind explicitly). And movies are expected to have metadata, which means cast and crew.

If people don't provide any, that's their issue entirely.

There is pretty much 0.01% of cases where there won't be anything in the right column, so imo it's pretty much a non-issue.

ThibaultNocchi:
(Also tags only appear on `xs` screens ;) )

MrTimscampi (approved):
LGTM. Only change would be dealing with the right column's content when in `xs` (hiding the crew), but that can be done in another PR, no need to hold this one up.

ThibaultNocchi:
How would you do that? As it's already in the bottom, do you mean like in an accordion?

MrTimscampi:
> How would you do that? As it's already in the bottom, do you mean like in an accordion?

Just hiding the crew section on `xs` with a v-if or a v-show, since its info is already (mostly) in the main section on that size.

ThibaultNocchi:
Oh yeah right, but you do have the pictures! :D

PR #451, opened 2020-12-26
refactor(swiper-progress): fix z-index issue, responsive placement and adds padding

MrTimscampi:
* Fix a z-index issue where the progress bars were going over the top bar
* Fix the positioning of the bar in the xs and sm breakpoints
* Add padding to the bar to fit better with the other padding around the page

Screenshot_2020-12-26 Jellyfin

Screenshot_2020-12-26 Jellyfin

### DIFF
--- a/assets/global.scss
+++ b/assets/global.scss
@@ -44,3 +44,10 @@
 .layout-leave-to {
   opacity: 0;
 }
+
+/* Vuetify overrides
+   We can't set these through variables, so override them here */
+
+.v-app-bar.v-app-bar--fixed {
+  z-index: 100 !important;
+}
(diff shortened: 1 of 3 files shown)


ferferga:
@MrTimscampi Are the bars still clickable? They weren't with z indexes below 5. 

I placed the progress bars below search so they don't take image space, but I don't mind either.

MrTimscampi:
> @MrTimscampi Are the bars still clickable? They weren't with z indexes below 5.

Ah, indeed. This is probably due to overlapping with another div. I have a solution, gimme a bit :)

MrTimscampi:
@ferferga Should be good now :+1: 

Though there's some usability issue with the bars (not related to this PR). They're difficult to click on due to the even zone being only on the bar itself, which is only a couple of pixels tall. Ideally there'd be a thumb zone around the bar where the event still works. But that's for another PR, I think :yum:

ferferga (approved):
Everything's good now

PR #455, opened 2020-12-27
feat(homeheader): Header is now more polite and smooth while content is loading

ferferga:
## Changes

* Welcome the user while the client is waiting for a response from the server. *Hello (user)* **always appears**, while *check out what's new* appears **as soon as the items to show are fetched and processed**

**Desktop**:

https://user-images.githubusercontent.com/10274099/103175300-3c938980-4869-11eb-8ef5-b4958c07efa8.mp4

**Mobile**:

https://user-images.githubusercontent.com/10274099/103175530-cbed6c80-486a-11eb-9402-ae2f4e2889df.mp4

* Same welcome message appears, **but indicating that there are no items to show when there are no results from server**, showing a **nice slide up animation, removing the jumpiness of the current implementation**

**Desktop**:

https://user-images.githubusercontent.com/10274099/103175358-b75ca480-4869-11eb-8219-e4a243f7f689.mp4

**Mobile**:

https://user-images.githubusercontent.com/10274099/103175553-0bb45400-486b-11eb-8f5e-20a0a1160750.mp4

* Added an indicator that shows why that item is in the swiper. This is a placeholder for now, **see 'Future Plans' for more info**:

image

* Fixed non-working transitions on BlurhashImage.
* Fixes cropped text (specially visible on **g** and accents) everywhere through the UI.

## Future plans (not for this PR)

* Would be cool if the text says good afternoon, good morning, etc based on local time.
* Customization options
* Show the *Hello* message only when loading the client from scratch, and not when coming back from another page.
* Make the home header swiper more "intelligent": 
      My concept for the perfect design for this would be something that shows you "mixed" kinds of content, **not only the recently added**. Imagine that I'm listening to music: **it would be great that a few slides with related tracks or albums for the current playing song/artist/album appear, but only while it's being played**

     Another example would be to show next up episodes or in-progress items, **so everything's at a hand without much hassle**.

     This, with proper customization of course, will bring a rich home experience, that currently doesn't exists for servers where not a lot of content is added (defeating the purpose). In my opinion, **the welcome messages that this PR adds sets the baseline for this experience, as we can show a different message depending on what kind of slide combinations we want to show at an specific moment**


ferferga:
As a side note, I think we should move the styles to some sort of shared ``scss``, as ``HomeHeaderItems`` and ``HomeHeaderWelcome`` share mostly the same styles. I thought about moving them to ``global.scss`` but that would make them not-scoped. Ideally, **it would be great if we can scope them to those two components only so we don't need to play with class names.** I researched a bit in a way to do this but I found nothing, so any input is appreciated.

ThibaultNocchi:
You could maybe use a simple style file in a `style` directory which you'd import?

MrTimscampi:
> You could maybe use a simple style file in a `style` directory which you'd import?

This is the correct way, I believe. Create a new SCSS file, then import it in the scoped block.

components/HomeHeader.vue
```diff
@@ -179,137 +67,59 @@ export default Vue.extend({
...
-    getOverview(item: BaseItemDto): string {
-      if (item.Overview) {
-        return this.sanitizeHtml(item.Overview);
+    hideWelcomeMessage(): void {
+      if (this.items.length === 0) {
+        const elem = this.$refs.headerWelcome as HTMLElement;
+        // As the height of the element varies as we're using automatic values, we set first
+        // the screen height of the element, so the animation plays correctly when the height is set to 0
+        elem.style.maxHeight = elem.scrollHeight + 'px';
+        elem.classList.add('no-items');
```
MrTimscampi, line 85: Use bound classes/styles for this.

components/HomeHeader.vue
```diff
@@ -179,137 +67,59 @@ export default Vue.extend({
...
+        elem.addEventListener('animationend', this.onNoItemsTransitionEnd);
```
MrTimscampi, line 86: Bind the event directly on the component.

components/HomeHeader.vue
```diff
@@ -179,137 +67,59 @@ export default Vue.extend({
...
-    // doesn't get the parameters passed correctly on components that calls to methods. Whenever the beginning or the end is reached,
-    // we force a BlurhashImage reload to fix this by updating it's key.
-    //
-    // TODO: Revisit this once we are using the original Swiper.js library.
-    forceReload(): void {
-      this.reloadSentinel = 1;
-      this.reloadSentinel = 0;
+    onNoItemsTransitionEnd(): void {
+      (this.$refs.headerWelcome as Element).remove();
+      this.$destroy();
```
MrTimscampi, line 93: It'd be better to use v-if to do this, as it's essentially the same and is cleaner.

ferferga:
@ThibaultNocchi @MrTimscampi That worked perfectly, for some reason I thought that all the styles under ``assets`` were loaded globally. I found that happens only for those defined in ``nuxt.config.ts``

Thank both very much, pushing!

components/HomeHeaderItems.vue
```diff
@@ -0,0 +1,223 @@
...
+          />
+        </div>
+        <div class="slide-content">
+          <v-container
+            fill-height
+            class="mx-md-10 mt-md-5 py-0 py-md-4 align-end align-sm-center align-md-start"
+          >
+            <v-row>
+              <v-col cols="12" sm="8" md="6" xl="5" class="py-0 py-md-4">
+                <p class="text-truncate mb-2 my-2">
```
MrTimscampi, line 26:
```suggestion
                <p class="text-overline text-truncate mb-2 my-2">
```

This looks a bit better, I think.

MrTimscampi (commented):
Some minor stuff.

I also think it'd be nice for the text to be more vertically centered on the `xs` breakpoint (It's right up against the header here, it feels a bit off).

Really like it otherwise :+1:

ThibaultNocchi:
Also I've been thinking about caching the results somewhere in the store, for library views for instance (kinda cumbersome to reload the whole list when you've checked a movie and get back to the list). So maybe we could implement some generic caching mechanism. Maybe even something with Axios already exists which would be pretty transparent for us?

components/HomeHeader.vue
ferferga: @MrTimscampi Although I agree that we should bind this kind of stuff as much as possible, I don't think here it makes sense. It adds overcomplication (we need to add another boolean variable to the equation, as ``!items.length && !loading`` contradicts each other):

* ``items.length`` will always be 0 when there are no items to show, **but there's a sensible amount of seconds that it's 0 while data is being fetched from server, making the content start the animation and strop halfway**
* Setting ``loading`` to false will immediately kill the Welcome component, as it will also trigger the ``v-if``
* Getting the element's scrollHeight will need a computed property (which are a few lines) while this is a single one.

ferferga:
@MrTimscampi Text jumps if aligned vertically to the center.

We can align text to the center horizontally though.

https://user-images.githubusercontent.com/10274099/103178151-93f12400-4880-11eb-89ae-d29d39cd69e2.mp4

MrTimscampi:
> Text jumps if aligned vertically to the center.

Maybe some top margin then? Could also make the text a bit smaller, it'll look less "too large"

ferferga:
New look

**MD**:
image

**LG**:
image

**XS**:
image

MrTimscampi:
Mobile looks good now :+1:

I think left-aligned works better on larger screen though (Larger area to look at, keeping it on the left keeps the focus of the eye to where the item's info will be appearing at)

ferferga:
@MrTimscampi Agree with you now that I'm looking it again :)

Done!

MrTimscampi (approved):

PR #459, opened 2020-12-27
fix(userimage): show avatar placeholder when loading user image

ThibaultNocchi:


components/UserImage.vue
```diff
@@ -1,6 +1,12 @@
 <template>
   <v-avatar v-if="userImage">
-    <v-img :src="userImage" :alt="$auth.user.Name" class="userImage absolute" />
+    <v-img :src="userImage" :alt="$auth.user.Name" class="userImage absolute">
```
ferferga, line 3:
```suggestion
    <v-img :src="userImage" :alt="$auth.user.Name" class="userImage">
```

This is some testing bits I forgot to remove in my PR, so if you can ammend your commit and do a force push it would be great :)

ferferga (changes requested):
Works great, requested changes are not your fault :)

components/UserImage.vue
ThibaultNocchi: Done :)

ferferga (approved):

PR #461, opened 2020-12-29
fix (swiperprogressbar): remove event listeners on destroy

ferferga:
After going from home to another page, sometimes, the event listeners for the animation end fired up when the progress elements were already destroyed.

This didn't crash, show or cause any visible error to the end user, but ``TypeErrors`` were appearing in the consoles and we don't want this :)

### DIFF
--- a/components/Layout/HomeHeader/SwiperProgressBar.vue
+++ b/components/Layout/HomeHeader/SwiperProgressBar.vue
@@ -62,6 +62,13 @@ export default Vue.extend({
       window.requestAnimationFrame(this.updateBars);
     });
   },
+  destroyed() {
+    const animEndFunction = this.onAnimationEnd;
+
+    this.bars.forEach((el: HTMLElement) => {
+      el.removeEventListener('animationend', animEndFunction);
+    });
+  },
   methods: {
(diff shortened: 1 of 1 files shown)


components/Layout/HomeHeader/SwiperProgressBar.vue
```diff
@@ -62,6 +62,13 @@ export default Vue.extend({
       window.requestAnimationFrame(this.updateBars);
     });
   },
+  destroyed() {
+    const animEndFunction = this.onAnimationEnd;
+
+    this.bars.forEach(function (el: HTMLElement) {
```
MrTimscampi, line 68: Any reason for not using an arrow function?

MrTimscampi (commented):
Just one question, otherwise it's good

components/Layout/HomeHeader/SwiperProgressBar.vue
ferferga: @MrTimscampi None really, never used them myself. Changed all the occurences in the code of this component, thanks for the suggestion! 👍

MrTimscampi (approved):

PR #464, opened 2020-12-29
Convert some things to functional components

MrTimscampi:
* Converts the connection monitor to a functional component
* Clean up the home header items by moving the logo/titles to functional components (One for albums, one for episodes and a generic one)

### DIFF
--- a/components/Layout/HomeHeader/HomeHeader.vue
+++ b/components/Layout/HomeHeader/HomeHeader.vue
@@ -44,6 +44,7 @@ export default Vue.extend({
         limit: this.pages,
         fields: [ItemFields.Overview],
         enableImageTypes: [ImageType.Backdrop, ImageType.Logo],
+        includeItemTypes: ['Episode', 'Movie', 'MusicAlbum', 'Series'],
         imageTypeLimit: 1
       })
     ).data;
--- a/components/Layout/HomeHeader/HomeHeaderItems.vue
+++ b/components/Layout/HomeHeader/HomeHeaderItems.vue
@@ -23,66 +23,29 @@
           >
             <v-row>
               <v-col cols="12" sm="8" md="6" xl="5" class="py-0 py-md-4">
-                <p class="text-overline text-truncate mb-2 my-2">
-                  {{ $t('homeHeader.items.recentlyAdded') }}
-                </p>
-                <v-img
-                  v-if="
-                    item.ParentLogoImageTag ||
-                    (item.ImageTags && item.ImageTags.Logo)
-                  "
-                  :max-width="$vuetify.breakpoint.mdAndUp ? '50%' : '40%'"
-                  aspect-ratio="2.58"
-                  contain
-                  :src="getLogo(item)"
+                <home-header-album-title
+                  v-if="item.Type === 'MusicAlbum'"
+                  class="mb-sm-n1"
+                  :item="item"
+                  :logo="getLogo(item)"
                 />
-                <h1
+                <home-header-episode-title
                   v-else-if="item.Type === 'Episode'"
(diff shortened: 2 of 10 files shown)


components/Layout/HomeHeader/HomeHeader.vue
```diff
@@ -44,6 +44,7 @@ export default Vue.extend({
         limit: this.pages,
         fields: [ItemFields.Overview],
         enableImageTypes: [ImageType.Backdrop, ImageType.Logo],
+        includeItemTypes: ['Episode', 'Movie', 'MusicAlbum', 'Series'],
```
ferferga, line 47: Why not use ``ItemType``? We use it for Image types, this would make things consistent imo.

components/Layout/HomeHeader/HomeHeaderItems.vue
```diff
@@ -124,12 +87,15 @@
...
 import Swiper, { SwiperOptions } from 'swiper';
+import { VImg } from 'vuetify/lib';
 import { mapActions } from 'vuex';
 import { BaseItemDto, ImageType } from '@jellyfin/client-axios';
 import htmlHelper from '~/mixins/htmlHelper';
 import imageHelper from '~/mixins/imageHelper';
 
 export default Vue.extend({
+  // eslint-disable-next-line vue/no-unused-components -- Used in child functional components
+  components: { VImg },
```
ferferga, line 98: Can't we import those in the child components?

ferferga (changes requested):
I'm not sure that doing the functional component thing is a good idea, [those are not recommended for Vue 3](https://v3.vuejs.org/guide/migration/functional-components.html#overview).

If they also have the hassle that you must import child components in the parent, I think it completely defeats it's purpose.

components/Layout/HomeHeader/HomeHeaderItems.vue
MrTimscampi: Functional components cannot use `components`. We import it in the parent, so that children can use them (Otherwise, the component is undefined).

MrTimscampi:
> [those are not recommended for Vue 3](https://v3.vuejs.org/guide/migration/functional-components.html#overview)

That's cool, but we're still a year away from moving to Vue 3. So I think it's warranted for now. We can re-evaluate once we migrate.

> If they also have the hassle that you must import child components in the parent, I think it completely defeats it's purpose.

Not really. The purpose is that it takes a lot less time to be created and added to the DOM, which is what we want for this kind of components.  
Declaring components in the parent isn't really an issue, imo.

ferferga:
@MrTimscampi I don't mind incorporating these really, you know already I love to try to squish as much performance as possible :).

I gave my point about Vue 3 because, although it's far in time, I think we should already be thinking in ways to make the transition as easy as possible. I, for instance, avoided the use of ``$destroy()`` as much as possible as it's another API that will be deprecated.

Right now we're manageable but with the pace we have who knows how messy removing this things could become when we're migrating to Vue 3.

As the other side of the coin, where I agree with you, we don't really know if we will be still on preview while we would be able to do the transition to Vue 3 or we will have already an stable version published. If we ever get stable in Vue 2, it makes total sense to do this. If we don't, then it doesn't really, as this client is still on preview.

The problem is **we don't know when the transition or the stable release will happen either** so this is difficult lol 😁.

MrTimscampi:
Closing this for now.

ferferga:
@MrTimscampi Why you don't cherry pick the tests and the division of the HomeHeaderItems to subcomponents?

Even if they're not functional, I think it's a good idea towards a better maintenance and I'm 100% for it.

PR #466, opened 2020-12-30
Force return types and arrow callbacks on functions

ferferga:
Added a rule for enforcing return types on all the functions, which should help us with type checking across all the code

We should also investigate how we can get ``VueI18n`` to return strings instead of ``TranslateResult``, so we don't need to invoke ``toString()`` or do any string casting.

### DIFF
--- a/components/Buttons/FilterButton.vue
+++ b/components/Buttons/FilterButton.vue
@@ -283,13 +283,13 @@ export default Vue.extend({
     };
   },
   watch: {
-    itemsType() {
+    itemsType(): void {
       this.refreshItems();
     }
   },
   methods: {
     ...mapActions('snackbar', ['pushSnackbarMessage']),
-    async refreshItems() {
+    async refreshItems(): Promise<void> {
       try {
         const response = (
           await this.$api.filter.getQueryFiltersLegacy({
@@ -317,7 +317,7 @@ export default Vue.extend({
         });
       }
     },
-    emitFilterChange() {
+    emitFilterChange(): void {
       this.$emit('change', {
         status: this.selectedStatusFilters,
         features: this.selectedFeatureFilters,
--- a/components/Buttons/UserButton.vue
+++ b/components/Buttons/UserButton.vue
@@ -29,6 +29,11 @@
 import Vue from 'vue';
 import { mapActions } from 'vuex';
 
+interface MenuItem {
+  title: string;
+  action: () => void;
+}
+
 export default Vue.extend({
   data() {
     return {
@@ -36,7 +41,7 @@ export default Vue.extend({
     };
   },
   computed: {
-    menuItems(): Array<Record<never, never>> {
+    menuItems(): MenuItem[] {
       return [
         {
           title: this.$t('logout'),
--- a/components/Buttons/UserButton.vue
+++ b/components/Buttons/UserButton.vue
@@ -36,11 +36,11 @@ export default Vue.extend({
     };
   },
   computed: {
-    menuItems() {
+    menuItems(): Array<Record<never, never>> {
       return [
         {
           title: this.$t('logout'),
-          action: () => {
+          action: (): void => {
             this.logoutUser();
           }
         }
@@ -50,7 +50,7 @@ export default Vue.extend({
   methods: {
     ...mapActions('user', ['clearUser']),
     ...mapActions('deviceProfile', ['clearDeviceProfile']),
-    logoutUser() {
+    logoutUser(): void {
       this.$disconnect();
       this.$auth.logout();
       this.clearDeviceProfile();
--- a/components/Forms/AddServerForm.vue
+++ b/components/Forms/AddServerForm.vue
@@ -57,7 +57,7 @@ export default Vue.extend({
   },
   methods: {
     ...mapActions('servers', ['connectServer']),
-    connectToServer() {
+    connectToServer(): void {
       this.loading = true;
       this.connectServer(this.serverUrl);
       this.loading = false;
--- a/components/Item/Metadata/PersonEditor.vue
+++ b/components/Item/Metadata/PersonEditor.vue
@@ -65,7 +65,7 @@ export default Vue.extend({
   props: {
     person: {
       type: Object,
-      default: (): Record<string, string> => ({
+      default: (): BaseItemPerson => ({
         Name: '',
         Type: '',
         Role: ''
--- a/components/Item/Metadata/PersonEditor.vue
+++ b/components/Item/Metadata/PersonEditor.vue
@@ -65,7 +65,7 @@ export default Vue.extend({
   props: {
     person: {
       type: Object,
-      default: () => ({
+      default: (): Record<string, string> => ({
         Name: '',
         Type: '',
         Role: ''
@@ -90,22 +90,22 @@ export default Vue.extend({
     };
   },
   watch: {
-    person(value: BaseItemPerson) {
+    person(value: BaseItemPerson): void {
       if (!value) return;
       this.editState = value;
     }
   },
   methods: {
-    handleSubmit() {
+    handleSubmit(): void {
       this.$emit('update:person', this.editState);
       this.$emit('update:dialog', false);
       this.reset();
     },
-    handleCancel() {
+    handleCancel(): void {
       this.$emit('update:dialog', false);
       this.reset();
     },
-    reset() {
+    reset(): void {
       this.editState = { Name: '', Type: '', Role: '' };
     }
   }
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -159,7 +159,7 @@ export default Vue.extend({
       return [
         {
           icon: 'mdi-home',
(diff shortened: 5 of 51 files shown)


ThibaultNocchi:
Maybe a simple global mixin to wrap the calls to i18n?

ferferga:
@ThibaultNocchi Just found this: https://github.com/kazupon/vue-i18n/issues/410#issuecomment-538465940

Making a commit to replace all occurences

components/Buttons/FilterButton.vue
```diff
@@ -9,7 +9,7 @@
         v-bind="attrs"
         v-on="on"
       >
-        {{ $t('filter') }}
+        {{ $tc('filter') }}
```
MrTimscampi, line 12: I'm not in favor of this.

`$tc()` is meant to handle plurals and should only be used to do that, imo.

I don't see the issue with using `.toString()` in TypeScript code. Templates won't need it and it doesn't add much.

Overall, this feels very hacky and like an abuse of the vue-i18n APIs.

I wouldn't be in favor of wrapping `$t()` either, mind you (We should avoid wrapping things up for the hell of it).

ThibaultNocchi:
@ferferga In the Vue i18n API, `$tc` also returns a `TranslateResult`, so I guess your typings should work with `$t`

components/Buttons/FilterButton.vue
ferferga: @MrTimscampi What about extending the interface?

```
declare module 'vue/types/vue' {
interface Vue {
$t: (name: string, attrs?: any) => string
```

ferferga:
@ThibaultNocchi That's what they say but doesn't translate into practice 🤷‍♂️

components/Buttons/FilterButton.vue
MrTimscampi: That should be fine for our use. It's a string or a set of key/values, but we should always get the string version with how we use it.

See: https://github.com/kazupon/vue-i18n/blob/v8.x/types/index.d.ts#L18

layouts/default.vue
```diff
@@ -143,7 +143,7 @@ export default Vue.extend({
           };
         })
     }),
-    items() {
+    items(): Array<Record<string, string>> {
```
Camc314, line 146: This should be a custom type

layouts/default.vue
```diff
@@ -152,7 +152,7 @@ export default Vue.extend({
         }
       ];
     },
-    configItems() {
+    configItems(): Array<Record<string, string>> {
```
Camc314, line 155: This should be a custom type

pages/person/_itemId/index.vue
```diff
@@ -172,7 +172,7 @@ export default Vue.extend({
       }
     },
     birthPlace: {
-      get() {
+      get(): null {
```
Camc314, line 175:
```suggestion
      get(): string | null {

File: pages/person/_itemId/index.vue [hunk 5]
```diff
@@ -190,7 +190,7 @@ export default Vue.extend({
       }
     },
     shows: {
-      get() {
+      get(): Array<Record<string, string>> {
```
Camc314, line 193:
```suggestion
      get(): BaseItemDto[] {

File: plugins/browserDetection.ts [hunk 6]
```diff
@@ -100,10 +100,10 @@ class BrowserDetector {
...
    */
-  isApple() {
-    return (
-      navigator.vendor && navigator.vendor.includes('Apple') && !this.isTizen()
-    );
+  isApple(): boolean {
+    return (navigator.vendor &&
+      navigator.vendor.includes('Apple') &&
+      !this.isTizen()) as boolean;
   }
```
Camc314, line 107:
```suggestion
  return navigator?.vendor.includes('Apple') && !this.isTizen();

File: store/homeSection.ts [hunk 7]
```diff
@@ -45,7 +45,9 @@ type MutationPayload = {
 };
 
 export const getters: GetterTree<HomeSectionState, AppState> = {
-  getHomeSectionContent: (state) => (section: HomeSection) => {
+  getHomeSectionContent: (state) => (
+    section: HomeSection
+  ): Array<never> | BaseItemDto | BaseItemDto[] => {
```
Camc314, line 50:
```suggestion
  ): BaseItemDto[] => {

File: store/tvShows.ts [hunk 8]
```diff
@@ -31,10 +31,14 @@ type MutationPayload = {
...
-  getSeasons: (state) => ({ itemId }: { itemId: string }) => {
+  getSeasons: (state) => ({ itemId }: { itemId: string }): BaseItemDto[] => {
     return state[itemId]?.seasons;
   },
-  getSeasonEpisodes: (state) => ({ itemId }: { itemId: string }) => {
+  getSeasonEpisodes: (state) => ({
+    itemId
+  }: {
+    itemId: string;
+  }): Array<BaseItemDto[]> => {
```
Camc314, line 41:
```suggestion
  }): BaseItemDto[][] => {

File: utils/hlsFormats.ts [hunk 9]
```diff
@@ -7,7 +7,9 @@ import { browserDetector } from '~/plugins/browserDetection';
  * @param {HTMLVideoElement} videoTestElement A HTML video element for testing codecs
  * @returns {boolean} Determines if the browser has AC3 in HLS support
  */
-function supportsAc3InHls(videoTestElement: HTMLVideoElement) {
+function supportsAc3InHls(
+  videoTestElement: HTMLVideoElement
+): boolean | string {
```
Camc314, line 12:
```suggestion
): boolean {

File: utils/mp4AudioFormats.ts [hunk 10]
```diff
@@ -57,7 +57,7 @@ function hasMp2AudioSupport(): boolean {
  * @param {HTMLVideoElement} videoTestElement A HTML video element for testing codecs
  * @returns {boolean} Determines if browserr has DTS audio support
  */
-function hasDtsSupport(videoTestElement: HTMLVideoElement) {
+function hasDtsSupport(videoTestElement: HTMLVideoElement): boolean | string {
```
Camc314, line 60:
```suggestion
function hasDtsSupport(videoTestElement: HTMLVideoElement): boolean {

File: components/Forms/AddServerForm.vue [hunk 11]
```diff
@@ -57,7 +57,7 @@ export default Vue.extend({
   },
   methods: {
     ...mapActions('servers', ['connectServer']),
-    connectToServer() {
+    connectToServer(): void {
```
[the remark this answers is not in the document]
ferferga: @Camc314 ``connectServer`` (from store) is ``async``, this one isn't.

components/Item/Metadata/PersonEditor.vue
```diff
@@ -65,7 +65,7 @@ export default Vue.extend({
   props: {
     person: {
       type: Object,
-      default: () => ({
+      default: (): Record<string, string> => ({
```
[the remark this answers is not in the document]
ferferga: @Camc314 I tracked this down and parent item is passing ``BaseItemPerson``, so I used it

layouts/default.vue
ferferga: @Camc314 Created ``LayoutButton`` type for this.

layouts/default.vue
ferferga: @Camc314 Created ``LayoutButton`` type for this.

pages/person/_itemId/index.vue
ferferga: @Camc314 Good catch

utils/hlsFormats.ts
ferferga: @Camc314 This function also returns the string from the ``canPlayType`` https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType

utils/mp4AudioFormats.ts
ferferga: @Camc314 Same as above

components/Buttons/UserButton.vue
```diff
@@ -36,11 +36,11 @@ export default Vue.extend({
...
-    menuItems() {
+    menuItems(): Array<Record<never, never>> {
       return [
         {
           title: this.$t('logout'),
-          action: () => {
+          action: (): void => {
             this.logoutUser();
           }
         }
```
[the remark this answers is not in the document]
ferferga: @Camc314 Created ``MenuItem`` type for this

MrTimscampi (approved):
LGTM

components/Forms/AddServerForm.vue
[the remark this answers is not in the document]
Camc314: Then there should be an `await` on `this.connectServer(this.serverUrl);`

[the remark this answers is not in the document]
ferferga: @Camc314 Fixed on #482, not sure why this wasn't async in the first place

PR #467, opened 2020-12-30
fix(global.scss): fix line height truncating a lot of letters

ThibaultNocchi:


ferferga:
@ThibaultNocchi Any reason why this is still a WIP?

ThibaultNocchi:
@ferferga yep, as both you and @MrTimscampi don't have the problem I have, I don't think it should be merged. Either it's not a common problem and it's not a really good idea to blanket erase all the `line-height` rules of Vuetify, or it's a common one and maybe there is / will be a better solution.

MrTimscampi:
Closing this for now, since it seems to be a user-specific issue. We can always re-open it if more people come out with the same problem.

ThibaultNocchi:
Do you know if I can delete the branch for now? Or should it stay to keep the commits for the PR?

MrTimscampi:
If you want to keep this PR, you should leave the branch. Otherwise, you can safely delete it.

The changes are so small that we could pretty easily make a new PR if need be, so it's up to you :)

PR #469, opened 2020-12-31
chore(a11y): add vue-axe when in dev mode

MrTimscampi:
In an effort to push us towards taking accessibility more seriously, this adds [vue-axe](https://github.com/vue-a11y/vue-axe) when in development mode.

It will run checks on the current page while it runs and point out areas where we can improve accessibility.

Here is an example of the output:
image

### DIFF
--- a/plugins/axe.ts
+++ b/plugins/axe.ts
@@ -1,0 +1,8 @@
+import Vue from 'vue';
+
+if (process.env.NODE_ENV === 'development') {
+  // eslint-disable-next-line
+  import('vue-axe').then(({ default: VueAxe }) => {
+    Vue.use(VueAxe);
+  });
+}
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -71,6 +71,8 @@ const config: NuxtConfig = {
    ** https://nuxtjs.org/guide/plugins
    */
   plugins: [
+    // Development
+    'plugins/axe.ts',
     // General
     'plugins/appInitPlugin.ts',
     'plugins/veeValidate.ts',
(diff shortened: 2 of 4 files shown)


ferferga:
@MrTimscampi Did you check with analyze if the bundle gets added in production? I'm not sure that the if is enough, I think we might need to tweak some other stuff in nuxt.config

plugins/axe.ts
```diff
@@ -0,0 +1,6 @@
+import Vue from 'vue';
+import VueAxe from 'vue-axe';
+
+if (process.env.NODE_ENV === 'development') {
+  Vue.use(VueAxe);
+}
```
Camc314, line 6: The docs use `require` after the if statement. Should we be doing that here?
```
import Vue from 'vue'

if (process.env.NODE_ENV === 'development') {
  const VueAxe = require('vue-axe').default
  Vue.use(VueAxe)
}
```

Or
```
import Vue from 'vue'

if (process.env.NODE_ENV === 'development') {
 import('vue-axe').then(({default: VueAxe}) => {
   Vue.use(VueAxe)
 })
  
}

```

ferferga:
@MrTimscampi Both production screnshots

**Your changes**

before

**With @Camc314 proposal**

after

plugins/axe.ts
ferferga, line 6:
```suggestion

if (process.env.NODE_ENV === 'development') {
  // eslint-disable-next-line @typescript-eslint/no-var-requires
  const VueAxe = require('vue-axe').default;
  Vue.use(VueAxe);
}
```

Camc314 (approved):

ferferga (approved):

PR #488, opened 2021-01-03
fix(library): trigger items refresh for user views

MrTimscampi:
For merged libraries, the CollectionType is different. This adds it to the list of valid types for refreshing items.


Camc314 (approved):
Untested, but looks correct

PR #496, opened 2021-01-04
fix: fix locales being detected from browser and being overriden

### LINKED ISSUE #494: Language reverts to English when clicking on any link
If you have a non-English language set, you'll notice the interface reverts to English when you click on any link. Though it doesn't change the setting server-side, so something must be changing the i18n value directly.

Wasn't present as of e94d67576a5d895fcda893e7e6d53cca4821f20d

Is present as of 6dc78e6356e6e4d1de726f0d33a3d488863d55c7

So it seems the 6dc78e6356e6e4d1de726f0d33a3d488863d55c7 commit introduced it, the one syncing changes to the server

ThibaultNocchi:
locale detection wasn't fully disabled and so our custom locale assignment was overriden by i18n

It uses the custom method I added to https://github.com/ThibaultNocchi/i18n-module/commit/c3f91e0dcd8a38e3a9a0d2f1f24d54e35c5e6e0b to reset the locale set

fix #494

### DIFF
--- a/store/displayPreferences.ts
+++ b/store/displayPreferences.ts
@@ -38,7 +38,12 @@ const updateMethods: { [key: string]: (value: string) => void } = {
   locale: (value: string) => {
     if (window.$nuxt) {
       if (value !== 'auto') window.$nuxt.$i18n.setLocale(value);
-      else window.$nuxt.$i18n.setLocale(navigator.language);
+      else
+        window.$nuxt.$i18n.setLocale(
+          window.$nuxt.$i18n.getBrowserLocale() ||
+            window.$nuxt.$i18n.defaultLocale ||
+            'en'
+        );
     }
   }
 };
--- a/store/displayPreferences.ts
+++ b/store/displayPreferences.ts
@@ -18,7 +18,7 @@ const defaultCustomPrefs: {
   darkMode: booleanToString(
     nuxtConfig.vuetify?.theme?.default === 'dark' || false
   ),
-  locale: nuxtConfig.i18n?.defaultLocale || 'en'
+  locale: 'auto'
 };
 
 const defaultState = (): DisplayPreferencesState => ({
@@ -36,7 +36,10 @@ const updateMethods: { [key: string]: (value: string) => void } = {
   },
 
   locale: (value: string) => {
-    if (window.$nuxt) window.$nuxt.$i18n.setLocale(value);
+    if (window.$nuxt) {
+      if (value !== 'auto') window.$nuxt.$i18n.setLocale(value);
+      else window.$nuxt.$i18n.setLocale(navigator.language);
+    }
   }
 };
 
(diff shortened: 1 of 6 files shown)


store/displayPreferences.ts
```diff
@@ -18,7 +18,7 @@ const defaultCustomPrefs: {
   darkMode: booleanToString(
     nuxtConfig.vuetify?.theme?.default === 'dark' || false
   ),
-  locale: nuxtConfig.i18n?.defaultLocale || 'en'
+  locale: navigator.language || nuxtConfig.i18n?.defaultLocale || 'en'
```
ferferga, line 21: (I'm not that familiar as you're with this library probably, so bear with me please) :P

I think here we should keep browser detection enabled in ``nuxt.config.ts``. Two reasons:

* Doing this way is *obscure*. If, in the future, we want to change something about this, we might wonder how, if ``detectBrowserLanguage`` is false in ``nuxt.config.ts``, we're getting the current locale.

* i18n team know better how to detect browser languages than we do :). There might be some edge cases that don't support what we do here.

## How I think we can accomplish this: 

As we discussed in Matrix, the problem is that login form is restored to the browser's language. With the ``beforeMount`` call, we fix part of the issue.

However, a more complete solution, imo, would be to enable the ``useCookie`` in ``nuxt.config.ts`` and use i18n's methods ``getLocaleCookie`` and ``setLocale`` when.

On initial load, i18n should have already made the cookie available, so we can replace ``navigator.language`` with a call to ``getLocaleCookie``. Then, whenever user change a language or we fetch something from server, we use ``setLocale``. If we want browser language to take place again, I think a call to ``setLocaleCookie('')`` should work.

ThibaultNocchi: > Doing this way is obscure. If, in the future, we want to change something about this, we might wonder how, if detectBrowserLanguage is false in nuxt.config.ts, we're getting the current locale.

> i18n team know better how to detect browser languages than we do :). There might be some edge cases that don't support what we do here.

I completely agree with you on both those points, but let me explain the problem with `displayPreferences`.

As those are fetched as the user is successfully logged in, it can either happen on first app load as you are reconnected, or when you manually login. This second case never was a problem.

In the first case, the store is populated by the login scheme found in `jellyfinScheme.ts` and happens really soon in the app lifecycle. As such, plugins such as `$vuetify` or `$i18n` aren't available yet and dark mode and locale aren't set in. That's why you can see a `callAllCallback()` in `beforeMount` of `default.vue`. It just retries editing those values in from the previously populated store.

Also, the scheme engages the disconnection procedure if you were previously logged and didn't have the "remember me" case checked. Which resets the `displayPreferences` store.

I didn't find how to reset the locale to the detected one after modifying it. Without this, I still need to depend on `navigator.language`.

So in order to improve the system, I'd need i18n to either have a way to be programmatically disabled or have a method to reset its locale to the browser's one.

Let me know if you have better idea or need clarifying :)

ferferga:
Blocked as we're waiting for https://github.com/nuxt-community/i18n-module/pull/1022

ferferga (approved):

PR #504, opened 2021-01-05
Fix logout

ferferga:
Sometimes, components updates triggered before the logout process completed successfully, crashing the whole client with errors.

I added a check in the two troublesome components: **HomeHeaderItemsWelcome** and **UserButton**

The first commit solves the issue completely, but in an additional commit (which I can revert if not everyone agrees with it/have other proposals), I added the chain operator to all the calls to ``this.$auth`` (except the scheme ones), so we can forget about chaining issues (stuff relying on it not being undefined will ofc still be in trouble) which can cause some problems that we're not able to reproduce.

I also tried logging out in the middle of requests being resolved (with first commit only) and I had no issues.

### DIFF
--- a/components/Buttons/UserButton.vue
+++ b/components/Buttons/UserButton.vue
@@ -7,7 +7,7 @@
         v-on="on"
       >
         <user-image />
-        <h1 class="font-weight-light pb-1 text-truncate">
+        <h1 v-if="$auth.user" class="font-weight-light pb-1 text-truncate">
           {{ $auth.user.Name }}
         </h1>
         <v-icon>mdi-chevron-down</v-icon>
--- a/components/Item/ItemMenu.vue
+++ b/components/Item/ItemMenu.vue
@@ -48,7 +48,7 @@ export default Vue.extend({
     items: {
       get(): MenuItem[] {
         const menuItems = [] as MenuItem[];
-        if (this.$auth.$state.user.Policy.IsAdministrator) {
+        if (this.$auth.user?.Policy?.IsAdministrator) {
           menuItems.push({
             title: this.$t('editMetadata'),
             action: () => {
--- a/components/Layout/HomeHeader/HomeHeaderWelcome.vue
+++ b/components/Layout/HomeHeader/HomeHeaderWelcome.vue
@@ -9,7 +9,10 @@
             class="mx-md-10 mt-md-5 mt-10 py-0 py-md-4 align-start justify-center text-center text-md-left justify-md-start"
           >
             <v-col cols="12" sm="8" md="6" xl="5" class="py-0 py-md-4">
-              <p class="text-h6 text-sm-h4 text-md-h2 text-lg-h2 text-xl-h2">
+              <p
+                v-if="$auth.user"
+                class="text-h6 text-sm-h4 text-md-h2 text-lg-h2 text-xl-h2"
+              >
                 {{
                   $t('homeHeader.welcome.helloUser', {
                     userName: $auth.user.Name
--- a/components/Buttons/FilterButton.vue
+++ b/components/Buttons/FilterButton.vue
@@ -293,7 +293,7 @@ export default Vue.extend({
       try {
         const response = (
           await this.$api.filter.getQueryFiltersLegacy({
-            userId: this.$auth.user.Id,
+            userId: this.$auth.user?.Id,
             parentId: this.$route.params.viewId,
             includeItemTypes: [this.itemsType]
           })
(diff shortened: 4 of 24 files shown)


components/Item/ItemMenu.vue
```diff
@@ -48,7 +48,7 @@ export default Vue.extend({
     items: {
       get(): MenuItem[] {
         const menuItems = [] as MenuItem[];
-        if (this.$auth.$state.user.Policy.IsAdministrator) {
+        if (this.$auth?.$state?.user?.Policy?.IsAdministrator) {
```
ThibaultNocchi, line 51:
```suggestion
        if (this.$auth?.user?.Policy?.IsAdministrator) {
```
It should work
print_screen_2021-01-05-16-33-20

ThibaultNocchi (changes requested):
Also, couldn't the linter warn/error us when we're accessing possibly undefined properties?

ferferga:
@ThibaultNocchi Yes we can, but I guess not for this PR

components/Layout/HomeHeader/HomeHeaderWelcome.vue
```diff
@@ -42,6 +42,17 @@ export default Vue.extend({
...
     }
+  },
+  computed: {
+    userName: {
+      get(): string {
+        const user = this.$auth?.user?.Name;
+        if (user) {
+          return user;
+        }
+        return '';
```
Camc314, line 53:
```suggestion
        return this.$auth?.user?.Name ? this.$auth?.user?.Name : '';

**Review by Camc314 — COMMENTED** [review]
`this.$auth` should always be defined, we only need to do `.?` after user

**Comment by ferferga** [comment]
Although I added checks in the components involved in the issue, I checked with only the changes of the last commit against master and it also solves the problem completely, as it forces the redirection to happen way before ``user`` is undefined.

It's anyway a good idea to have this type of checking imo, but it doesn't hurt if we add more safeguards in our auth scheme :)

**Comment by ferferga** [comment]
@Camc314 We should still enable ``"strictNullChecks": true`` in tsconfig.json imo.

Making a seperate PR for that at #505

File: components/Item/ItemMenu.vue (hunk 1 above)
Comment (ferferga) [reply components/Item/ItemMenu.vue:51]: @ThibaultNocchi Good catch

**Review by ThibaultNocchi — APPROVED** [review]

File: components/Buttons/UserButton.vue [hunk 3]
```diff
@@ -8,7 +8,7 @@
       >
         <user-image />
         <h1 class="font-weight-light pb-1 text-truncate">
-          {{ $auth.user.Name }}
+          {{ userName }}
```
MrTimscampi, line 11: The component is entirely useless and shouldn't even be shown if there is no user, so I don't really see the reason for this.

MrTimscampi: It's also usually nicer to do this in the template instead of creating a computed property just for a simple check like this.

components/Layout/HomeHeader/HomeHeaderWelcome.vue
```diff
@@ -12,7 +12,7 @@
               <p class="text-h6 text-sm-h4 text-md-h2 text-lg-h2 text-xl-h2">
                 {{
                   $t('homeHeader.welcome.helloUser', {
-                    userName: $auth.user.Name
+                    userName: userName
```
MrTimscampi, line 15: Same as above.

MrTimscampi (changes requested):
The two computed properties shouldn't be necessary, and if they really are, it should be handled in the template instead.

components/Buttons/UserButton.vue
ferferga: @mrtimscampi it's the one causing issues in fact, as the home header welcome one it's not active most of the time.

I can move the checks there, though, although as you mention, they are no longer *that* necessary as the logout process is fired earlier.

That's why I think it's a good idea to keep them, but in the template, just for safety.

Also, I noticed (specially when working with home header) that there are some times when *if* checks happen with some delay, so computed properties take place before, causing issues.

MrTimscampi:
Do note that we can't use optional chaining in templates, unfortunately :/

ferferga:
@MrTimscampi That's what I was looking in Vue documentation.

How can we solve this though without a computed property?

components/Buttons/UserButton.vue
```diff
@@ -8,7 +8,7 @@
       >
         <user-image />
         <h1 class="font-weight-light pb-1 text-truncate">
-          {{ $auth.user.Name }}
+          {{ $auth?.user?.Name ? $auth?.user?.Name : '' }}
         </h1>
```
MrTimscampi, line 12: This should work:
```suggestion
        <h1 v-if="$auth.user" class="font-weight-light pb-1 text-truncate">
          {{ $auth.user.Name }}
        </h1>
```

MrTimscampi (approved):

PR #505, opened 2021-01-05
chore: enable strict null checks for typescript

ferferga:


Camc314 (approved):

MrTimscampi:
According to the docs, this shouldn't be necessary, as `strict` already enabled it (along with all the other strict mode options).

See: https://www.typescriptlang.org/tsconfig#strict

What's the rationale for adding an explicit one?

ThibaultNocchi:
This was after my comment https://github.com/jellyfin/jellyfin-vue/pull/504#pullrequestreview-561890457 but as @Camc314 pointed out https://github.com/jellyfin/jellyfin-vue/pull/504#issuecomment-754755388, it's because `$auth.user` is of type any.

MrTimscampi:
Closing this, then, as it essentially does nothing and the proper fix would be fixing the typing of $auth.user :)

ferferga:
@MrTimscampi not only for that, but for future developments.

I didn't know that strict enabled it, so thank you for pointing that out!

PR #520, opened 2021-01-08
feat: Full screen music UI, playback and home header improvements

ferferga:
## Changes

* Volume slider now applies the changes immediately (before, it only was applied when the slider was released).
* Volume slider icon changes depending on the selected volume.
* Playlists work now.
* Add shuffle and repeat functionality
* Refactor image handling for further modularity and moved all the checks from ``BlurhashImage`` to ``imageHelper``.
* First draft of a full screen music player (with smooth transitions everywhere):

https://user-images.githubusercontent.com/10274099/104018417-04217300-51ba-11eb-97d8-4c6d1970ccfe.mp4

https://user-images.githubusercontent.com/10274099/104018431-0be11780-51ba-11eb-8066-048cbeb6a049.mp4

* Move progress slider to its own component.
* Refactor backdrop store. Backdrop is now fully reactive and it's opacity is fully configurable.
* Add blurhash of the current displayed item in home header to the home page:

https://user-images.githubusercontent.com/10274099/104018638-6ed2ae80-51ba-11eb-8e4a-ae15654fb439.mp4

## To do

* Improvements to responsive design (for another PR).
## Updates

Changed the miniplayer design between commits:

image

**Full screen controls**:

image

**Active controls (with elevation)**:

image

**Reduced window**:

image


components/Item/TimeSlider.vue
```diff
@@ -63,6 +62,23 @@ export default Vue.extend({
...
+       *
+       * @param {string} seconds - Number to format
+       * @returns {string} Formatted seconds number
+       */
+      function formatSeconds(seconds: string): string {
+        return ('0' + seconds).slice(-2);
+      }
+
+      return `${minutes}:${formatSeconds(seconds.toString())}`;
+    },
```
Camc314, line 81: Why is this added?

components/Item/Metadata/ImageEditor.vue
```diff
@@ -127,7 +127,7 @@ export default Vue.extend({
         })
       ).data;
     },
-    imageFormat(imageInfo: ImageInfo): string {
+    imageFormat(imageInfo: ImageInfo): string | undefined {
       return this.getImageUrlForElement(imageInfo.ImageType as ImageType, {
```
Camc314, line 131: this will cause errors if imageInfo.ImageType is undefined

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
-              class="text-truncate link"
-              :to="`/item/${getCurrentItem.AlbumId}`"
+  <transition name="fade" mode="in-out">
+    <v-footer
+      v-if="isPlaying && getCurrentlyPlayingMediaType() === 'Audio'"
+      key="audioControls-footer"
+      app
+      :color="footerColor"
+    >
+      <v-container v-if="isFullScreenPlayer" fluid>
```
Camc314, line 9: what is wrong with using `isMinimized`?

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
+        <time-slider />
+      </v-container>
+      <v-container fluid>
+        <v-row>
+          <v-col cols="9" md="3" class="d-flex flex-row pa-0">
+            <v-avatar
+              v-if="!isFullScreenPlayer"
```
Camc314, line 16: what is wrong with using `isMinimized`?

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
+              <div class="d-flex align-center justify-center">
+                <v-btn
+                  icon
+                  :input-value="isShuffling"
+                  class="mx-1"
+                  @click="toggleShuffle"
+                >
+                  <v-icon>{{
+                    isShuffling ? 'mdi-shuffle' : 'mdi-shuffle-disabled'
+                  }}</v-icon>
```
Camc314, line 65:
```suggestion
                  <v-icon>
                    {{ isShuffling ? 'mdi-shuffle' : 'mdi-shuffle-disabled' }}
                  </v-icon>

File: components/Layout/AudioControls.vue [hunk 6]
```diff
@@ -1,51 +1,122 @@
...
+                  <v-icon>mdi-stop</v-icon>
+                </v-btn>
+                <v-btn icon class="mx-1" @click="setNextTrack">
+                  <v-icon>mdi-skip-next</v-icon>
+                </v-btn>
+                <v-btn icon class="mx-1" @click="toggleRepeatMode">
+                  <v-icon>{{ repeatIcon }}</v-icon>
+                </v-btn>
+              </div>
+              <time-slider v-if="!isFullScreenPlayer" />
```
Camc314, line 85: same here

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
-            {{ getCurrentItem.AlbumArtist }}
-          </nuxt-link>
-        </div>
-      </v-col>
-      <v-col cols="6" class="pa-0 d-none d-md-inline">
-        <div class="d-flex flex-column justify-center">
-          <div class="d-flex align-center justify-center">
-            <v-btn disabled icon class="mx-1">
-              <v-icon>mdi-shuffle</v-icon>
+            <v-btn icon>
```
Camc314, line 115:
```suggestion 
            <v-btn icon disabled>

File: components/Layout/AudioControls.vue [hunk 8]
```diff
@@ -58,77 +129,106 @@
...
       'setPreviousTrack',
       'unpause',
-      'pause'
+      'pause',
+      'toggleShuffle',
+      'toggleRepeatMode'
+    ]),
+    ...mapGetters('playbackManager', [
+      'getCurrentItem',
+      'getCurrentlyPlayingMediaType'
```
Camc314, line 222: These should be computed

Camc314 (commented):
image

this album is not centred

IMHO the album art is too big, How about having the are on the LHS, and putting the queue or the playing info on the RHS.

layouts/default.vue
```diff
@@ -2,6 +2,7 @@
   <v-app ref="app">
     <backdrop />
     <v-navigation-drawer
+      v-if="$store.state.page.showNavDrawer"
```
Camc314, line 5: Why hide the nav drawer?

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
+                  <v-icon dark>mdi-album</v-icon>
+                </template>
+              </v-img>
+            </v-avatar>
+            <div class="d-flex flex-column justify-center ml-4">
+              <span class="font-weight-medium mt-md-n2">
+                <nuxt-link
+                  tag="span"
+                  class="text-truncate link"
+                  :to="`/item/${getCurrentItem()}`"
```
Camc314, line 33: this doesn't work

```suggestion
                  :to="`/item/${getCurrentItem().Id}`"
```

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
+                <template #activator="{ on, attrs }">
+                  <nuxt-link tag="span" :to="'/playback'">
+                    <v-btn icon class="ml-2" v-bind="attrs" v-on="on">
+                      <v-icon>mdi-fullscreen</v-icon>
+                    </v-btn>
+                  </nuxt-link>
+                </template>
+                <span>{{ $t('fullScreen') }}</span>
+              </v-tooltip>
+            </transition>
```
Camc314, line 109: What is wrong with expanding the modal to full screen?

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
+                  <v-icon large>
+                    {{ isPaused ? 'mdi-play' : 'mdi-pause' }}
+                  </v-icon>
+                </v-btn>
+                <v-btn icon class="mx-1" @click="stopPlayback">
+                  <v-icon>mdi-stop</v-icon>
+                </v-btn>
+                <v-btn icon class="mx-1" @click="setNextTrack">
+                  <v-icon>mdi-skip-next</v-icon>
+                </v-btn>
```
Camc314, line 80: Should we disable this if there is no next item in the queue

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
+                  }}</v-icon>
+                </v-btn>
+              </span>
+              <nuxt-link
+                tag="span"
+                class="text--secondary text-caption text-truncate mt-md-n2 link"
+                :to="`/artist/${getCurrentItem().AlbumArtists[0].Id}`"
+              >
+                {{ getCurrentItem().AlbumArtist }}
+              </nuxt-link>
```
Camc314, line 51: we should probably check if there is an album artist.

components/Layout/AudioControls.vue
```diff
@@ -1,51 +1,122 @@
...
-          </span>
-          <nuxt-link
-            tag="span"
-            class="text--secondary text-caption text-truncate mt-md-n2 link"
-            :to="`/artist/${getCurrentItem.AlbumArtists[0].Id}`"
+              <v-img :src="getImageUrl(getCurrentItem())">
+                <template #placeholder>
+                  <v-icon dark>mdi-album</v-icon>
+                </template>
+              </v-img>
```
Camc314, line 26: alt needed

components/Layout/AudioControls.vue
```diff
@@ -58,77 +129,106 @@
...
+      } else {
+        return undefined;
+      }
     },
     isPaused(): boolean {
       return this.$store.state.playbackManager.status === PlaybackStatus.paused;
+    },
+    isPlaying(): boolean {
+      return (
+        this.$store.state.playbackManager.status !== PlaybackStatus.stopped
```
Camc314, line 170: Should isPlaying be true if there is an error? Wouldn't it be better to check if  `this.$store.state.playbackManager.status` is equal to  `PlaybackStatus.playing`?

components/Layout/AudioControls.vue
```diff
@@ -58,77 +129,106 @@
...
+      } else {
+        return 'mdi-repeat-off';
+      }
+    },
+    isShuffling(): boolean {
+      return this.$store.state.playbackManager.isShuffling;
+    },
+    isFullScreenPlayer(): boolean {
+      return this.$route.name === 'playback';
+    },
```
Camc314, line 191: Why not use isMinimized?

components/Layout/AudioControls.vue
```diff
@@ -58,77 +129,106 @@
...
+    nextTrackName(): string | undefined {
```
Camc314, line 192: Would this be better as a getter in `playbackManager.ts` store

Camc314, line 192: Would this be better as a getter in `playbackManager.ts` store

components/Layout/AudioControls.vue
ferferga: @Camc314 If going from /playback to / directly in the address bar, the store doesn't get the mutation, so the fullscreen UI is still there.

I also reproduced this when switching between users/servers. We should maybe make an store call on page reload so everything gets cleared (but login info and DisplayPreferences)

layouts/default.vue
ferferga: @Camc314 it's a full screen player.

MrTimscampi:
> IMHO the album art is too big, How about having the are on the LHS, and putting the queue or the playing info on the RHS.

This is meant to be a full screen, kiosk-like experience :p There's a separate queue screen planned with a different view. Main goal here is to just show the album art and give basic info and controls, to show on like a TV screen or a wall tablet or something.

components/Layout/AudioControls.vue
ferferga: @Camc314 Swiper's weirdness, DOM performance (slides are expensive, and with items in the background, much worse) and a bug in Vuetify where the scrollbar doesn't disappear and allowed you to scroll the things that were below the v-dialog.

components/Layout/AudioControls.vue
ferferga: @Camc314 That doesn't cover pause.

ferferga:
@Camc314 Are you able to reproduce the off-center issue with the latest force-pushed changes?

Original commits already had the hack enabled, but not on mount. Check

Camc314:
> @Camc314 Are you able to reproduce the off-center issue with the latest force-pushed changes?

Yes, no change

Camc314:
> This is meant to be a full screen, kiosk-like experience :p There's a separate queue screen planned with a different view. Main goal here is to just show the album art and give basic info and controls, to show on like a TV screen or a wall tablet or something.

Forgive me, I'm don't think I quite understand. Is this layout going to be used for TVs and full screen desktops then, and a different one used for mobile and non-full screen desktops?

ferferga:
@Camc314 This is similar to what YouTube Music and Tidal has. It's just a page where you can leave the player sitting there while relaxing :P

It's like opening VLC, it also displays the cover at full screen.

Some example photos from other services:

image
image
image

Camc314:
[](url)

> @Camc314 This is similar to what YouTube Music, Tidal and Spotify has. It's just a page where you can leave the player sitting there while relaxing :P

Ok, that makes more sense

ferferga:
Updated screenshots of the new miniplayer design:

image

**Full screen controls**:

image

**Active controls (with elevation)**:

image

MrTimscampi:
A few small things before I start reviewing the code proper:

image
The context menu should probably be shown on a right-click on the cover, title or artist. It feels weird for it to be there.

If anything, we should at least invert the close/menu buttons and do this in a future PR.

image

The spacing here in the mini player is also super large. It looks a bit weird.

image

The full screen one looks correct.

On to reviewing the code now :rocket:

components/Item/Metadata/ImageEditor.vue
```diff
@@ -127,11 +127,13 @@ export default Vue.extend({
...
-    imageFormat(imageInfo: ImageInfo): string {
-      return this.getImageUrlForElement(imageInfo.ImageType as ImageType, {
-        itemId: this.metadata.Id,
-        tag: imageInfo.ImageTag as string
-      });
+    imageFormat(imageInfo: ImageInfo): string | undefined {
+      if (imageInfo.ImageType) {
+        return this.getImageUrlForElement(imageInfo.ImageType, {
+          itemId: this.metadata.Id,
+          tag: imageInfo.ImageTag as string
```
ThibaultNocchi, line 134: I don't know much this API, but is there a reason for not checking in the if the `ImageTag`?

layouts/default.vue
ThibaultNocchi: @ferferga Could it maybe be tied to fullscreen-ness of the player? Also the drawer already has a `drawer` model and it could be seen as redundant?

ThibaultNocchi: Maybe even have an event being emitted from the audio controls to play with the drawer model directly?

ferferga:
@ThibaultNocchi Don't know why but I can't answer in the thread of https://github.com/jellyfin/jellyfin-vue/pull/520#discussion_r556729723

I'll reply you here: I didn't think about the v-model honestly :). Thought about doing all the stuff in the store so it can be extensible in the future for whatever thing we might want to do, instead of deeply lock it to an specific feature.

components/Item/Metadata/ImageEditor.vue
ferferga: Good catch

ThibaultNocchi: Another thing to refactor later on haha

ThibaultNocchi:
@ferferga  that's because initially it's an answer to an existing cam review, and you'll see the same comment there and you'll be able to answer ^^ but about the navbar, maybe you could split out the `drawer` model to a computed getter/setter which is mapped to the store?

store/playbackManager.ts
```diff
@@ -201,9 +222,6 @@ export const mutations: MutationTree<PlaybackManagerState> = {
   ) {
     state.currentTime = time;
   },
-  RESET_CURRENT_TIME(state: PlaybackManagerState) {
-    state.currentTime = 0;
-  },
```
Camc314, line 206: Why remove this?

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,315 @@
...
+                  </nuxt-link>
+                </template>
+                <span>{{ $t('fullScreen') }}</span>
+              </v-tooltip>
+            </transition>
+            <v-tooltip top>
+              <template #activator="{ on, attrs }">
+                <v-btn
+                  icon
+                  class="ml-2"
```
Camc314, line 162:
```suggestion
```

Currently looks slightly off with the enlarged gap between icons

image

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,315 @@
...
+                  v-bind="attrs"
+                  v-on="on"
+                  @click="stopPlayback"
+                >
+                  <v-icon>mdi-close</v-icon>
+                </v-btn>
+              </template>
+              <span>{{ $t('stopPlayback') }}</span>
+            </v-tooltip>
+            <item-menu :item="getCurrentItem" :absolute="false" :dark="false" />
```
Camc314, line 172: This is up for discussion, but I think currently, having this here makes the menu feel a bit too cluttered, how about moving it to the main playback page

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,315 @@
...
-            <v-btn disabled icon class="mx-1">
-              <v-icon>mdi-shuffle</v-icon>
-            </v-btn>
-            <v-btn icon class="mx-1" @click="setPreviousTrack">
-              <v-icon>mdi-skip-previous</v-icon>
-            </v-btn>
-            <v-btn icon large class="mx-1" @click="togglePause">
-              <v-icon large>
+            <v-btn
+              elevation="6"
```
Camc314, line 179: I think this looks better without elevation since none of the other buttons have it.

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,315 @@
...
+            <item-menu :item="getCurrentItem" :absolute="false" :dark="false" />
+          </v-col>
+        </v-row>
+        <div
+          v-if="isFullScreenPlayer && getNextItem"
+          class="d-flex justify-center align-center"
+        >
+          <div>
+            <h4 class="text-overline font-italic">
+              {{ $t('upNext') }}: {{ getNextItem.Name }}
```
Camc314, line 233: Make this whole line translated like we have done in Card.vue L233

`$t('upNext', {upNextItemName: getNextItem.Name})`

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,315 @@
...
-      return this.ticksToMs(this.getCurrentItem.RunTimeTicks) / 1000;
+    ...mapGetters('playbackManager', [
+      'getCurrentItem',
+      'getCurrentlyPlayingMediaType',
+      'getPreviousItem',
+      'getNextItem'
+    ]),
+    footerColor(): string | undefined {
+      if (this.isFullScreenPlayer) {
+        return 'rgba(0,0,0,0.15)';
```
Camc314, line 261:
```suggestion
        return 'rgba(0, 0, 0, 0.15)';

File: pages/playback/index.vue [hunk 25]
```diff
@@ -0,0 +1,164 @@
...
+      v-if="currentQueue"
+      ref="playbackSwiper"
+      class="d-flex justify-center align-center"
+      :options="swiperOptions"
+      @slideChange="onSlideChange"
+    >
+      <swiper-slide
+        v-for="item in currentQueue"
+        :key="item.Id"
+        class="swiper-no-swiping"
```
Camc314, line 13: Why disable touch events?

pages/library/_viewId.vue
```diff
@@ -132,7 +132,7 @@ export default Vue.extend({
 
       if (
         this.collectionInfo &&
-        ['CollectionFolder', 'Folder', 'UserView'].includes(
+        ['CollectionFolder', 'Folder', 'UserView', 'playlists'].includes(
```
Camc314, line 135: AFAIK, a collection can't have type playlists?

Or at least my playlists collection doesnt have type 'playlists'

Camc314 (commented):
We are still not centred I'm afraid.

image



image

image

When clicking on some buttons, it is staying selected. (First image after initial click, second after clicking elsewhere)


Responsiveness needs to be improved in further PRs

store/playbackManager.ts
ferferga: @Camc314 Redundant, we commit time = 0 and it's the same thing. We can add it if we need it at some point, but I don't see an use case for this

pages/playback/index.vue
ferferga: @Camc314 Messes with centering when queues have only 1 or 2 items, until we don't have that fixed it's better to leave it disabled imo.

pages/library/_viewId.vue
ferferga: @Camc314 I'm not sure to be honest, playlists doesn't load for me in master and with this line here pointed out by @MrTimscampi it works

components/Layout/AudioControls.vue
ferferga: @Camc314 The idea is to have it as we do in current web at some point, with all the actions quickly accesible. I don't mind really, for mobile design though I would leave it.

components/Layout/AudioControls.vue
ferferga: @Camc314 I think it makes sense to make it more prominent since it's the main button of any music player 😋.

I agree with you in dark theme, but in light theme it also looks really plain without this.

ferferga:
@Camc314 Any idea on how to fix the buttons that keep the focus? The default behaviour it's to not keep them focused but, for some reason, it doesn't happen for these buttons.

components/Layout/AudioControls.vue
MrTimscampi: The light theme looks really weird with it elevation, imo. It's already emphasized by the fact that it's circled, too, so it feels a bit redundant.

components/Layout/HomeHeader/HomeHeaderItems.vue
```diff
@@ -204,9 +200,20 @@ export default Vue.extend({
   },
   mounted() {
     this.swiper = (this.$refs.homeSwiper as Vue).$swiper as Swiper;
+    const hash = this.getBlurhashHash(this.items[0], ImageType.Backdrop);
```
MrTimscampi, line 203:
```suggestion
    const hash = this.getBlurhash(this.items[0], ImageType.Backdrop);
```

It feels a bit redundant with two hashes.

components/Layout/TimeSlider.vue
```diff
@@ -78,3 +77,11 @@ export default Vue.extend({
   }
 });
 </script>
+
+<style lang="scss" scoped>
```
MrTimscampi, line 81: This doesn't seem to work anymore.

layouts/default.vue
MrTimscampi: I agree with Thibault, v-model already does this and it's be better to use that instead

layouts/default.vue
```diff
@@ -2,6 +2,7 @@
   <v-app ref="app">
     <backdrop />
     <v-navigation-drawer
```
MrTimscampi, line 4: It'd probably be interesting to investigate how feasible it'd be to remove the transition for this when moving to the full screen player.

The current animation when hiding the drawer looks a bit weird. Not really blocking though.

locales/en-US.json
```diff
@@ -280,5 +280,6 @@
   "writing": "Writing",
   "year": "Year",
   "years": "Years",
-  "youMayAlsoLike": "You may also like"
+  "youMayAlsoLike": "You may also like",
+  "stopPlayback": "Stop playback"
```
MrTimscampi, line 284: Can we keep this file in alphabetical order, please.

`i18n-ally` for VSCode should do this automatically for you when adding translations through it.

mixins/imageHelper.ts
```diff
@@ -64,14 +96,156 @@ declare module 'vue/types/vue' {
...
+ *
+ * @param {BaseItemDto|BaseItemPerson} obj - The object to check
+ * @returns {boolean} Returns true if the object is a person, false otherwise.
+ */
+function isPerson(obj: BaseItemDto | BaseItemPerson): obj is BaseItemPerson {
+  if ((obj as BaseItemPerson)?.Role) {
+    return true;
   }
+  return false;
 }
```
MrTimscampi, line 131:
```suggestion
function isPerson(obj: BaseItemDto | BaseItemPerson): obj is BaseItemPerson {
  return (obj as BaseItemPerson)?.Role !== undefined;
}
```

mixins/imageHelper.ts
```diff
@@ -64,14 +96,156 @@ declare module 'vue/types/vue' {
...
+     */
+    getImageTag(
+      item: BaseItemDto | BaseItemPerson,
+      type: ImageType,
+      index = 0,
+      checkParent = true
+    ): string | undefined {
+      if (item) {
+        if (!isPerson(item)) {
+          if (item.ImageTags?.[type]) {
```
MrTimscampi, line 154: What happens here when an item has a parent and should show its parent image, but also has images of its own?

mixins/imageHelper.ts
```diff
@@ -100,16 +276,32 @@ const imageHelper = Vue.extend({
...
-            'item.ImageTags must not be null or undefined when an item object is passed'
-          );
+        if (!tag) {
+          tag = this.getImageTag(item, type, backdropIndex, false);
+          if (!tag && checkParent) {
+            tag = this.getImageTag(item, type, backdropIndex, true);
+            itemId = this.getParentId(item);
+            if (!tag || !itemId) {
+              // eslint-disable-next-line no-console
+              console.error(
```
MrTimscampi, line 292: It's probably not needed. If nothing has images, it'll just spam the console with errors.

mixins/imageHelper.ts
```diff
@@ -100,16 +276,32 @@ const imageHelper = Vue.extend({
...
+                `Parent of itemId ${item.Id} doesn't have a valid tag for the image type requested`
+              );
+              return undefined;
+            }
+          } else if (!tag) {
+            // eslint-disable-next-line no-console
+            console.error(
```
MrTimscampi, line 299: Same for this.

mixins/imageHelper.ts
```diff
@@ -88,6 +262,8 @@ const imageHelper = Vue.extend({
      * @param {number} [options.maxHeight=element?.clientHeight] - Maximum height of the image in CSS pixels (optional).
      * @param {number} [options.quality=90] - Quality level of the image (optional, only relevant for jpeg format).
      * @param {boolean} [options.limitByWidth=false] - Use the element's width instead of its height for the size calculation.
+     * @param {number} [options.backdropIndex=0] - Index of the backdrop image to use (when image type is 'Backdrop').
+     * @param {boolean} [options.checkParent=true] - If it's needed to look for the parent's image tags in case the item's requested image type/tag doesn't exist.
      * @returns {string} The URL for the image, with the base URL set and the options provided.
      */
     getImageUrlForElement(
```
MrTimscampi, line 269: Fallbacks should probably be handled by this. So pass it an item and an image type, and it'll handle the fallbacks.

There's mainly one that I can think of right now, which is `thumb -> backdrop`, but I believe web also falls back to Primary for that. 

This can technically be done later, but since this refactors pretty most of the function, it should probably take this into account from the get go to avoid further refactorings.

MrTimscampi (changes requested):

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,315 @@
...
+            <div class="d-flex flex-column justify-center">
+              <div class="d-flex align-center justify-center">
+                <v-btn
+                  icon
+                  fab
+                  small
+                  :elevation="isShuffling ? '3' : '0'"
+                  class="mx-1"
+                  :class="isShuffling ? '' : 'text--disabled'"
+                  :color="isShuffling ? 'primary' : 'text--disabled'"
```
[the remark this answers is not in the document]
MrTimscampi: For reference, Spotify's desktop, player doesn't use a disabled color for these buttons. They just just go from normal color to primary when enabled.

components/Layout/AudioControls.vue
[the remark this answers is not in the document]
MrTimscampi: In the future, we could see about launching the instant mix for the album artist at the end of playback, maybe (Probably as a configurable option)

mixins/imageHelper.ts
ferferga: @mrtimscampi own item images will always take precedence, so parents will be ignored. We can add an option to force query for parents if we need it.

mixins/imageHelper.ts
ferferga: @mrtimscampi My idea was that, in case we have many possibilities for the images, use the function directly in an v-if or computed property and check if it returns undefined or not.

Remember also that BlurhashImage has now the possibility to replace images that doesn't exist with placeholders, which can be even more useful.

Same for getBlurhash function, if we're not sure there's a blurhash, we should check first calling this function directly and checking for undefined

Handling all the possible fallbacks here would make the mixin really big for no really huge improvement. Majority of components can already call the mixin without no checks, but the scarce cases where we need it we only need two lines more of code.

mixins/imageHelper.ts
MrTimscampi: That was my concern, there are a few situations where we'd want the parent's images instead of the items :)

ferferga: I said adding another option for that when I was afk, but now that I'm addressing your requests, I remembered that I also made a getParentid method for internal use of the mixin but also external. So we can call that method and pass whatever id it returns to us to the getImageUrl method without the need of further options.

components/Layout/AudioControls.vue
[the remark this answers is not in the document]
ThibaultNocchi: Yep, I'd go this route too

components/Layout/AudioControls.vue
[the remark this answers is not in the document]
ThibaultNocchi: Yeah as an option, a lot of people don't like this feature (me included)

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,308 @@
...
-              <v-icon>mdi-shuffle</v-icon>
-            </v-btn>
-            <v-btn icon class="mx-1" @click="setPreviousTrack">
-              <v-icon>mdi-skip-previous</v-icon>
-            </v-btn>
-            <v-btn icon large class="mx-1" @click="togglePause">
-              <v-icon large>
+            <v-btn
+              elevation="6"
+              fab
```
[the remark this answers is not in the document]
ThibaultNocchi: That'd also fix the focus problem, cause I think it's tied to fabs

components/Layout/AudioControls.vue
```diff
@@ -1,135 +1,315 @@
...
+                  class="mx-1"
+                  @click="togglePause"
+                >
+                  <v-icon large>
+                    {{ isPaused ? 'mdi-play' : 'mdi-pause' }}
+                  </v-icon>
+                </v-btn>
+                <v-btn
+                  icon
+                  :disabled="!getNextItem"
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi That's what happens right now but the disabled buttons also let you check at a glance if there are items in the queue. It also simplified fixing some bugs in the fullscreen player, as this disables errors like "no item in index" in the playback store and such.

layouts/default.vue
ferferga: For future reference, and as it was discussed in Matrix, we're going to leave this as an ``v-if`` so toggling the nav drawer doesn't mess with the user's behaviour. Going the v-model route would require to store another variable that contains whether the user has the navdrawer hidden or not before toggling ot, so we can act accordingly after unhiding it.

If we want to switch everything to v-model at some point, we need to take care of that.

mixins/imageHelper.ts
```diff
@@ -64,14 +96,153 @@ declare module 'vue/types/vue' {
...
+     * @param {boolean} [checkParent=true] - Looks for tag/image type for the parent if the passed item doesn't have the ImageType requested
+     * @returns {string|undefined} Returns the tag, undefined if the specific ImageType doesn't exist.
+     */
+    getImageTag(
+      item: BaseItemDto | BaseItemPerson,
+      type: ImageType,
+      index = 0,
+      checkParent = true
+    ): string | undefined {
+      if (item) {
```
Camc314, line 149: We can change this to 
`if (!item) return`
to reduce complexity

mixins/imageHelper.ts
```diff
@@ -64,14 +96,153 @@ declare module 'vue/types/vue' {
...
+        if (!isPerson(item)) {
```
Camc314, line 150: And then move the isperson handeling to the top

mixins/imageHelper.ts
```diff
@@ -64,14 +96,153 @@ declare module 'vue/types/vue' {
...
+          if (item.ImageTags?.[type]) {
```
Camc314, line 151: I've rewritten this. Feel free to use if it if you think it is a bit easier to understand:

```
      if (!item) return;

      if (isPerson(item)) {
        if (item.PrimaryImageTag && type === ImageType.Primary) {
          return item.PrimaryImageTag;
        } else {
          return;
        }
      }

      if (item.ImageTags?.[type]) {
        return item.ImageTags?.[type];
      } else if (
        type === ImageType.Backdrop &&
        item.BackdropImageTags?.[index]
      ) {
        return item.BackdropImageTags[index];
      }

      if (checkParent) {
        switch (type) {
          case ImageType.Primary:
            if (item.AlbumPrimaryImageTag) {
              return item.AlbumPrimaryImageTag;
            } else if (item.ChannelPrimaryImageTag) {
              return item.ChannelPrimaryImageTag;
            } else if (item.ParentPrimaryImageTag) {
              return item.ParentPrimaryImageTag;
            }
            break;
          case ImageType.Art:
            if (item.ParentArtImageTag) {
              return item.ParentArtImageTag;
            }
            break;
          case ImageType.Backdrop:
            if (
              item.ParentBackdropImageTags &&
              item.ParentBackdropImageTags[index]
            ) {
              return item.ParentBackdropImageTags[index];
            }
            break;
          case ImageType.Logo:
            if (item.ParentLogoImageTag) {
              return item.ParentLogoImageTag;
            }
            break;
          case ImageType.Thumb:
            if (item.ParentThumbImageTag) {
              return item.ParentThumbImageTag;
            }
        }
      }
```

ferferga: Thank you very much for proposing this, don't know why but I didn't think about using an switch here and it's much clear 👍

MrTimscampi (approved):

Camc314 (approved):

PR #521, opened 2021-01-08
Item page improvements

ThibaultNocchi:
I'm trying to improve the background code of the item page.

* Factorized the track selector into their own component, which emits an `input` event when the track is changed with the index of the track. It also chooses the default value, disabled the selector if necessary, applies placeholders... 
* Factorized the `cols` `sm` and `class` tags of the two-columns lines in the info part (genres, writing, selectors...) to easily edit them later to find the right sizing 🤌
* Warning message instead of the selectors when no media source are found
* Fix the play button not being centered anymore on mobile size


ferferga:
Take in mind that backdrop store doesn't work in master, is fixed in #520

ThibaultNocchi:
Waiting for #530 as it was rebased against

components/Item/TrackSelector.vue
```diff
@@ -0,0 +1,249 @@
...
+    item: {
+      type: Object as PropType<BaseItemDto>,
+      required: true
+    },
+    // Current media source used (current movie version for instance)
+    mediaSourceIndex: {
+      type: Number,
+      default: 0
+    },
+    // Which media type to consider for this selector
```
ferferga, line 63: We should use jsdoc syntax for these as well imo

components/Item/TrackSelector.vue
```diff
@@ -0,0 +1,249 @@
...
+  },
+  watch: {
+    /**
+     * @param {number} newVal - New index value choosen in the v-select
+     */
+    trackIndex(newVal: number): void {
+      this.$emit('input', newVal);
+    },
+    /**
+     * When the media source index is changed by the parent, we reset the selected track as they have changed
```
ferferga, line 168:
```suggestion
     * When the media source index is changed by the parent, we reset the selected track as it has changed
```

pages/item/_itemId/index.vue
```diff
@@ -155,29 +152,19 @@
...
-                v-if="
-                  item &&
-                  ((item.MediaSources && item.MediaSources.length > 1) ||
-                    videoTracks.length > 0 ||
-                    audioTracks.length > 0 ||
-                    subtitleTracks.length > 0)
-                "
-                class="mt-2"
-              >
+              <div v-if="item && item.MediaSources" class="mt-2">
```
ferferga, line 155: Why the other conditions were removed?

pages/item/_itemId/index.vue
```diff
@@ -198,136 +185,64 @@
                     </v-select>
                   </v-col>
                 </v-row>
-                <v-row v-if="videoTracks.length > 0" align="center">
+                <v-row align="center">
```
ferferga, line 188: Same

ferferga (changes requested):

pages/item/_itemId/index.vue
ThibaultNocchi: I wanted to show the different selectors even when there was nothing to pick, as having "No video/audio/sub available" is better than getting nothing at conveying the fact that the media has no tracks.

ThibaultNocchi: Though I should add a placeholder in the media source vselect

ferferga: And disable it in that case, so the "No available X" is always shown but interaction with the v-select is completely disabled.

pages/item/_itemId/index.vue
```diff
@@ -380,21 +304,27 @@
...
+interface TwoColsInfoColumn {
+  lCols: number;
+  lSm: number;
+  rCols: number;
+  rSm: number;
+  lClass: { [key: string]: boolean };
+}
 
 export default Vue.extend({
+  components: { TrackSelector },
```
ferferga, line 327: This is not needed as we have all the components autoimported

pages/item/_itemId/index.vue
```diff
@@ -198,139 +188,73 @@
...
+                  <v-col
+                    :cols="twoColsInfoColumn.rCols"
+                    :sm="twoColsInfoColumn.rSm"
+                  >
+                    <track-selector
+                      :item="item"
+                      :media-source-index="currentSourceIndex"
+                      :type="'Subtitle'"
+                      @input="currentSubtitleTrack = $event"
+                    ></track-selector>
```
ferferga, line 248:
```suggestion
                    />
```

pages/item/_itemId/index.vue
```diff
@@ -198,139 +188,73 @@
...
+                  <v-col
+                    :cols="twoColsInfoColumn.rCols"
+                    :sm="twoColsInfoColumn.rSm"
+                  >
+                    <track-selector
+                      :item="item"
+                      :media-source-index="currentSourceIndex"
+                      :type="'Audio'"
+                      @input="currentAudioTrack = $event"
+                    ></track-selector>
```
ferferga, line 228:
```suggestion
                    />
```

pages/item/_itemId/index.vue
```diff
@@ -198,139 +188,73 @@
...
+                  <v-col
+                    :cols="twoColsInfoColumn.rCols"
+                    :sm="twoColsInfoColumn.rSm"
+                  >
+                    <track-selector
+                      :item="item"
+                      :media-source-index="currentSourceIndex"
+                      :type="'Video'"
+                      @input="currentVideoTrack = $event"
+                    ></track-selector>
```
ferferga, line 208:
```suggestion
                    />
```

ferferga (changes requested):

pages/item/_itemId/index.vue
ThibaultNocchi: Done

pages/item/_itemId/index.vue
ThibaultNocchi: Done

pages/item/_itemId/index.vue
ThibaultNocchi: Done

pages/item/_itemId/index.vue
ThibaultNocchi: Done

pages/item/_itemId/index.vue
```diff
@@ -42,7 +42,7 @@
             </div>
             <v-row
               class="mt-4 align-center"
-              :class="{ 'text-center': !$vuetify.breakpoint.mdAndUp }"
+              :class="{ 'justify-center': !$vuetify.breakpoint.mdAndUp }"
             >
               <v-btn
                 v-if="canPlay(item)"
```
ThibaultNocchi, line 48: Here we're using a v-if which calls to the mixin and a disabled which calls to local method, do these overlap? @MrTimscampi

ferferga: @ThibaultNocchi they're combined, see audio controls for an example

ferferga: You can remove the braces though

ThibaultNocchi:
I've removed the icon and now the play button gets disabled when there's no media source
Screenshot_2021-01-16 Jellyfin(1)

ferferga (approved):

PR #550, opened 2021-01-14
Reset all stores at logout

### LINKED ISSUE #545: Stores don't get cleared on logout

ThibaultNocchi:
- [x]  Add a root action to call all stores reset actions
- [x]  Add reset actions to all stores
- [x]  Factorize the default state
- [x]  Call the root reset action at logout

fix #545

### DIFF
--- a/store/index.ts
+++ b/store/index.ts
@@ -1,5 +1,5 @@
 import Vue from 'vue';
-import { MutationTree } from 'vuex';
+import { MutationTree, ActionTree } from 'vuex';
 import { TvShowsState } from './tvShows';
 import { ServerState } from './servers';
 import { PageState } from './page';
@@ -13,17 +13,17 @@ import { DeviceState } from './deviceProfile';
 import { DisplayPreferencesState } from './displayPreferences';
 
 export interface AppState {
-  tvShows: TvShowsState;
-  page: PageState;
-  servers: ServerState;
-  snackBar: SnackbarState;
-  user: UserState;
-  userViews: UserViewsState;
-  homeSection: HomeSectionState;
-  playbackManager: PlaybackManagerState;
   backdrop: BackdropState;
   device: DeviceState;
   displayPreferences: DisplayPreferencesState;
+  homeSection: HomeSectionState;
+  page: PageState;
+  playbackManager: PlaybackManagerState;
+  servers: ServerState;
+  snackBar: SnackbarState;
+  tvShows: TvShowsState;
+  user: UserState;
+  userViews: UserViewsState;
 }
 
 export interface RootState {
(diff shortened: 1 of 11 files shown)


store/index.ts
```diff
@@ -69,3 +69,23 @@ export const mutations: MutationTree<RootState> = {
...
+export const actions: ActionTree<RootState, RootState> = {
+  async reset({ dispatch }, { clearCritical }: { clearCritical: boolean }) {
+    const promises = [];
+    promises.push(dispatch('backdrop/clearBackdrop', { root: true }));
+    promises.push(dispatch('deviceProfile/clearDeviceProfile', { root: true }));
+    promises.push(dispatch('displayPreferences/resetState', { root: true }));
+    promises.push(dispatch('homeSection/clearHomeSection', { root: true }));
+    promises.push(dispatch('page/clearPage', { root: true }));
+    promises.push(dispatch('playbackManager/stop', { root: true }));
+    if (clearCritical)
```
ferferga, line 82: For a one line thing it's good to do it this way, but in this huge block I would use the brace style

ferferga (changes requested):

store/index.ts
ThibaultNocchi: Wdyn? The if only applies to the server clear. Or I misunderstand you :p

ferferga: I meant to scope the if block in braces:

if (whatever) {
//Whatever
}

ThibaultNocchi: Done

store/index.ts
```diff
@@ -69,3 +69,24 @@ export const mutations: MutationTree<RootState> = {
...
+    promises.push(dispatch('playbackManager/stop', { root: true }));
+    if (clearCritical) {
+      promises.push(dispatch('servers/clearServers', { root: true }));
+    }
+    promises.push(dispatch('snackbar/resetMessage', { root: true }));
+    promises.push(dispatch('tvShows/clearTvShows', { root: true }));
+    promises.push(dispatch('user/clearUser', { root: true }));
+    promises.push(dispatch('userViews/clearUserViews', { root: true }));
+
+    await Promise.all(promises);
```
ferferga, line 90: Sorry for not pointing it out before, but why ``Promise.all``? ``await`` at each of the dispatches should also run each of them after the last one, without the need of pushes and arrays.

ThibaultNocchi: It's not doing the same thing. Doing `await` before every dispatch means that we wait for the first one to finish before starting the second. `Promise.all` returns a promise that's resolved when all of the promises passed as argument are resolved. So here we launch the different dispatch concurrently and wait for them all to finish. The speed gain here is rather small, but there's no need to wait for each and every one of them before starting the next one ;)

ferferga (approved):
Tested and works!

Rebase so we can merge this :)

PR #552, opened 2021-01-14
fix(darkmodetoggle): tooltip adaptive to current mode

ThibaultNocchi:


locales/en-US.json
```diff
@@ -252,7 +252,8 @@
   "title": "Title",
   "tooltips": {
     "changeLanguage": "Language",
-    "toggleDarkMode": "Toggle dark mode"
+    "darkMode": "Dark mode",
```
ferferga, line 255: What about "Switch to X mode"?

ThibaultNocchi: I think it's not needed, I checked the first similar example I had at hand (Flood) and it only said "dark theme" "light theme"

ThibaultNocchi: But that's really a preference haha

ferferga: Imo it's clearer and 100% explicit, without place to imagination.

If an user it's in dark mode and see then tooltip as "Light mode" he/she might wonder what mode it's currently in. For a lot of people this is obvious but maybe not so much for elderly or color blind :).

With switch there's no possible doubt.

ThibaultNocchi: Done

ferferga (approved):

PR #556, opened 2021-01-14
Small pre release fixes

MrTimscampi:
* Remove the background from the welcome message
* Fix missing editMetadata message in en-US
* Cleanup valid library types for reusability
* Fix undefined user property on UserButton in some cases
* Link episode page from the season tabs instead of playing it
* Use light menu color on cards when in light theme (Needs a light theme PR to properly work, weirdly enough)
* Fix footer color in the full screen player when in light mode

### DIFF
--- a/components/Item/Card.vue
+++ b/components/Item/Card.vue
@@ -92,6 +92,7 @@ import { mapActions } from 'vuex';
 import { BaseItemDto, ImageType } from '@jellyfin/client-axios';
 import imageHelper from '~/mixins/imageHelper';
 import itemHelper from '~/mixins/itemHelper';
+import { validLibraryTypes } from '~/utils/items';
 
 export default Vue.extend({
   mixins: [imageHelper, itemHelper],
@@ -140,10 +141,7 @@ export default Vue.extend({
   computed: {
     itemLink: {
       get(): string {
-        if (
-          this.item.Type === 'Folder' ||
-          this.item.Type === 'CollectionFolder'
-        ) {
+        if (this.item.Type && validLibraryTypes.includes(this.item.Type)) {
           return `/library/${this.item.Id}`;
         } else if (this.item.Type === 'Person') {
           return `/person/${this.item.Id}`;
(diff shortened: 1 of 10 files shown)


components/Item/Card.vue
```diff
@@ -324,7 +322,7 @@ export default Vue.extend({
 }
 
 .card-content {
-  background-color: #{map-get($material-dark, 'menus')};
+  background-color: #{map-get($material-light, 'menus')};
```
ThibaultNocchi, line 325: How does it look like on cards with dark image?

locales/en-US.json
```diff
@@ -43,6 +43,7 @@
   "discNumber": "Disc {discNumber}",
   "dislikes": "Dislikes",
   "edit": "Edit",
+  "editMetadata": "Edit metadata",
```
ThibaultNocchi, line 46: Tabulation is weird, isn't it?

components/Item/Card.vue
MrTimscampi: It only shows up when the image and the blurhash isn't loaded.

Currently, it shows up transparent, because for some reason it's not picking up the variable, but the idea is to avoid having a super dark card in the middle of a light page.

locales/en-US.json
MrTimscampi: Wdym?

ferferga:
Leaving this as a collection of all the small things I found while using the client and I can remember right now:

**Library browser**:

* Two filtering menus can be opened at the same time if they are opened while the UI is loading, right after the mini freeze that happens before items are populated. Maybe we can hide those until everything is populated?
* Can't browse movie Collections: ``Id undefined``

**Login:**

* Server selection appears for a glance when inputting the server address.

**General**: Some strings are not lowercase in the characters that are not in the first position. We might need to find a way to force strict checks in Weblate or in our linter config for this.

ferferga:
Additional stuff:

image

**Is this how we want cards?**
image

locales/en-US.json
ThibaultNocchi: Sorry, I was on the phone and all the other lines weren't "tabuled" the same

ThibaultNocchi:
@ferferga Maybe this could be split into its own bullet point issue, or the issue to track the first release points?

ThibaultNocchi (approved):

PR #567, opened 2021-01-15
fix(norwegian_locale): change the i18n locale.name for nb-NO

### LINKED ISSUE #559: Language selector shows only "Bokmål", not "Norsk" or "Norsk Bokmål"
"Bokmål" is a form of Norwegian but I think it's incorrect to show it without "Norsk bokmål" or some variant that starts with the main language.

Wikipedia for reference:
http://puu.sh/H70qY/e369f3cf47.png

Vue:
http://puu.sh/H70rz/499dd97a04.png

For ease of copying
```
Norsk bokmål
Norsk nynorsk
```

~Other alternatives~ doesn't sort well
```
Norsk
Nynorsk
```

AlfHou:
Change the locale name of nb-NO from 'Bokmål' to 'Norsk'

Fixes #559

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -155,7 +155,7 @@ const config: NuxtConfig = {
       { code: 'en', iso: 'en-US', name: 'English', file: 'en-US.json' },
       { code: 'es', iso: 'es-ES', name: 'Español (España)', file: 'es.json' },
       { code: 'fr', iso: 'fr-FR', name: 'Français', file: 'fr-FR.json' },
-      { code: 'nb', iso: 'nb-NO', name: 'Bokmål', file: 'nb_NO.json' },
+      { code: 'nb', iso: 'nb-NO', name: 'Norsk', file: 'nb_NO.json' },
       { code: 'nl', iso: 'nl-NL', name: 'Nederlands', file: 'nl.json' },
       { code: 'pl', iso: 'pl-PL', name: 'Polski', file: 'pl.json' },
       { code: 'ro', iso: 'ro-RO', name: 'Română', file: 'ro.json' },


ferferga:
Shouldn't we add nynorsk as indicated in the linked issue?

We might not be able to translate it, but somebody can take the translations over in Weblate.

nuxt.config.ts
```diff
@@ -155,7 +155,7 @@ const config: NuxtConfig = {
       { code: 'en', iso: 'en-US', name: 'English', file: 'en-US.json' },
       { code: 'es', iso: 'es-ES', name: 'Español (España)', file: 'es.json' },
       { code: 'fr', iso: 'fr-FR', name: 'Français', file: 'fr-FR.json' },
-      { code: 'nb', iso: 'nb-NO', name: 'Bokmål', file: 'nb_NO.json' },
+      { code: 'nb', iso: 'nb-NO', name: 'Norsk', file: 'nb_NO.json' },
```
MrTimscampi, line 158:
```suggestion
      { code: 'nb', iso: 'nb-NO', name: 'Norsk bokmål', file: 'nb_NO.json' },
```

This is the proper name for it. As there are two variants of "Norsk", we should explicitly mention which one it is.

MrTimscampi (changes requested):

MrTimscampi:
> Shouldn't we add nynorsk as indicated in the linked issue?

We'd need to wait until someone on Weblate translated it, then we add it. It's pretty useless to add an empty language just for the sake of it.

MrTimscampi (approved):
Saw the discussion about this on the other issue, so it's all good. We can change it if someone comes around and starts doing a Nynorsk translation.

hawken93:
Nice :+1:

PR #578, opened 2021-01-16
fix(video-player): video container does not scale

### LINKED ISSUE #569: SD videos not scaled correctly
Running master branch of `jellyfin-vue` and `jellyfin-server`
Tested on both chrome and firefox

When viewing SD videos they are scaled incorrectly (see fig. 1). This does not happen on HD movies (fig. 2)

The SD video is an AVC @ 480 x 360 video
The HD video is an AVC @ 1920 x 816 video

image
_Fig. 1: SD Image_

image
_Fig.2: HD Image_

AlfHou:
Fixes the issue where SD videos did not scale to fill the screen

Fixes #569 

Tested on chrome and firefox. Verified that HD still works as well (see screenshot)

**SD**
image

**HD**

### DIFF
--- a/components/Players/VideoPlayer.vue
+++ b/components/Players/VideoPlayer.vue
@@ -1,6 +1,6 @@
 <template>
   <v-container fill-height fluid class="pa-0 justify-center">
-    <div ref="videoContainer">
+    <div ref="videoContainer" style="height: 100%; width: 100%">
       <video
         ref="videoPlayer"
         :poster="poster"
--- a/components/Players/VideoPlayer.vue
+++ b/components/Players/VideoPlayer.vue
@@ -1,16 +1,14 @@
 <template>
   <v-container fill-height fluid class="pa-0 justify-center">
-    <div id="videoContainer" ref="videoContainer">
-      <video
-        ref="videoPlayer"
-        :poster="poster"
-        autoplay
-        @timeupdate="onVideoProgress"
-        @pause="onVideoPause"
-        @play="onVideoProgress"
-        @ended="onVideoStopped"
-      />
-    </div>
+    <video
+      ref="videoPlayer"
+      :poster="poster"
+      autoplay
+      @timeupdate="onVideoProgress"
+      @pause="onVideoPause"
+      @play="onVideoProgress"
+      @ended="onVideoStopped"
+    />
   </v-container>
 </template>
 
@@ -237,10 +235,5 @@ video {
   max-height: 100vh;
   width: 100%;
   height: 100%;
-}
-#videoContainer {
-  height: 100%;
-  width: 100%;
-  overflow: hidden;
 }
 </style>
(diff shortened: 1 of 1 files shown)


AlfHou:
Note that I tried using the `object-fit` as suggested in the issue, but it didn't work

ferferga:
What about mini player (pip)?

AlfHou:
@ferferga good call!
**SD**
image

**HD**
image

**Master for comparison**
image


It seems like I am getting some kind of scrollbar. I'll need to make some changes

AlfHou:
Gone :)

image

Everything else seems to work properly as well

components/Players/VideoPlayer.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <v-container fill-height fluid class="pa-0 justify-center">
-    <div ref="videoContainer">
+    <div
+      ref="videoContainer"
+      style="height: 100%; width: 100%; overflow: hidden"
```
MrTimscampi, line 5: Could we use a class for this, please?

Also, wouldn't `overflow: hidden` cut part of the video in some cases? (mini-player for example, since you had a scrollbar before, I'm assuming the video was too big for the dialog)

MrTimscampi (changes requested):
Could you also reword the commit to better respect the format used elsewhere?

The context (`sd_small` in your commit) should be the component you edited, in kebab case. So here, a good commit message would be `fix(video-player): video container does not scale` or something similar.

components/Players/VideoPlayer.vue
MrTimscampi: We might also just be able to remove this `div` and use the `v-container` instead, transferring the props from this one to its parent.

AlfHou: > Could you also reword the commit to better respect the format used elsewhere?
> 
> The context (`sd_small` in your commit) should be the component you edited, in kebab case. So here, a good commit message would be `fix(video-player): video container does not scale` or something similar.

Sure, my bad.

> Could we use a class for this, please?

I tried it the first time and probably forgot to refresh properly because it didn't work. I'll change it to an id

> Also, wouldn't overflow: hidden cut part of the video in some cases? (mini-player for example, since you had a scrollbar before, I'm assuming the video was too big for the dialog)

I don't think so. I tested it with a few files again now, and as seen in the image below it seems like it only exposes some padding bellow.
image

AlfHou: > We might also just be able to remove this `div` and use the `v-container` instead, transferring the props from this one to its parent.

Just tried this and it worked flawlessly! I will push the change

AlfHou: Screenshots:

**SD**
image

**HD**
image


**PiP**
image

MrTimscampi (approved):

Issue #495, opened 2021-01-04
Audio controls play button doesn't update when resuming via mpris


Maxr1998:
When resuming playback via mpris (e.g. `playerctl`) the play button in the audio controls doesn't update, it works when pausing.

This probably happens because the PlaybackManager state isn't updated.

A possible fix could be adding a call to `unpause` in `onAudioProgress` at [AudioPlayer.vue#L193](https://www.github.com/jellyfin/jellyfin-vue/tree/master/components%2FPlayers%2FAudioPlayer.vue#L193).

PR #582, opened 2021-01-17
Add MediaMetadata support

MrTimscampi:
This adds the necessary MediaMetadata object to get proper metadata for the audio and video players.

Screenshot from 2021-01-17 06-19-07

Note: While officially recommended, the typings file for the MediaSession API seems to have a few issues with regards to the MediaMetadata class. Errors have been silenced until we find a fix.

### DIFF
--- a/components/Players/PlayerManager.vue
+++ b/components/Players/PlayerManager.vue
@@ -146,8 +146,10 @@
 </template>
 
 <script lang="ts">
+import { ImageType } from '@jellyfin/client-axios';
 import Vue from 'vue';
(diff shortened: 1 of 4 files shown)


ferferga:
The times I played with MediaSession I used this package: https://www.npmjs.com/package/@types/wicg-mediasession

Could we also have MediaSession and not MediaMetadata alone, so we fix #495 and 100% finish MediaSession API, right now this is a bit incomplete.

MrTimscampi:
That was planned for after, since it may be a bit more involved :)

Camc314 (approved):

PR #589, opened 2021-01-17
feat(localeswitcher): add automatic locale picking

ThibaultNocchi:
As i18n allows us to detect the locale, we could have an item in the picker to pick an `auto` one which would follow the browser setting. This is a draft implementation which has drawbacks:

* The page doesn't change language until it's fully refreshed or we click on the auto language again. Right now `displayPreferences` does the work of setting the locale when it sees the `auto` keyword when updating the preferences, or loading them from the server. So changing the browser's language with the JF client opened wouldn't change the one in the client. Though this is due to how i18n works, as it sets the locale to the cookie and only changes when we call its `setLocale` method.
* In the same vein, the text in the picker doesn't change without refreshing or clicking on `Auto` yet again.

This is how it looks like:
print_screen_2021-01-17-16-44-21

I'm waiting for when we'll have a better implementation of the display preferences where the store only handles fetching/updating the server and we'll have those client locales react to the store.

### DIFF
--- a/components/System/LocaleSwitcher.vue
+++ b/components/System/LocaleSwitcher.vue
@@ -19,9 +19,11 @@
     </template>
     <v-list class="overflow-y-auto">
       <v-list-item
-        v-for="(item, index) in $i18n.locales"
-        :key="index"
-        :input-value="item.code === $i18n.locale"
+        v-for="item in locales"
+        :key="item.code"
+        :input-value="
+          item.code === $store.state.displayPreferences.CustomPrefs.locale
+        "
         @click="editCustomPref({ key: 'locale', value: item.code })"
       >
         <v-list-item-title>{{ item.name }}</v-list-item-title>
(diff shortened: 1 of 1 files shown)


ferferga:
I think that it's not possible to detect in a reliable way that the user changed the browser's languague without a refresh as you say, at least not without an interval or something polling in the background, which imo is extreme for such an edge case.

I tested a few pages and none of them react to the language change of the browser without a refresh, so I don't think we should be concerned about this :)

ThibaultNocchi:
You're right, let's close it for now :)

ferferga:
@ThibaultNocchi I think I worded myself wrong ^^"

I actually think this is a good PR, in fact after #632 is merged the first problem will be gone. In fact it should be a matter of just add the "Auto" option to locale switcher and that's it.

What I meant about the browser thing is that we shouldn't have headaches with it and simply leave it as it's not 100% our fault and nobody does this correctly either, so :)

I'll reopen this as I'm interested in working in this PR if you aren't. If you're still interested, ping me :).

ThibaultNocchi:
In that case I think I'll be able to finish this off :)

PR #590, opened 2021-01-17
fix(login): fix some responsive problems

ThibaultNocchi:
On lower ends of some breakpoints, user cards were too narrow. Also, on really narrow (<500px) screens the buttons and locale picker disappeared on the right.

# Current
print_screen_2021-01-17-17-35-42
print_screen_2021-01-17-17-35-28
print_screen_2021-01-17-17-35-19

# PR
Screenshot_2021-01-25 Jellyfin(1)
Screenshot_2021-01-25 Jellyfin

### DIFF
--- a/components/System/LocaleSwitcher.vue
+++ b/components/System/LocaleSwitcher.vue
@@ -6,7 +6,8 @@
           <v-btn
             :icon="!fab"
             :fab="fab"
-            :small="fab"
+            :small="!large & fab"
+            :large="large"
             :class="{ 'mr-n1': !fab, 'ml-1': fab }"
             v-bind="{ ...attrsMenu, ...attrsTooltip }"
             v-on="{ ...onMenu, ...onTooltip }"
@@ -37,7 +38,11 @@ export default Vue.extend({
   props: {
     fab: {
       type: Boolean,
-      required: true
+      required: false
+    },
+    large: {
+      type: Boolean,
+      required: false
     }
   },
   methods: {
--- a/pages/login.vue
+++ b/pages/login.vue
@@ -3,8 +3,8 @@
     <v-row align="center" justify="center">
       <v-col
         v-if="isEmpty(currentUser) && !loginAsOther && publicUsers.length > 0"
-        sm="7"
-        md="6"
+        sm="10"
+        md="7"
         lg="5"
       >
         <h1 class="text-h4 mb-6 text-center">{{ $t('selectUser') }}</h1>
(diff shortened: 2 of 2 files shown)


ferferga:
@ThibaultNocchi What do you think about putting the locale icon at the bottom page when it's not in the same line of the buttons?

I think it will look better than having it floating around.

MrTimscampi:
For this and narrower:
image

I would think it'd be better to do two lines:
```
Manual login
Change server - Locale switcher
```

ThibaultNocchi:
I can't put it at the bottom *when* it's put on the second line, but I can always put in on the bottom when the breakpoint is XS (the only breakpoint where the locale picker is on the second line sometimes)

ThibaultNocchi:
Ok your idea MrTim would solve the problem of the locale picker being put on the second line

ThibaultNocchi:
@MrTimscampi  Feels kinda weird to me though 
print_screen_2021-01-18-07-46-06

MrTimscampi:
Both rows need to have the same width, that's why it feels weird.

Look at the server selection screen, it should work that way already, if I'm not mistaken.

ferferga:
@MrTimscampi I think it doesn't make sense to mix it in any of the layouts to be honest: the buttons are for the login process, the switcher is for changing locale.

They serve different purposes so, as it's goals are different, I would say it makes sense to have a differentiation UI wise.

MrTimscampi:
@ferferga It's more for consistent sizing across the various screens.

They all follow a similar layout:

| Title |
|-------|
| Form |
| Buttons |

So putting that layout in a dedicated component that would be shared between these would ensure we get a consistent layout across the board.

ferferga:
@MrTimscampi But why not leave the switcher centered, alone, at the bottom in all the layouts?

Most webpages have the locale switcher in the bottom of the page as well, so I also think it will be more expected to users to have it there. 

The fact that there are also 3 buttons (when usually most of the times in all login forms are 1 or 2) also makes it a bad place imo. In fact, when we add the "Forgot password" one I think it should always be in a second line, in all layouts.

I don't mind the current place but it's something that's been floating around my mind since the first time I saw that button.

ThibaultNocchi:
I updated the first message with new screens

components/System/LocaleSwitcher.vue
```diff
@@ -37,7 +38,13 @@ export default Vue.extend({
   props: {
     fab: {
       type: Boolean,
-      required: true
+      required: false,
+      default: false
```
MrTimscampi, line 42: This is unnecessary, since the default is always false (It works like HTML attributes)

components/System/LocaleSwitcher.vue
```diff
@@ -37,7 +38,13 @@ export default Vue.extend({
...
+    },
+    large: {
+      type: Boolean,
+      required: false,
+      default: false
```
MrTimscampi, line 47: Same here

MrTimscampi (approved):
LGTM, with two small things that aren't necessary :)

ferferga (approved):
Tested and LGTM!

Address @MrTimscampi comments and I think we can merge this :)

components/System/LocaleSwitcher.vue
ThibaultNocchi: Done

PR #591, opened 2021-01-17
fix: reset playback before starting a new one and transitions to logs & activity

ferferga:
* Add transitions to the two tables of logs and activity, as they take a while to load if there's a lot of data.
* Removed custom transitions in favor of vuetify's ones, which should reduce bundle size (as we're using less styles)
* Playback store is reset when playing a new item or starting a new queue.


pages/settings/logsAndActivity.vue
```diff
@@ -2,88 +2,114 @@
...
-          </v-card-title>
-          <v-card-text v-if="loadingLogsStatus.errorMessage">
-            {{ loadingLogsStatus.errorMessage }}</v-card-text
+        <v-fade-transition group>
+          <h2 key="logs-title" class="text-h6 mb-2">
+            {{ $t('logsAndActivity.logs') }}
+          </h2>
+          <v-list
+            v-if="logFiles && logFiles.length > 0"
+            key="log-list"
```
ThibaultNocchi, line 11: What are the `key` for without a loop rendering?

ThibaultNocchi (commented):
Nice catch on the playback manager! :p

pages/settings/logsAndActivity.vue
ferferga: @thibaultnocchi it's needed for transitions to work.

ThibaultNocchi (approved):

PR #592, opened 2021-01-17
fix(itemmenu): don't force dark theme

masterkoppa:
Don't force dark theme on ItemMenu

## Current
image

## Fixed
image

## Note
This is my first PR against the new jellyfin-vue repository, so please let me know if there's anything I did wrong or anything I could do better in the future. I found the issue while exploring the new interface. It's looking great!

### DIFF
--- a/components/Item/Card/Card.vue
+++ b/components/Item/Card/Card.vue
@@ -79,7 +79,7 @@
             class="card-lower-buttons d-flex justify-center align-center"
           >
             <like-button v-if="canPlay(item)" :item="item" dark />
-            <item-menu :item="item" />
+            <item-menu :item="item" dark />
           </div>
         </div>
       </div>
--- a/components/Item/ItemMenu.vue
+++ b/components/Item/ItemMenu.vue
@@ -66,7 +66,7 @@ export default Vue.extend({
     },
     dark: {
       type: Boolean,
-      default: true
+      default: false
     },
     outlined: {
       type: Boolean,
--- a/components/Layout/AudioControls.vue
+++ b/components/Layout/AudioControls.vue
@@ -143,7 +143,7 @@
                 <span>{{ $t('fullScreen') }}</span>
               </v-tooltip>
             </v-fade-transition>
-            <item-menu :item="getCurrentItem" :dark="false" />
+            <item-menu :item="getCurrentItem" />
           </v-col>
           <v-col
             cols="3"


ferferga (changes requested):
This doesn't fix the problem.

Afaik, the ``dark`` prop is to indicate that we want that to be in the *inverted* color theme we're using. i.e: be black on light theme and white in dark theme

So, if you remove this, it might look nice in ligh theme, but not in dark. Also, I remember I added that prop because there were specific issues with card overlays, where it looked way too dark in ligh theme. Did you check that?

Try the following:
* Add the prop to the ``v-icon`` instead.
* Change ``HomeHeaderWelcome`` so the item menu component has the ``dark`` prop as false.

MrTimscampi:
> Afaik, the `dark` prop is to indicate that we want that to be in the _inverted_ color theme we're using. i.e: be black on light theme and white in dark theme

No, it overrides the link to the theme of the `v-app`. So the icon will be white in dark mode and black in light mode.

Since we want to be able to pass the dark prop in some places (Card overlays, for example), the proper fix here would be to set the dark prop to false here: https://github.com/jellyfin/jellyfin-vue/blob/master/components/Layout/HomeHeader/HomeHeaderItems.vue#L124

Or better yet, default it to false and specify true where needed (Which is the proper way props should work)

masterkoppa:
@ferferga:
I'm not super familiar with the dark/light theme setup here, but from my observation by being `dark = true` by default (and having no overrides) the button was stuck in the dark theme regardless of the setting. With the fix, things looked correct in both light and dark theme.

> default it to false and specify true where needed (Which is the proper way props should work)

That's fair, happy to make that change. Not sure where it should be set to `true`, but I'll take a look around on where it's used.

masterkoppa:
@MrTimscampi:

Correct me if I'm wrong, but this looks like it's a one off Component for the edit metadata dropdown on a specific item page. (The link you posted is for the home header one, which is different). With this being a one off, do you still think it should have a `dark` prop for overrides?

(Sorry if this sounds wrong / dumb, my background is in React not Vue)

ThibaultNocchi:
There are for now 4 instances of this component, and it is intended to be used almost everywhere we need the item menu. Do note that the HTML tags are `<item-menu>`

> That's fair, happy to make that change. Not sure where it should be set to true, but I'll take a look around on where it's used.

If you invert the default `dark` data, you'll need to also invert the `dark` prop on all four `<item-menu>`. So everywhere there was `:dark="false"` you drop it, and everywhere there's nothing you add `:dark="true"`.

MrTimscampi:
> The link you posted is for the home header one, which is different

Ah, indeed, sorry, I posted in passing, late at night :laughing: 

As @ThibaultNocchi said, it's not a one-off, it's intended to be used everywhere we have the three-dots menu, like cards, the item pages, the audio controls, etc (It's already used in most of those places)  
So the override makes sense in this context, since we use them in different places.

As for inverting the prop, the idea is to follow what standard HTML is doing: not having a property makes it be "false". Adding it means it's true (So you can do things like `<item-menu :item="item" dark />`). This is more consistent with what people would expect (I should've caught it when reviewing the original PR that added the component).

> Sorry if this sounds wrong / dumb, my background is in React not Vue

No worries, we've all pretty much been learning Vue on the fly as we work on this :) You contribution is appreciated.

ThibaultNocchi (approved):

ferferga (approved):

PR #604, opened 2021-01-19
Add optional SSR support to the build config

MrTimscampi:
This adds a `NUXT_SSR` environment variable, allowing one to build the client with server-side rendering, both in dev and in production.

Some adjustments have been made to the config in regards to loading plugins only on the client, as well as to the audio and video players, since Shaka Player requires to be imported on the client only.

**Note for users: As part of making this the best client it can, this switches the default docker image to use a NodeJS-hosted version of the client, with server-side rendering. As a result, the docker image is a bit heavier in both memory usage and size. Though it should run just fine on a Raspberry Pi, resource-conscious users can use the `*-static` variants of the docker images, which reproduce the old behavior of building a static version of the client and hosting it with Nginx.**


MrTimscampi:
@Shadowghost Requesting your review for the docker image. It seems really huge, but I'm not sure if it's correct or can be optimized.

MrTimscampi:
be03a9c fixes the initial Nuxt loading page in dev mode:

image

Camc314:
Should the Docker build be failing?

components/Players/PlayerManager.vue
```diff
@@ -312,6 +310,9 @@ export default Vue.extend({
       }
     });
   },
+  beforeMount() {
+    this.supportedFeatures = getSupportedFeatures();
+  },
```
Camc314, line 315: Is this done client/server side?

It needs to be done client.

plugins/appInit.ts
```diff
@@ -1,10 +0,0 @@
-import { Plugin } from '@nuxt/types';
-
-const appInitPlugin: Plugin = (context) => {
-  const { serverUsed } = context.store.state.servers;
-  if (serverUsed) {
-    context.$axios.setBaseURL(serverUsed.address);
-  }
-};
-
-export default appInitPlugin;
```
Camc314, line 10: Why is this deleted?

components/Players/PlayerManager.vue
MrTimscampi: It seems done client-side here. At least it's not crashing like it was in created().

Ideally though, we should move this to a plugin and define it as client-side that way.

plugins/appInit.ts
MrTimscampi: It was unused, there's an appInitPlugin.ts with the same contents, which was the one being imported in the config.

MrTimscampi:
So, about Docker, here is a summary of the changes:

The `latest` and `unstable` tags now point to the SSR version of the client. Since server-side rendering is generally a better experience (once we iron out the few kinks remaining), it makes sense to use this as the default.  
These images are also labeled as `*-ssr` variants.

For all images, a `*-static` variant is provided, which is essentially the old image.

.ci/azure-pipelines-build.yml
```diff
@@ -1,110 +1,122 @@
...
           script: |
             docker buildx use --builder builder
-            docker buildx build --file ./Dockerfile.dev --no-cache --push \
+            docker buildx build --file ./Dockerfile --no-cache --push \
               --platform linux/arm,linux/arm64,linux/amd64 \
-              --tag jellyfin/jellyfin-vue:unstable-$(Build.BuildNumber) \
+              --tag jellyfin/jellyfin-vue:unstable-ssr-$(Build.BuildNumber) \
+              --tag jellyfin/jellyfin-vue:unstable-ssr \
               --tag jellyfin/jellyfin-vue:unstable .
+            docker buildx build --file ./Dockerfile --no-cache --push \
```
crobibero, line 96:
```suggestion
            docker buildx build --file ./Dockerfile.static --no-cache --push \
```

.ci/azure-pipelines-build.yml
```diff
@@ -1,110 +1,122 @@
...
             docker buildx use --builder builder
-            docker buildx build --file ./Dockerfile.dev --no-cache --push \
+            docker buildx build --file ./Dockerfile --no-cache --push \
               --platform linux/arm,linux/arm64,linux/amd64 \
-              --tag jellyfin/jellyfin-vue:$(Version) \
-              --tag jellyfin/jellyfin-vue:stable \
+              --tag jellyfin/jellyfin-vue:$(Version)-ssr \
+              --tag jellyfin/jellyfin-vue:stable-ssr \
               --tag jellyfin/jellyfin-vue:latest .
+            docker buildx build --file ./Dockerfile --no-cache --push \
```
crobibero, line 112:
```suggestion
            docker buildx build --file ./Dockerfile.static --no-cache --push \
```

.ci/azure-pipelines-build.yml
```diff
@@ -1,110 +1,122 @@
...
-      vmImage: 'ubuntu-latest'
+      vmImage: "ubuntu-latest"
 
     steps:
       - task: NodeTool@0
-        displayName: 'Install Node'
+        displayName: "Install Node"
         inputs:
-          versionSpec: '12.x'
+          versionSpec: "12.x"
```
crobibero, line 12: Should probably use same nodejs version as docker?
I don't know what the differences are between 12.x and 14, but I would expect one to have better performance / be better supported

h1dden-da3m0n:
Regarding the Dockerfiles, AFAIK the use of `COPY` is preferred over `ADD`, unless you have a explicit need of one of its 'additiona' functionality. Especially since you use a mix out of `COPY` and `ADD`, however that does not affect the size of the resulting image.

Dockerfile
```diff
@@ -1,25 +1,33 @@
...
-# Install build dependencies for node modules
-RUN apk add git python make g++
-
-# Set workdir
 WORKDIR /app
 
-# Add package.json and yarn.lock
+RUN apk add --no-cache --virtual .build-deps git python make automake autoconf g++ libpng-dev libtool nasm file
+
 ADD package.json yarn.lock ./
```
h1dden-da3m0n, line 7:
```suggestion
COPY package.json yarn.lock /app
```

Dockerfile
```diff
@@ -1,25 +1,33 @@
...
 
-# Install dependencies
-RUN yarn install
+RUN yarn install --frozen-lockfile
 
-# Copy resources
 ADD . .
```
h1dden-da3m0n, line 11:
```suggestion
COPY . /app
```

Dockerfile
```diff
@@ -1,25 +1,33 @@
...
+# Build final image
+FROM node:14-alpine
+
+WORKDIR /app
+
+ADD .docker/package.json .docker/nuxt.config.js ./
+
+# Copy client files from the build image
+COPY --from=build /app/.nuxt ./.nuxt
+COPY --from=build /app/static ./static
```
h1dden-da3m0n, line 26:
```suggestion
COPY --from=build /app/.nuxt /app/.nuxt
COPY --from=build /app/static /app/static
```

Dockerfile
```diff
@@ -1,25 +1,33 @@
...
+# Build SSR app for production in standalone mode
+
+RUN yarn build:ssr --production --standalone
+
+# Build final image
+FROM node:14-alpine
+
+WORKDIR /app
+
+ADD .docker/package.json .docker/nuxt.config.js ./
```
h1dden-da3m0n, line 22:
```suggestion
COPY .docker/package.json .docker/nuxt.config.js /app
```

Dockerfile.static
```diff
@@ -0,0 +1,25 @@
+FROM node:14-alpine AS build
+
+# Install build dependencies for node modules
+RUN apk add git python make g++
+
+# Set workdir
+WORKDIR /app
+
+# Add package.json and yarn.lock
+ADD package.json yarn.lock ./
```
h1dden-da3m0n, line 10:
```suggestion
COPY package.json yarn.lock /app
```

Dockerfile.static
```diff
@@ -0,0 +1,25 @@
...
+
+# Install dependencies
+RUN yarn install
+
+# Copy resources
+ADD . .
```
h1dden-da3m0n, line 16:
```suggestion
COPY . /app
```

h1dden-da3m0n (commented):
Regarding the Dockerfiles, AFAIK the use of `COPY` is preferred over `ADD`, unless you have a explicit need of one of its 'additiona' functionality. Especially since you use a mix out of `COPY` and `ADD`, however that does not affect the size of the resulting image.

ferferga (approved):
LGTM

h1dden-da3m0n (approved):

Dockerfile
```diff
@@ -1,25 +1,33 @@
...
 
-# Add package.json and yarn.lock
-ADD package.json yarn.lock ./
+COPY .docker/package.json .docker/nuxt.config.js ./
 
-# Install dependencies
-RUN yarn install
+# Copy client files from the build image
+COPY --from=build /app/.nuxt ./.nuxt
+COPY --from=build /app/static ./static
```
ThibaultNocchi, line 26: Not sure about the requirements of `./` before the dests (I'm just nitpicking)

Dockerfile.static
```diff
@@ -0,0 +1,25 @@
+FROM node:14-alpine AS build
+
+# Install build dependencies for node modules
+RUN apk add git python make g++
+
+# Set workdir
+WORKDIR /app
+
+# Copy package.json and yarn.lock
+COPY package.json yarn.lock ./
```
ThibaultNocchi, line 10: Again nitpicking, but I'm not sure the last `/` is required

README.md
```diff
@@ -68,10 +68,15 @@ $ yarn install
...
+
+# build for production
+# you will need a web server to host the client
 $ yarn build
-$ yarn start
 
-# generate static project
-$ yarn generate
+# build for production with server-side rendering and launch server
+$ yarn build:ssr
```
ThibaultNocchi, line 79: I'd detail this part saying that we can copy the `.nuxt` and `static` dirs somewhere else and run the `yarn install` with the custom package.json file to save space

ThibaultNocchi (approved):

Dockerfile
MrTimscampi: The linter in VSCode gave me shit if I didn't use them :yum:

README.md
```diff
@@ -68,10 +68,37 @@ $ yarn install
...
+
+When using the SSR version of the client, you can run the client in standalone mode in order to save space.
+
+```bash
+# install dependencies
+$ yarn install
+
+# build for production with server-side rendering and launch server
+$ yarn build:ssr
+$ yarn start:ssr
```
ThibaultNocchi, line 94: You can remove this I guess

README.md
```diff
@@ -68,10 +68,37 @@ $ yarn install
...
+
+# move the server and required files to a dedicated directory
+$ mkdir -p /opt/jellyfin-vue
+$ cp .nuxt .docker/nuxt.config.js .docker/package.json /opt/jellyfin-vue
+
+# install the required dependency and start the client
```
ThibaultNocchi, line 100:
```suggestion
# install the required runtime dependency and start the client
```

ThibaultNocchi (changes requested):

README.md
ferferga: ``required dependencies``?

ThibaultNocchi: There's only one though :)

ThibaultNocchi: https://github.com/jellyfin/jellyfin-vue/blob/272c6072929a4f1e804ddc81752c30e193532f64/.docker/package.json

Dockerfile
h1dden-da3m0n: Personally I would use absolute target paths (so its clear at first sight where things end up), however this is just IMHO

README.md
MrTimscampi: Why? It's different than the standalone mode.

ThibaultNocchi: But those instructions are already there a few lines above. But maybe they're also needed in standalone?

ThibaultNocchi: But there already are instructions for the non standalone mode above, isn't it?

package.json
```diff
@@ -4,9 +4,12 @@
...
   "scripts": {
-    "dev": "nuxt-ts",
-    "build": "nuxt-ts build",
-    "start": "nuxt-ts start",
+    "dev": "NUXT_SSR=true nuxt-ts",
+    "dev:static": "nuxt-ts",
+    "build": "NUXT_SSR=true nuxt-ts build",
+    "build:static": "nuxt-ts build",
+    "start": "NUXT_SSR=true nuxt-ts start",
+    "start:static": "nuxt-ts start",
```
ThibaultNocchi, line 12:
```suggestion
    "dev": "NUXT_SSR=1 nuxt-ts",
    "dev:static": "nuxt-ts",
    "build": "NUXT_SSR=1 nuxt-ts build",
    "build:static": "nuxt-ts build",
    "start": "NUXT_SSR=1 nuxt-ts start",
    "start:static": "nuxt-ts start",
```
Usually env variables are set to 1

ThibaultNocchi (approved):
If the CI part works it's good to me

Shadowghost (approved):
Seems fine to me too.
We could still merge the two docker jobs in the CI definition to prevent some redundancy but that's a matter of preference. It's not really faster as it is right now (compared to a merged approach) because the CI just copies some files to the images, so registering binfmt and the other stuff most likely takes longer than the actual copying.

PR #607, opened 2021-01-19
feat(players): throttle time updates to one every 500ms

ThibaultNocchi:


components/Players/VideoPlayer.vue
```diff
@@ -47,7 +48,10 @@ export default Vue.extend({
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       player: null as any,
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      ui: null as any
+      ui: null as any,
+      throttledOnVideoProgress: null as DebouncedFunc<
```
MrTimscampi, line 52: Why not define a method instead of this?

ThibaultNocchi: As we discussed, the method way cause the error about `this`. So I used this which is error-free. But the methods work and I can disable the errors

ferferga: @mrtimscampi I thought the same at first but I checked the branch out and I discovered that the problem is not exactly with this because we could always do: ``throttle(this.updateTime, 500);``

However, it's not a problem about this or anything, it's a matter that throttle returns a function, which is the one we need to run, not ``throttle`` itself.

components/Players/VideoPlayer.vue
```diff
@@ -4,7 +4,7 @@
       ref="videoPlayer"
       :poster="poster"
       autoplay
-      @timeupdate="onVideoProgress"
+      @timeupdate="throttledOnVideoProgress"
       @pause="onVideoPause"
       @play="onVideoProgress"
```
ferferga, line 9:
```suggestion
      @play="onPlay"
```

ferferga (changes requested):
Unrelated to this PR but can be fixed here, so there's a PR less to review :P:

VideoPlayer has the exact same problem I fixed the other day in #596 

Change @play event, so it goes directly to a method that calls unpause on the store. See the relevant PR for more details in the implementation, it's just a matter of mirror the changes in AudioPlayer to VideoPlayer

ferferga (changes requested):
Just had another idea on how to make all of this cleaner: set the throttled function directly in data(), like this: https://github.com/jellyfin/jellyfin-vue/blob/master/components/Layout/SwiperSection.vue#L62

The variable doesn't depend on DOM elements or anything that is required in the mounted hook, so we can perfectly set it since the beginning.

components/Players/VideoPlayer.vue
ThibaultNocchi: Is this what you had in mind? a86039b75509fd5925211c8c37719851c3375dea

ThibaultNocchi: I also took the liberty to apply my shortcut to your previous fix in the audio player, tell me if it bothers you

components/Players/AudioPlayer.vue
```diff
@@ -2,16 +2,17 @@
   <audio
     ref="audioPlayer"
     autoplay
-    @timeupdate="onAudioProgress"
+    @timeupdate="throttledOnAudioProgress"
     @pause="onAudioPause"
-    @play="onPlay"
+    @play="unpause"
```
ferferga, line 7:
```suggestion
    @play="onPlay"
```

As discussed [here](https://matrix.to/#/!flDUDlxzYhNVyLEcSA:bonifacelabs.ca/$HSWHNictJYCEaOyNqRF6tbX57JqPxSA53kOMYibIcDo?via=matrix.org&via=matrix.possumlodge.me&via=bonifacelabs.ca), let's stick to the conventional names for event callbacks: ``onX``

ferferga (changes requested):

ferferga:
Also, time slider works chuggier with this, maybe we should consider removing https://github.com/jellyfin/jellyfin-vue/blob/master/components/Layout/TimeSlider.vue#L82-L88 as part of this PR

components/Players/VideoPlayer.vue
ThibaultNocchi: Done

components/Players/AudioPlayer.vue
ThibaultNocchi: Done

ThibaultNocchi:
@ferferga Removed the snippet you pointed me to but it doesn't seem much better

https://user-images.githubusercontent.com/1619359/105690980-f596c380-5efc-11eb-995b-db0564a2a19d.mp4

ferferga:
@ThibaultNocchi Hmm, probably better to leave it then, so manual changes to the slider appear instantaneously.

MrTimscampi (approved):
I'd prefer to define the throttled function as a method, because this doesn't seem clean at all, but at least it works.

ThibaultNocchi:
@MrTimscampi This is how it's done now, I was just doing it when you commented ;)

MrTimscampi (approved):

ferferga (approved):

PR #609, opened 2021-01-19
feat: queue management, add to queue, play next and global shuffle

ferferga:
#### Add play and shuffle buttons to music related libraries to start playback at a glance:

image

#### Queue manager in window:

https://user-images.githubusercontent.com/10274099/105100317-6f6a1f80-5aad-11eb-8f0e-c5f3a2ca82ae.mp4

image

*Screenshot is more recent, video shows a non-finished design*

* Refactor playbackUtils and timeUtils

* Playback manager can now be made aware about the origin of the playback event, so it allows us to display this kind of text:

   * When playback is started from library or any non-specified place:
       
image

   * When playback is started from library or any non-specified place and it's shuffled since the beginning of it (shuffle button in miniplayer !== start shuffled):

image

   * When playback is started from album or playlist (without shuffle):

image

   * When playback is started from album or playlist (shuffled):

~~*This is not yet implemented. The equivalent in current Jellyfin Web would be to click on the item menu and select "Shuffle"*. **However, the logic it's already done in playbackmanager and the following text will be displayed**:~~ **See below for the implementation**

           ``Playing {playlist|album|item} in shuffle``

## To do (for another PR)

* ~~Add to queue and play next in item menu~~ **See below for the implementation**

* Full screen player queue (this design is mainly intended for the miniplayer)

* ~~Make like buttons~~ (**done**) and save queue buttons work

* ~~Add runtime information, highlight on hover, like button to each of the tracks (same information shown in tracklist)~~ **Done in this PR**

* Batch select of tracks to reorder tracks in batch.

* Remove items from queue.

### Extra stuff

* Added like buttons to cards

* Removed overlay and link from card in item pages.


ferferga:
I'm a noob at tests, so any advice on how to implement my new changes in our testing suite is welcome :)

ferferga:
I'm a noob at tests, so any advice on how to implement my new changes in our testing suite is welcome :)

hamburglar2160:
So for scenarios where the playback queue isn't created from artist or album view, could we get the album artist listed here? Very cool to have the origin recorded.

ferferga:
@hamburglar2160 Album artists are already displayed below the song name at the left of the audio controls.

hamburglar2160:
True, but for users managing a queue that includes more than one artist (so, probably added from a playlist), having that information is helpful. For instance, if I add all the tracks named "round midnight" to my queue, I'll have no idea which of the 17 artist's version is next it until it's playing.

ferferga:
@hamburglar2160 See the To do section of the main post.

ferferga:
Adding ``playNext`` and ``addToQueue`` resulted being easier than expected. Fixing tests was also easy. **New features**:

* Added icons to item menu

* Item menu when content **is not** playing:

image

* Item menu when content **is** playing:

image

* Awareness of playbackManager when shuffling item **(see queue manager header)**:

image

locales/en-US.json
```diff
@@ -142,6 +144,16 @@
   "people": "People",
   "person": "Person",
   "play": "Play",
+  "playback": {
+    "stopPlayback": "Stop playback",
+    "playbackSource": {
+      "item": "Playing from {item}",
+      "shuffle": "Playing in shuffle mode",
+      "shuffleItem": "Playing {item} in shuffle",
```
hamburglar2160, line 152: "shuffle mode" to match previous line.

ferferga: Thank you for pointing that out.

Actually the latter is more recent than the former, so the former should be "Playing in shuffle".

The reasoning behind is that I think we should call "shuffle mode" what's enabled by the button on the player to shuffle the actual queue, while keeping "shuffle" term alone for starting new playbacks with shuffled tracks in the beginning.

Will change it tomorrow.

hamburglar2160: A viable alternative for starting new playbacks could be "shuffle play"

ferferga: @hamburglar2160 I'm going to use that one for the button's tooltips 👍.

ferferga:
Final design for now:

image

**Implemented favoriting as well**

hamburglar2160:
At some point, I'd consider hiding the overflow menus and making them visible when hovering over the items to match the album page.

ferferga:
@hamburglar2160 Considered and tested but looks awkward while dragging stuff and content pops a bit.

ThibaultNocchi:
Nice work, here are my 2 cents:
* Bug: playing a song "activates" (the list item class) all instance of the same song in the list 
print_screen_2021-01-22-17-22-53
* The queue icon seems to be a little bit too high, maybe it's how the icon is designed though
* UX wise, clicking outside of the queue should close it, and the cross can be mistaken for a clear queue button
* Idk if it's possible to hide the "ghost" element being taken. jf-web has it, but maybe we could avoid it (or not)? https://user-images.githubusercontent.com/1619359/105516069-bcc8d580-5cd5-11eb-8616-d4fcb0183bae.mp4
* I'd start the track number at 1, I think you just used the track index in the queue so you can just add 1 :P
* I'd add a draggable icon somewhere on each line, cause it's not noticeable we can rearrange. Maybe it could be the track number switching to a draggable icon when hovering a track?
* I'd remove the fav and more button for the "nav bar" of the opened queue as it seems redundant print_screen_2021-01-22-17-24-10
* Can the list slide to the current playing item when opening it?
* Also the queue takes some time to appear, is this normal? What can we do for this? If there's nothing, I think the queue `v-btn` could use some `loading` prop

I think that's all ^^'

ferferga:
* Good catch!
* Probably how it's designed. For me, it looks normal, maybe I'm not paying that much of attention 😅. I didn't change it at all, just removed the disabled prop for it.
* The idea, as discussed with @Perseusdehond in Matrix the other day, is to allow you to browse your library and add tracks to the queue and allow you to rearrange how that queue is going in one go, without the need of reopening the queue every time. You might want your queue to end at an specific hour, you can keep adding tracks and stop once the time marker reaches what you want. With the window being closable, we're loasing all this powerful stuff that, afaik, it's only possible in this player, no other has similar features (except GPM, now dead). It took me a while to get used to it as well, even after designing it purposefully, but after I got accostumed, it was really nice. 

We can add an option to change that behaviour at some point, in fact I think we must, but that's for another pass imo. About the icon, there will be another icon specifically for clearing the queue, which will be obvious. But for another PR as well, this is already massive :P.

*(In fact, if you think about it more, don't you see strange that, even with only one item, the queue window height is always the same size? That's why: to allow more items to be added with the window opened)*

* Yes, it's possible, noted.
* For a second pass, I thought about keeping the current track always as 0 and the previous ones as negative numbers (like Gelli). Also taking in account @Perseusdehond suggestions about only counting "end at" time from the current item, not the entire queue.
* I tried adding all the fav and item menu icons on hover only (like in tracklist, see @hamburglar2160 suggestion above) but it doesn't play nicely with the draggable stuff. I can try it again, but no guarantees.
* It isn't: the nav bar of the queue doesn't show you the fav or menu icons of the current item: **they're for the item used to start the playback.** So, for instance, if you hit like there, and you started the playback from a playlist, it will favorite the playlist. Maybe you're confused because you're playing items from the same album in your screenshots and that makes it "unclearer" maybe :P.
* Good point, and yes, it can
* Yes and your idea, although would be great and I thought of it, won't do anything: **the reason of the delay is that the DOM nodes are being created** after opening the queue (and destroyed 500 ms after closing it, so they free up memory but they're still visible during the transition). As the DOM is busy, the loading icon will be frozen. I explored using our virtual scroller (worst idea ever), vuetify's virtual scroller and virtual list packages: much better, but vue draggable doesn't get the correct item. So, if you drag item in position 5, it will ignore the movement and skip to the next track. There's a package that combines both but doesn't seem to be given so much love, so I want first to try some other stuff. But that's another thing for another PR I think :P

I already tried to give loading "notifications" everywhere possible for huge playlists (as clicking play or shuffle in the main library took a while for 1800 tracks), but, unfortunately, until we go the virtual scroller route, impossible here.

components/Layout/AudioControls.vue
```diff
@@ -156,14 +143,14 @@
                 <span>{{ $t('fullScreen') }}</span>
               </v-tooltip>
             </v-fade-transition>
-            <item-menu :item="getCurrentItem" :absolute="false" :dark="false" />
+            <item-menu :item="getCurrentItem" :dark="false" />
             <v-tooltip top>
```
MrTimscampi, line 147: This would probably be best replaced by a `Clear` button in the queue dialog.

components/Players/DraggableQueue.vue
```diff
@@ -0,0 +1,131 @@
+<template>
+  <v-list>
```
MrTimscampi, line 2: This should have a `v-list-item-group`.

components/Players/DraggableQueue.vue
```diff
@@ -0,0 +1,131 @@
...
+    <draggable v-model="queue" v-bind="dragOptions" class="list-group">
+      <v-hover
```
MrTimscampi, line 4: The usability of this is a bit confusing.

There's a grab handle on the left when hovering, which makes sense, but you can click and hold anywhere to move it, yet a single click plays the item.

Not sure how to make it make sense, but as-is it feels weird.

components/Players/DraggableQueue.vue
```diff
@@ -0,0 +1,131 @@
...
+        v-slot="{ hover }"
+        class="pa-0 ma-0"
+      >
+        <v-list-item ripple class="pa-0 ma-0" @click="onClick(index)">
+          <v-list-item-action
+            v-if="!hover"
+            class="list-group-item d-flex justify-center d-flex text-caption"
+            :class="{ 'primary--text': isPlaying(index) }"
+          >
+            {{ index }}
```
MrTimscampi, line 16: This starts counting from 0. Humans usually start counting at 1 :yum:

MrTimscampi: It'd also make sense to have this index change to some icon to indicate that playback is in progress on this item. The color already suggests it, but making it more explicit would be nice, I think.

components/Players/DraggableQueue.vue
```diff
@@ -0,0 +1,131 @@
...
+              {{ getArtists(item) }}
+            </v-list-item-subtitle>
+          </v-list-item-content>
+
+          <v-list-item-action>
+            <favorite-button :item="item" />
+          </v-list-item-action>
+          <v-list-item-action class="mr-2">
+            <item-menu :item="item" />
+          </v-list-item-action>
```
MrTimscampi, line 46: These are superfluous, since both of these are already right below the dialog.

components/Players/DraggableQueue.vue
```diff
@@ -0,0 +1,131 @@
...
+  },
+  methods: {
+    ...mapGetters('playbackManager', ['getCurrentItem']),
+    ...mapActions('playbackManager', ['setNewQueue', 'setCurrentIndex']),
+    isPlaying(index: number): boolean {
+      return index === this.$store.state.playbackManager.currentItemIndex;
+    },
+    getArtists(item: BaseItemDto): string | null {
+      if (item.Artists) {
+        return item.Artists.join(' - ');
```
MrTimscampi, line 89: Why a dash and not a simple comma?

Spotify does it well, imo:
image

mixins/timeUtils.ts
```diff
@@ -92,6 +104,56 @@ const timeUtils = Vue.extend({
...
+     *
+     * @param {number} ticks - Ticks of the item to calculate
+     * @returns {string} The resulting string
+     */
+    getRuntimeTime(ticks: number): string {
+      const ms = this.ticksToMs(ticks);
+      return this.$dateFns.formatDuration(
+        intervalToDuration({ start: 0, end: ms }),
+        {
+          format: [this.$t('hours'), this.$t('minutes')]
```
MrTimscampi, line 143: The documentation for date-fns doesn't suggest these should be translated. They seem to be simple references to Duration object properties.

pages/library/_viewId.vue
```diff
@@ -26,6 +26,17 @@
         :items-type="viewType"
         @change="onChangeFilter"
       />
+      <v-divider inset vertical class="mx-2 hidden-sm-and-down" />
```
MrTimscampi, line 29:
```suggestion
      <v-spacer />
```

This aligns the playback buttons to the right, keeping them consistent with the Genre pages.

MrTimscampi (changes requested):
Not too fan of some of the design decisions.

The initiator concept assumes a very specific kind of playback scenario, which won't fit with how most people consume music nowadays.  
See for example this:
image

It tells me I started playback from the Kitrist album, which is completely useless since I'm queuing random albums manually.

Storing where playback started it good, in that it allows us to refine the TrackList component, but showing it here isn't very useful, in my opinion.

There are also a lot of buttons on that dialog, most of which aren't necessary since they're replicated in the controls below.

Overall, I find Spotify or Tidal's way of doing this much nicer and cleaner:
image
image

This feels more like something out of a Google product than something out of Jellyfin Vue, if that makes sense.

I left more detailed comments for all the points that bug me.

MrTimscampi:
Here is how Tidal deals with origin item in case stuff is added to the playlist:
image

MrTimscampi:
Sorry for the triple post, but I also just noticed that there is no way to remove something from the queue, which is very inconvenient.

I can add stuff to it and move it around, but I can't remove anything from it.

hamburglar2160:
Qobuz, too, which ties for Tidal with "really well executed UX"
image

components/Buttons/LibraryPlayButton.vue
```diff
@@ -0,0 +1,64 @@
+<template>
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi The circular progress is necessary imo because here it takes like 5 seconds to load the 2000 tracks I have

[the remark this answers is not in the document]
MrTimscampi: There's already a loading indicator for the cards themselves, though, so the play button could just show up?

Or better yet, disable it while stuff isn't loaded.

ferferga:
@MrTimscampi Thank you very much for your thorough review. Will take a look at it later this afternoon ^^.

Some quick thoughts so:

* You are comparing this to Tidal and Spotify UIs. However this isn't comparable to these UIs: all of them are displayed in a "fullscreen" format, which is not what this implementation tries to accomplish. It's meant for a "quick review and editing" of the current queue. On another pass I'm going to design the fullscreen UI that will be used in mobile and will have all the features in desktop (runtime, item menu, tags, etc), just like those services. This one will have most of the "crowded" stuff removed in favor of that fully--featured design after that, **but just wanted to make the first implementation as complete as possible for daily usage for the preview users, which will also be useful for feedback**

Implementing all the ideas for responsive design that I have would make this PRs massive :).

* **About the closing/pin stuff**: read third point [here](https://github.com/jellyfin/jellyfin-vue/pull/609#issuecomment-765672404)
The idea of using close was exactly for the reasons you mentioned (we use it everywhere already) but I liked the idea of @ThibaultNocchi [in matrix](https://matrix.to/#/!flDUDlxzYhNVyLEcSA:bonifacelabs.ca/$W9gJ_PjS1rnQLFrqo6_iylxf2Qtadt4iIc3uylCSQj8?via=matrix.org&via=matrix.possumlodge.me&via=bonifacelabs.ca)

MrTimscampi:
>     * You are comparing this to Tidal and Spotify UIs. However this isn't comparable to these UIs: all of them are displayed in a "fullscreen" format, which is not what this implementation tries to accomplish

Tidal isn't displayed in a full screen dialog, it's a floating dialog like yours, just taller :wink:

ferferga:
Didn't know that ^^, thank you 👍🏻.

components/Buttons/LibraryPlayButton.vue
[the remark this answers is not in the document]
ferferga: @MrTimscampi No, I mean after clicking on it. They're already disabled while content is loading. The time between pressing the button and starting playback is what it takes more than 5 seconds.

https://user-images.githubusercontent.com/10274099/105629927-2831b500-5e46-11eb-8611-bf857d939ae4.mp4

mixins/timeUtils.ts
ferferga: Those strings were there before my PR, just untranslated. Can't find the documentation that suggests how to use this method, can you point it to me please?

components/Item/Card.vue
```diff
@@ -291,6 +311,23 @@ export default Vue.extend({
...
 <style lang="scss" scoped>
 @import '~vuetify/src/styles/styles.sass';
+.nuxt-link {
+  text-decoration: none;
+  color: inherit;
+}
+
+.link-disabled {
+  user-select: none;
+  pointer-events: none !important;
```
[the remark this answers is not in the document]
ferferga: Exactly what you mention is what I was looking for, thanks!

components/Item/ItemMenu.vue
```diff
@@ -1,9 +1,8 @@
 <template>
   <div v-if="items.length > 0">
-    <v-menu :absolute="absolute">
+    <v-menu :z-index="zIndex" top>
```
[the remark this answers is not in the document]
ferferga: This is not for that, but for making menus not overlap each other. If we don't add z-index, there's no way to open item menus in the queue (they're behind the menu). 

The absolute positioning was for the cards mainly, and the styling of it were moved to the card component itself (the only place where it was used btw): https://github.com/jellyfin/jellyfin-vue/pull/609/files#diff-c3ada57447b6cebb331b0776249d8d75790a6fde39fdc3efc69f95a5ab663a21R325-R328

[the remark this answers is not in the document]
MrTimscampi: The absolute property here specifies that the menu opens where the mouse click happens, which is nicer for these that overlapping the button, that's what I meant ;)

The absolute you're pointing out is a different thing entirely (for positioning the button itself)

mixins/timeUtils.ts
MrTimscampi: Here it is: https://date-fns.org/v2.16.1/docs/formatDuration

They're not strings that are shown to the user, they simply specify what parts of the duration to show. The locale parameter is what's used to define the language.

components/Players/DraggableQueue.vue
ferferga: Spotify doesn't have a handle and everybody understand how it works. It's nice to have but imo we're good without it, people is capable of figuring that stuff themselves.

components/Buttons/LibraryPlayButton.vue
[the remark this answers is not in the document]
MrTimscampi: Ah, I hadn't seen that since it's instant for me.

Then the circular progress is fine, but I still think they should be proper buttons on desktop, to fit better with all the rest :yum:

components/Players/DraggableQueue.vue
ferferga: That's another PR with howler maybe

components/Players/DraggableQueue.vue
ferferga: Disabled them for the current playing item

components/Buttons/LibraryPlayButton.vue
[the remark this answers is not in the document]
ferferga: @Mrtimscampi Yeah, I'll label this as blocked so your PR goes first

components/Buttons/FavoriteButton.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <v-btn :absolute="absolute" icon @click.stop.prevent="toggleFavorite">
+    <v-icon medium :class="fav ? 'red--text' : ''">
```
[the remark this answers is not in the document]
ferferga: With red it makes it more noticeable across all the UI, specially in light mode. That's why I used a combination of outline and full: people who doesn't use favorite sthat much will not find anything too outstanding of the outlined icons, while people who might use it more will find it nicer to check at a glance what's favorited. In a tracklist with a lot of items one can't notice which ones are enabled if scrolling fast for instance

components/Buttons/FavoriteButton.vue
```diff
@@ -0,0 +1,59 @@
...
+      required: true
+    },
+    absolute: {
+      type: Boolean,
+      default: false
+    }
+  },
+  data() {
+    return {
+      fav: false
```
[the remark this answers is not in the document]
ferferga: The idea was to give instant feedback of the action and reverting it back in case it fails, for instance. 

Thanks for pointing that out, because I had in my todo list the catching of errors and I forgot about it, will add it in my follow up PR.

components/Buttons/FavoriteButton.vue
```diff
@@ -0,0 +1,59 @@
...
+    };
+  },
+  watch: {
+    item: {
+      immediate: true,
+      handler(): void {
+        if (this.item.UserData?.IsFavorite) {
+          this.fav = this.item.UserData.IsFavorite;
```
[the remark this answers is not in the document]
ferferga: Explained above

components/Buttons/QueueButton.vue
```diff
@@ -0,0 +1,173 @@
...
+          <v-btn icon v-bind="attrs" v-on="{ ...tooltip, ...menu }">
+            <v-icon>mdi-playlist-play</v-icon>
+          </v-btn>
+        </template>
+        <span>{{ $t('queue') }}</span>
+      </v-tooltip>
+    </template>
+    <v-card>
+      <v-list>
+        <v-list-item>
```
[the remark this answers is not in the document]
ferferga: I added another divider to make it clearer :)

components/Item/Card.vue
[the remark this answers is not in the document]
ferferga: Already mentioned in Matrix, but paging this here for reference: https://github.com/nuxt/nuxt.js/issues/2949

components/Item/ItemMenu.vue
```diff
@@ -47,33 +56,69 @@ export default Vue.extend({
...
-      dialog: false
+      metadataDialog: false
     };
   },
   computed: {
+    ...mapGetters('playbackManager', ['getCurrentItem']),
     items: {
       get(): MenuItem[] {
         const menuItems = [] as MenuItem[];
         if (this.$auth.user?.Policy?.IsAdministrator) {
```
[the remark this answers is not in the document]
ferferga: I think the separators are for another PR, this needs a cleanup anyway imo

ferferga:
Waiting for #605

hamburglar2160:
One other queue to look at from this inspired subsonic client, taken from https://sublimemusic.app/#features
image

MrTimscampi (approved):

Darkwolf1515:
Absolutely beautiful, I’m so happy the vue client ended up working out.

PR #617, opened 2021-01-24
Use fetch() for data fetching in components

MrTimscampi:
Replaces our manual async/await API calls + loading states with Nuxt's built-in `fetch()` hook, which handles all of that for us and is more flexible and standardized than what we were doing before.

### DIFF
--- a/pages/genre/_itemId/index.vue
+++ b/pages/genre/_itemId/index.vue
@@ -1,32 +1,37 @@
 <template>
   <div>
-    <v-app-bar shrink-on-scroll flat fixed class="second-toolbar">
-      <v-toolbar-title class="d-flex genre-toolbar">
-        <span>{{ genre.Name }}</span>
-        <v-fade-transition>
-          <play-button v-if="loaded" :item="genre" />
-        </v-fade-transition>
-        <v-btn
-          v-if="loaded"
-          class="play-button mr-2"
-          min-width="8em"
-          outlined
-          rounded
-          nuxt
-          :to="`./${genre.Id}/shuffle`"
-        >
-          {{ $t('playback.shuffleAll') }}
-        </v-btn>
-      </v-toolbar-title>
+    <v-app-bar fixed flat dense class="second-toolbar">
+      <span class="text-h6 hidden-sm-and-down">
+        {{ genre.Name }}
+      </span>
+      <v-spacer />
+      <v-fade-transition>
+        <play-button v-if="!$fetchState.pending" :item="genre" />
+      </v-fade-transition>
+      <v-btn
+        v-if="!$fetchState.pending"
+        class="play-button mr-2"
+        min-width="8em"
+        outlined
+        rounded
+        nuxt
+        :to="`./${genre.Id}/shuffle`"
+      >
+        {{ $t('playback.shuffleAll') }}
+      </v-btn>
     </v-app-bar>
-    <v-container class="second-toolbar-follow">
-      <v-row v-if="!loaded">
+    <v-container class="after-second-toolbar">
+      <v-row v-if="$fetchState.pending">
         <v-col cols="12" class="card-grid-container">
           <skeleton-card v-for="n in 24" :key="n" />
         </v-col>
       </v-row>
-      <item-grid v-if="items.length" :items="items" />
-      <v-row v-else-if="loaded" justify="center">
+      <item-grid
+        v-if="items.length"
+        :items="items"
+        :loading="$fetchState.pending"
+      />
+      <v-row v-else-if="!$fetchState.pending" justify="center">
         <v-col cols="12" class="card-grid-container empty-card-container">
           <skeleton-card v-for="n in 24" :key="n" boilerplate />
         </v-col>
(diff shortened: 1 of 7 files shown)


Dockerfile.dev
```diff
@@ -1,4 +1,16 @@
...
+
+COPY .docker/package.json .docker/nuxt.config.js ./
+
+# Copy client files from the build image
+COPY dist/ ./.nuxt/
+
+# Install runtime dependencies
+RUN yarn install --production
+
+EXPOSE 80
```
dkanada, line 14: If we're going to bother with a useless line only good for documentation, might as well use a `VOLUME` directive as well to persist state even when someone forgets to create a volume.

MrTimscampi: These are changes from the SSR PR, not this one :p I'm only basing this branch on that PR to make sure it works correctly in both cases ;)

MrTimscampi:
I'm leaving the HomeHeader and the library page for subsequent PRs, since they're quite a bit larger and deserve some proper refactoring.

pages/login.vue
```diff
@@ -60,6 +60,14 @@ import { UserDto } from '@jellyfin/client-axios';
 export default Vue.extend({
   layout: 'fullpage',
   middleware: 'serverMiddleware',
+  async asyncData({ $api }) {
+    const publicUsers = (await $api.user.getPublicUsers({})).data;
+
+    const brandingData = (await $api.branding.getBrandingOptions()).data;
+    const disclaimer = brandingData.LoginDisclaimer || '';
```
ThibaultNocchi, line 67: Maybe we could `v-if`y the disclaimer block?

pages/settings/logsAndActivity.vue
```diff
@@ -146,64 +157,10 @@ export default Vue.extend({
...
   beforeMount() {
     this.setPageTitle({ title: this.$t('settingsSections.logs.name') });
-    this.getActivities();
-    this.getLogs();
   },
   methods: {
     ...mapActions('page', ['setPageTitle']),
     ...mapActions('snackbar', ['pushSnackbarMessage']),
-    async getActivities(): Promise<void> {
-      this.loadingActivityStatus.status = 'loading';
```
ThibaultNocchi, line 156: Maybe you can removed references to this and its logs version in the template and data?

ThibaultNocchi (commented):
Can you remove the mappings to the snackbar in the few components where it's not used anymore?

pages/genre/_itemId/index.vue
```diff
@@ -69,43 +73,31 @@ export default Vue.extend({
...
     return {
       title: this.$store.state.page.title
     };
   },
-  async beforeMount() {
+  beforeMount() {
     this.setAppBarOpacity({ opaqueAppBar: true });
     this.setPageTitle({
       title: this.genre.Name
     });
```
ferferga, line 100: Not related, but why not change it to ``head`` hook in this PR as well? :P

pages/login.vue
```diff
@@ -103,28 +107,6 @@ export default Vue.extend({
...
-    },
-    async getUsers(): Promise<void> {
-      try {
-        this.publicUsers = (await this.$api.user.getPublicUsers({})).data;
-      } catch (error) {
-        this.pushSnackbarMessage({
-          message: this.$t('unableGetPublicUsers'),
-          color: 'error'
-        });
-      }
```
ferferga, line 127: Why are we removing all this error catching?

pages/settings/index.vue
```diff
@@ -243,13 +250,9 @@ export default Vue.extend({
       ]
     };
   },
-  async beforeMount() {
+  beforeMount() {
     this.setAppBarOpacity({ opaqueAppBar: true });
     this.setPageTitle({ title: this.$t('settings') });
```
ferferga, line 255: Same as above

pages/login.vue
MrTimscampi: `fetch()` does it for us.

ferferga: But it doesn't display an error message. Imo we shouldn't silence errors but make them easier to understand for users.

ferferga: I guess what we need is a watcher for $fetchState.error that pushes the snackbar

Dockerfile.dev
h1dden-da3m0n: > If we're going to bother with a useless line only good for documentation, might as well use a `VOLUME` directive as well to persist state even when someone forgets to create a volume.

I know this is a comment not relevant to this PR, however the `VOLUME` instruction in Dockerfiles **is not** just a documentation instruction. IF used this it will **guarantee** that a volume is bound to the location specified and will end up creating a anonymous-volume if there was no bind for the location.
There are also other implication that this instruction will lead to, such as that the path specified in the `VOLUME` array can no longer be modified by any following instruction of the Dockerfile or any subsequent Dockerfiles using it as a dependency.

(sorry felt the need to specify this)

pages/login.vue
MrTimscampi: I was thinking of finally making that error modal I've been talking about, and generally using more visual indicators for errors than a simple snackbar (For example, replace the itemgrid on the library page with the error message), but that all felt like further PRs to improve things, rather than trying to cram it all in this one and essentially touching all the files and making a 1k lines PR.

ferferga: @MrTimscampi Okay, but don't forget about it 👍🏻

ferferga (approved):

PR #620, opened 2021-01-25
refactor: login process, clean up login strings and another responsive pass

ferferga:
* Moves login-related strings to its own subkey.
* Fixes the server selection menu appearing spontaneously on login by creating a dedicated "Add server" page
* Moves the locale switcher to the footer, in a position where it will never cover the username inputted, even in mobile. Last design, alongside the buttons, gave some problems in that part.
* Fix wrong condition in locale switcher.

https://user-images.githubusercontent.com/10274099/105722580-19baca80-5f26-11eb-85cc-ccac576147b6.mp4


layouts/fullpage.vue
```diff
@@ -3,6 +3,9 @@
     <v-main>
       <nuxt />
     </v-main>
+    <v-footer app color="rgba(0, 0, 0, 0)">
+      <locale-switcher :fab="false" />
```
ThibaultNocchi, line 7:
```suggestion
      <locale-switcher large />
```
First fab=false isn't needed for the prop to be false. Secondly, I'd enlarge the button a little bit

components/Forms/AddServerForm.vue
```diff
@@ -52,14 +54,29 @@ export default Vue.extend({
...
     async connectToServer(): Promise<void> {
       this.loading = true;
-      await this.connectServer(this.serverUrl);
+      try {
+        await this.connectServer(this.serverUrl);
+        if (this.previousServerLength === 0) {
+          this.$router.push('/login');
+        } else {
+          this.$router.push('/selectserver');
+        }
```
ThibaultNocchi, line 74:
```suggestion
        const nbr = this.$store.state.servers.serverList.length;
        await this.connectServer(this.serverUrl);
        if (nbr === 0) {
          this.$router.push('/login');
        } else {
          this.$router.push('/selectserver');
        }
```
I'd also replace the variable in the v-if by `this.$store.state.servers.serverList.length` in order to remove it

ferferga: That would make the button appear, as the transition will not have ended before the store is populated.

ferferga:
Fixed progress bar stretching server card in latest rebase as well:


https://user-images.githubusercontent.com/10274099/105751153-e8ea8d80-5f45-11eb-9681-0200425225d5.mp4

components/Forms/AddServerForm.vue
ThibaultNocchi: Fair enough, maybe add a comment there to explicit what's being done  :)

ThibaultNocchi:
@ferferga There's an issue in master which isn't fixed by your PR but I guess could be. Let's say you have the `serverUsed` value set to a server (due to being logged in or just picked a server on the selection screen). If the server is down and you load the /login page, it displays you the login form (which it shouldn't). This could be easily done by adding a similar check as [in this try-catch](https://github.com/jellyfin/jellyfin-vue/blob/83a797efddfc0442bfa094415738f5ad9a7b6cdb/components/Item/ServerCard.vue#L48) in the beforeMount method.

Also I'd just reduce a little bit the margin between server cards :)

ferferga:
@ThibaultNocchi Done! I alsoo checked and make sure that all the margins are consistent across all the pages.

Right now appears more than one error message when entering to /login and before being redirected back, that's going to be fixed by #617

pages/SelectServer.vue
```diff
@@ -2,23 +2,34 @@
...
     <v-row align="center" justify="center">
       <v-col sm="6" md="6" lg="5">
-        <h1 class="text-h4 mb-6 text-center">{{ $t('selectServer') }}</h1>
-        <div v-if="$store.state.servers.serverList">
+        <h1 class="text-h4 mb-6 text-center">
+          {{ $t('login.selectServer') }}
+        </h1>
+        <div v-if="serverList">
           <server-card
             v-for="server in $store.state.servers.serverList"
```
MrTimscampi, line 10:
```suggestion
            v-for="server in serverList"
```

pages/SelectServer.vue
```diff
@@ -28,8 +39,22 @@ export default Vue.extend({
...
+  watch: {
+    serverList(): void {
+      if (this.serverList.length === 0) {
+        this.$router.push('/addserver');
+      }
+    }
+  },
   created() {
-    this.setPageTitle({ title: this.$t('selectServer') });
+    if (!this.serverList.length) {
```
MrTimscampi, line 53: This should probably be done in `asyncData` so we check even before rendering the page. You can use `context.redirect()` to redirect from there.

pages/login.vue
```diff
@@ -67,6 +64,13 @@ export default Vue.extend({
       disclaimer: ''
     };
   },
+  async fetch() {
```
MrTimscampi, line 67: Same here, it'd make more sense to check this before the page is rendered, so in `asyncData`.

MrTimscampi (changes requested):

MrTimscampi (approved):

PR #621, opened 2021-01-25
fix: rotate 90 degrees progress circular

ferferga:
Maybe this is subjective, but I always recall seeing this kind of progress indicators starting from the bottom.

**Before**:

https://user-images.githubusercontent.com/10274099/105772891-2e698380-5f63-11eb-9b1a-be151f4d099e.mp4

**After**

https://user-images.githubusercontent.com/10274099/105772910-36c1be80-5f63-11eb-9c77-07876416db32.mp4

### DIFF
--- a/components/Item/Card.vue
+++ b/components/Item/Card.vue
@@ -16,6 +16,7 @@
           <v-progress-circular
             v-if="refreshProgress > 0"
             class="card-chip"
+            rotate="-90"
             :value="refreshProgress"
             color="white"
             size="24"


Nickbert7:
Shouldn't it start from the top like a clock?

Codex-:
So typically when it comes to the best or most common use of a user experience element like a deterministic circular progress indicator I check the material guidelines: https://material.io/components/progress-indicators#circular-progress-indicators

This indicates that we should start the indicator from _real_ 0, not 90 which is considered 0 here. 

I'd suggest adjusting this to as @Nickbert7 suggested, to start at real 0 at the top:
https://kstatic.googleusercontent.com/files/22d91d17b8958635c1cc769856551760bf3f7c8ea274e01e017d2ca7b5b3833a3cea4a4155bdf32c3d365dc9d3b5d0757d87be8f6c97854f16b4b419df6fa030

So the adjustment would be -90?

ThibaultNocchi:
Yes -90 is seems to be the rotation to start from the top

ferferga:
Changed so it starts from the top now

MrTimscampi (approved):

Camc314 (approved):

PR #626, opened 2021-01-27
fix: endless login

ferferga:


components/Forms/LoginForm.vue
```diff
@@ -102,8 +102,15 @@ export default Vue.extend({
 
       this.loading = true;
       this.setDeviceProfile();
-      await this.loginRequest(this.login);
-      this.loading = false;
+      try {
+        await this.loginRequest(this.login);
+        this.$router.replace('/');
```
MrTimscampi, line 107: Why replace instead of push?

pages/wizard.vue
```diff
@@ -124,7 +124,7 @@ export default Vue.extend({
       try {
         await this.$api.startup.completeWizard();
         // Redirect to setup complete page
-        this.$router.push('/login');
+        this.$router.replace('/login');
```
MrTimscampi, line 127: Same here?

components/Forms/LoginForm.vue
ferferga: So there's no way to go back. Doesn't make much sense if you're already logged in.

pages/wizard.vue
ferferga: Same as above

MrTimscampi (approved):

PR #631, opened 2021-01-28
refactor(home-header): use dedicated components for item titles

MrTimscampi:


components/Layout/HomeHeader/HomeHeaderEpisodeTitle.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <div>
+    <p class="text-overline text-truncate mb-2 my-2">
+      {{ $t('homeHeader.items.recentlyAdded') }}
+    </p>
+    <v-img
+      v-if="item.ParentLogoImageTag || (item.ImageTags && item.ImageTags.Logo)"
```
ferferga, line 7: What about computing this using imageHelper's mixin so it's more concise and clearer?

components/Layout/HomeHeader/HomeHeaderItems.vue
```diff
@@ -31,82 +31,33 @@
...
-                  aspect-ratio="2.58"
-                  contain
-                  data-swiper-parallax="-300"
-                  :alt="item.Name"
-                  :src="getLogo(item)"
+                <home-header-album-title
+                  v-if="item.Type === 'MusicAlbum'"
+                  class="mb-sm-n1"
+                  :item="item"
+                  :logo="getLogo(item)"
```
ferferga, line 42: Imo this should be moved to each of the components or to a dedicated mixin.

getImageUrlFromElement should help you as well

MrTimscampi: The main issue here is that it needs the related items, which are stored on the parent.

We need to refactor how this all works, since blurhashes for the parents are actually already available on BaseItemDto, so we don't need to fetch the releted items at all.

However, in the interest of keeping PRs small and self-contained, this should be done in a future one.

ferferga: getBlurhash (mixin function) should already take care of that. You can get image tag using getImageTag and pass it to that function.

Everything in the mixin is refactored in a way that doesn't forbid us to do anything (maybe the only thing we need to do is to force parent results not only when the item passed doesn't have the specified tag), what we need now is a prop or something in BlurhashImage that passes the prop value (true or false) to the mixin's function. That should be relatively easy and leave this PR somewhat more complete.

For a further refactor, what we need also imo is to take care of adding image placeholders to BlurhashImage (icons when there are no images), so we don't need to add additional logic for that in each of the components: a simple "call" to BlurhashImage solves all of that. That would lead to major reduction in core components like HomeHeader and Card.

components/Layout/HomeHeader/HomeHeaderGenericTitle.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <div>
+    <p class="text-overline text-truncate mb-2 my-2">
```
ferferga, line 3: Also, I noticed you moved this line outside the main component. Why? It's repeated in all the three, it can live in the parent.

components/Layout/HomeHeader/HomeHeaderItems.vue
MrTimscampi: The blurhash here isn't a concern (It's a logo, logos don't use blurhashes).

The issue here is that currently, passing the item itself to `getImageUrlForElement` leads to some logos not being shown, hence me keeping it like this, because I mainly want this PR to simplify the markup for the info in the home header and leave the refactoring to its own PR.

MrTimscampi: There might be some confusion here. I'm talking about blurhashes because the way things are currently setup is to get the parent item in the home header because of blurhashes. But we're using that same parent for the logo.

Passing the normal item for the logo doesn't work all the time, as it is set now. Hence the need for refactoring all of this.

ferferga: @MrTimscampi Okay, I see now.

Address the comment about the "Recently added" text and I think this is good to be merged 👍🏻

ferferga (approved):

PR #633, opened 2021-01-29
Settings rework

ferferga:
- ~~Translated Search label~~ Cherrypicked in #635
- ~~Fix off-centering of fullscreen music player definitely *(I applied the fix to the whole ``fullscreen`` route so, we can apply this same fix to all the future pages that require the default layout but need to be in fullscreen)*~~ Cherrypicked in #635

https://user-images.githubusercontent.com/10274099/106332177-8f0bff80-6286-11eb-83ed-e668cb9ef80e.mp4

- ~~Fix jumpiness of the back button in index page:~~ Cherrypicked in #635

**After**

https://user-images.githubusercontent.com/10274099/106331039-8f0b0000-6284-11eb-8a06-8ab89cf69846.mp4

**Before**

https://user-images.githubusercontent.com/10274099/106331609-849d3600-6285-11eb-93dd-3a3b3c50ff8a.mp4

- Refactored like button to better handle websocket messages.

### Settings rework

- Refactored logs & activity: searchable and with virtual scroll (an limit must be specified in the request, so it's hardcoded to 5000, but we can increase this) *This however has the disadvantage that the list sizes are fixed*
- Refactored settings pages templating
- Created placeholders for each of the settings page, which includes titles and middleware for admin pages (making it easy not to forget about it)
- Fixed alignment issues with titles in different settings sections.

**Walkthrough**

https://user-images.githubusercontent.com/10274099/106337630-f54a4f80-6291-11eb-8744-431a97067080.mp4

**Mobile walkthrough**

https://user-images.githubusercontent.com/10274099/106337784-65f16c00-6292-11eb-89f9-4b724dc120af.mp4

*What I'm aiming for in this design*

- **Not overwhelm**: One of the problems of our current web client, imo, is that settings are overwhelming and difficult to discover with the mix of dashboard/user settings and tabbed design.

- **Easier to scale through all user interfaces**: I think we tend to cram a lot of info with pages. If we design using this card approach, we will be accostumed to show stuff "better": We can keep settings concise and think more about how to present the info, instead of having "free real state" like we have in pages. **All the current existent settings could be moved to this new design without issues, which I think it's enough to question us how we're using screen space**
Things that will probably be most used should be shown first, while extra stuff should redirect to full screen pages, modals or windows. *However, imo, if we ever need this, it's because we're not doing something right and we're overcrowding the settings*.

**See my approach**

https://user-images.githubusercontent.com/10274099/106334207-7c93c500-628a-11eb-9a14-bbc562883757.mp4

- **Inspired on Android & iPadOS settings design**

- **Easier to follow where we are located thanks to the menu**: when exploring settings, it might be difficult to follow where the user is located, specially if using the back button. With navigation, you can easily have a history.

https://user-images.githubusercontent.com/10274099/106334123-48b89f80-628a-11eb-952c-87967fffc65f.mp4

- **Keep alive**: Pages keep their state cached, so forms and such will preserved between page switches.

- **Less popping**: As we use transitions for every page change, changing between them might be annoying if you're a first time user that wants to explore everything. With this approach, we only swap and animate 50% of the page while it's loading, which imo is more pleasant.


neopc10:
It's great to see that logs and activity are consolidated here. I don't know if there are server changes needed for this but I'd really like to see Clear Logs and Clear Activity buttons here that will clear the entire slate as opposed to the far less useful scheduled task functions as they exist now in jellyfin-web (and on a different page).

MrTimscampi:
Can we split this up into multiple PRs, please?

Pull requests should be relatively self-contained. This feels like at least 3 different ones smashed together:
* The settings rework
* The Swiper fixes
* The misc fixes

All this does is make reviews longer and more annoying to do, along with blocking some legitimate fixes when the change accompanying them has issues or needs changes.

dkanada:
That back button seems awkward even with your improvement, although I can agree that your change is definitely a net positive. Elements should never jump around on the page like that unless the user expects it.

MrTimscampi:
Like said yesterday on Matrix, I think the "double navigation" is awkward at best. Having both the navigation drawer and the settings navigation doesn't look and feel good.

It can be addressed by hiding the navigation drawer like this (Note, I adjusted the left part to 25% here, because 35% is way too big, imo):
image

But then there's still other issues, mainly around how the right part is organized and presented. Everything there is too big.

The *huge* logo in the above screenshot is a good first example, along with the *really large* links right below it. But other screens suffer from this as well:
image

image

image

Every control is way too large, taking up the entire width of the block (So 75% of the page here). While this looks good on mobile (And even then I have reservations I'll get into later), it looks and feels really terrible on desktop.

The idea of the current design is to be able to arrange smaller blocks on the page on desktop, while having them naturally fall into a line on mobile, making for smaller controls and a less overwhelming presentation in terms of size.

For "responsivity", there is one **major** usability issue with the way this is setup. If I do the following:
* Go to the settings on a large window
* Navigate around on a few pages
* Resize the window until you reach the mobile sizes

This leads to the navigation being completely nonsensical when going back, jumping through every sub-page I went through before finally bringing me back to the menu.

For mobile specifically, I think the card background looks weird and out of place there:
image

Basing ourselves on the material design guidelines, there should be no backdrop:
image

While good for navigation *when it helps to break up lists of links into sections*, I think it shouldn't be a thing here.

Drilling down into the subpages now.

### Account
image

This suffers from the same issue of size mentioned before. Very wide controls, very little information, awkward placement.  
There's no real reason to use an item-list everywhere, it makes everything more awkward, if anything.

The user account type is far too detached visually from the actual account, it'd be nicer like this (Along with indicating normal users, not only admins. In the future, when we have custom groups, that could be used there):
image

The "Upload image" button could be replaced by a clickable user avatar, which is a common pattern, so would be expected by the user. I also don't really like the use of icons on buttons, since we use them nowhere else. Let's be consistent here.

The password inputs are too wide. For reference, this is how the page would look like on a (simulated) ultrawide:
image

This is caused in part by your used of a CSS grid instead of the grid system, which is something that should, imo, be changed as well.

### Playback
image

The section titles here look awkward in the middle, since they make your eye go to the center and are the only thing that lives there. So scanning the page would be:
* Eyes to the left: "Playback"
* Eyes to the center: "Video"
* Eyes back on the left for individual settings
* Eyes all the way back to the right for the actual control for something

This doesn't feel very nice or polished.

Again, not a fan of the icons everywhere, since we do that nowhere else.

image

These kind of toggles should be on one line. For the crossfade specifically, the toggle seems like unnecessary complexity. You could very well have a "0s" setting in the slider below, which disabled crossfade.

It also suffers from the usual issue of everything being too wide.

### Media players
image

The warnings about unsupported features are nice. I see you already mention tying this to settings and generating the incompatibilities in a better way, so good thing there.

I'm not too sure about where these appear, and how they appear and disappear, though:
https://user-images.githubusercontent.com/19396809/106350744-927ba700-62d7-11eb-9714-1a0910585cea.mp4

It feels somewhat confusing to have it move the other controls downwards, shifting the entire page. Plus the controls "jump" between positions, making it a bit jarring.

The idea here was also to have sub-pages for the various player's settings. So any Google Cast settings would go into that page. I could see something like this:

```
| Airplay       | Toggle |
| Configure this player  |
| Google Play   | Toggle |
| Configure this player  |
```
The link to the configuration page would be disabled if the player itself is disabled.

### Subtitles

There is no page for this and it's a holdover from the current design, but maybe we should merge this into the Media Players section.

### Logs & Activity
image

This page looks a bit weird, by having cards stacked upon each other.

### Scheduled tasks
image

This suffers from the same weird look as the Logs & Activity one. 

Furthermore, the title-to-subtitle ratio is a bit too big:
image

The main page title could very likely be reduced a lot here. Also not too fan of these titles using non-vuetify font sizes, since that makes it completely decoupled from everything else in the app, both in term of potential changes in the future, user customization and actual typographic style.

### Miscellaneous

After using it for a bit, the card-layout you've used is a bit weird, compared to how your inspirations for this are. On Android settings, for example, the left part takes up the full page. It doesn't cut-off on the top and bottom.

Even other similar user interfaces do this. Here's Ubiquiti's Unifi Controller, for example:

https://user-images.githubusercontent.com/19396809/106351053-f43d1080-62d9-11eb-9dee-191533f98a14.mp4

ferferga:
@MrTimscampi Cherry-picked what was possible into #635 and thank you very much for your thorough review

### General stuff

I didn't pay much attention to sizing per-se in the new pages as I wanted to get some input from the reviews as well. The main goal was to show the distribution and get feedback for stuff like that.

As for the navigation bar sizing, I agree with you and I prefer your exact sizing in fact, gives the best balance imo. However, **went with 35% so description text is visible in the widest range of screens possible**, which isn't the case of 25%. Maybe we should consider **Removing ``text-truncate`` so text wraps?** Having description texts that are partly visible defeats the whole purpose of it imo.

### Mobile

As for the card design in mobile, I completely agree with you, thought about it before pushing. That's an easy fix, just adding a class for ``smAndDown`` that makes the background transparent.

As for the go back when resizing from desktop, I don't know how we can "solve" it. We can intercept back buttons presses when the layout changes. It's not a bug per-se, if you start in mobile layout, you will always be greeted by the menu once you hit back, the problem as you say it's when you resize to that size from desktop which, imo, **it's a really edge case**. Those situations can happen when listening to music or using pip, but I don't think nobody is going to set that layout in desktop explicitly, it's counter-intuitive.

### About

As said above, didn't pay much attention to sizing aside from the right/card layout. I'll switching the image to ``vh`` sizing.

### Typography

What about setting these using a variable, so we can customize them later through settings? I think they look nice and we can use them for all the windows and popups.

### Playback settings & players

The crossfade behaviour was straight taken from Spotify:


https://user-images.githubusercontent.com/10274099/106355463-5a389080-62f8-11eb-9c95-2aeda91df336.mp4

The whole design around doing Title/Subtitle/divider was inspired from Spotify as well. I placed subtitles alongside titles as Spotify titles are generic (the crossfade one says 'Playback', but doesn't make sense for us to make an special section for it when we're indeed in the 'Playback' settings), so I thought it was a good idea to leave descriptions and title together for consistency. In the queue sync, for instance, we don't need the left space below it for anything, but putting the description there loses all the consistency. Same approach for media players page.

As for the titles of the sections, I thoght it was a nice touch to add the icons so it could be seen at a glance what each section is about, but I can remove them and align to the left.

About the alert popping things, I have no idea how we can do it in a better way, imo a window is too intrusive for such a simple thing. We could also only show it after the toggle has been enabled (so subsequent visits to that page without changes in the toggle won't show the alert). The same applies to the loading bars in the scheduled tasks menu, I don't know why sometimes they make the cards jump a bit.

### Logs & activity

Can't really fix that if we want virtual scrollers.

### Card layout

The fact that both navigation and content cards are sized to cover the maximum space possible is to avoid situations like this:


https://user-images.githubusercontent.com/10274099/106355830-c4523500-62fa-11eb-9920-e5d4a8e303ad.mp4


As you can see **if the content is smaller than navigation, it doesn't look symmetrical**. The other way around if the content is bigger than navigation, you stop seeing the navigation menu. Ubiquiti solves this by setting a fixed position to the drawer and letting left content to be scrolled with the page, however, there are two differences:

* The way their navigation drawer is designed makes it **never to go bigger than the current screen**. Our server dashboard navigation will always be bigger than what a standard 20" 1080p screen (the vast majority of computer users) and will always overflow.

* Their layout is designed in a way that most pages doesn't require scroll at all:


https://user-images.githubusercontent.com/10274099/106355992-3a0ad080-62fc-11eb-9baf-d41312c03d4e.mp4

I haven't tried fixed/sticky positioning for the navigation drawer and I couldn''t get the right card not overflow the left without using grid. Also, maybe we can get rid of the cardbox at the left to get a design similar to UI Controller and keep content "feel" like they belong to the page and they're not "isolated" in a card. The point is that **we should keep independent control of the scrolling position of the navigation and the content**, which is not possible with flex as I demoed in the first video.

Trying sticky/fixed and removing background of the left content sounds good to you so it's worth a try?

components/Buttons/LikeButton.vue
```diff
@@ -36,12 +36,12 @@ export default Vue.extend({
     }
   },
   mounted() {
-    this.$store.subscribe((mutation) => {
+    this.$store.subscribe((mutation, state) => {
```
Camc314, line 39: Imo we shouldn't listen to the websocket messages here, we are already listening to them in the `card` component, which should update this via props when there is a change

ferferga: We can rework this on #639. The like button is not always used in card components, but also in pages and such.

components/Buttons/UserButton.vue
```diff
@@ -23,6 +23,9 @@
         :key="`bottomMenuItems-${index}`"
```
Camc314, line 23: If this file goes in a separate PR now, it can be merged

components/Layout/Settings/AddApiKeyDialog.vue
```diff
@@ -2,7 +2,7 @@
   <v-dialog
```
Camc314, line 2: I don't think this file should be in layout. It doesn't have any slots. Thoughts @ThibaultNocchi ?

pages/settings/admin/apikeys.vue
```diff
@@ -49,8 +58,13 @@ export default Vue.extend({
   data() {
     return {
       apiKeys: [] as AuthenticationInfo[],
-      addingNewKey: false,
-      newKeyAppName: ''
+      newKeyAppName: '',
+      loading: false
```
Camc314, line 62: This uses async data, so I don't think loading on the buttons is necessary.

pages/settings/admin/devices.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <settings-page page-title="settings.devices.devices">
+  <settings-page
+    page-title="settings.devices.devices"
+    help-link="https://jellyfin.org/docs/general/server/devices.html"
```
Camc314, line 4: Would it be better to use an object in the `settings-page` component with every link, and just provide a key here? If for any reason, the links needed to be updated, it would be significantly easier to have them all in one place.

pages/settings/admin/dlna.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <settings-page page-title="settingsSections.admin.dlna.name">
+    <template #content>Placeholder for DLNA settings</template>
```
Camc314, line 3: Probably easier just to leave these pages out for now. And add them when they're functional, instead of placeholders

pages/settings/admin/tasks.vue
```diff
@@ -0,0 +1,248 @@
+<template>
```
Camc314, line 1: Seperate PR please 😉

ferferga:
Uploading user images will be possible after https://github.com/jellyfin/jellyfin/issues/5197 is fixed.

Will be done in another PR.

pages/settings/admin/apikeys.vue
ferferga: Loading is shown while the request (add/remove) api is being processed, which can be seen in slow networks.

pages/settings/admin/dlna.vue
ferferga: The idea of placeholders is so we don't forget about adding the titles of it and the middleware redirection. We already had some of this stuff missing (api keys was missing the middleware and logs activity was missing the ``head`` iirc). Also, without them, clicks on the navigation drawer for the pages will keep multiple components in an active state (as no redirection will be happening), which looks really bad, specially if we're going to do a preview sooner than later.

pages/settings/admin/tasks.vue
ferferga: Fine for me but it might look akward UI-wise as it depends on the changes made to ``SettingsPage`` component.

pages/settings/admin/devices.vue
ferferga: Looks good to me. Which approach should we follow to check which page we're in?

pages/settings/admin/tasks.vue
Camc314: That's fine

pages/settings/admin/dlna.vue
Camc314: My only thoughts are since we are heading to a RC in the next few weeks, it'd be better not to have blank pages with placeholder text, and only add them when the page is implemented

pages/settings/admin/apikeys.vue
Camc314: Ah ok, we should probably specify which area is loading, so a loading state for addApiKeys, removeApiKeys...

pages/settings/admin/devices.vue
Camc314: Up to you, we could even query the router instead. 

Something like below would be best for the settings-page
```
{
  'helpLink': 'https://...jellyfin docs...',
  'translationKey': 'settings.devices.devices'
}
```

components/Layout/Settings/SettingsPage.vue
```diff
@@ -0,0 +1,74 @@
+<template>
```
Camc314, line 1: Can we update the tests for this file please

ferferga:
Labeling as blocked so all the cherrypicked PRs go before this one

locales/en-US.json
```diff
@@ -214,11 +212,30 @@
   "serverVersion": "Server version",
   "serverVersionTooLow": "Server version needs to be 10.7.0 or higher",
   "settings": {
+  "settings": {
+    "about": "About",
```
ThibaultNocchi, line 216: There's a duplicate here :p

MrTimscampi:
Closing this as it's been discussed before that it's not the direction I want to go in, and I don't think it's an improvement.

We can revisit some improvements to the settings page once we're further along in development if we need to.

PR #634, opened 2021-01-29
build(node): downgrade to node 12

crobibero:
Node 14 occasionally crashes with 
```
➜ docker logs -f jellyfin-vue
yarn run v1.22.5
$ nuxt-start
ℹ Modern bundles are detected. Modern mode (server) is enabled now.
ℹ Listening on: http://172.18.0.4:80/

<--- Last few GCs --->

[28:0x55951a5f4e20]   187782 ms: Scavenge (reduce) 4088.1 (4103.6) -> 4087.7 (4104.1) MB, 14.0 / 0.0 ms  (average mu = 0.178, current mu = 0.061) allocation failure
[28:0x55951a5f4e20]   190374 ms: Mark-sweep (reduce) 4088.5 (4101.1) -> 4087.7 (4102.3) MB, 2577.2 / 0.4 ms  (average mu = 0.106, current mu = 0.024) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
error Command failed with signal "SIGABRT".
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

### DIFF
--- a/.ci/azure-pipelines-build.yml
+++ b/.ci/azure-pipelines-build.yml
@@ -9,7 +9,7 @@ jobs:
       - task: NodeTool@0
         displayName: "Install Node"
         inputs:
-          versionSpec: "14.x"
+          versionSpec: "12.x"
 
       - task: Cache@2
         displayName: "Check Cache"
@@ -41,7 +41,7 @@ jobs:
       - task: NodeTool@0
         displayName: "Install Node"
         inputs:
-          versionSpec: "14.x"
+          versionSpec: "12.x"
 
       - task: Cache@2
         displayName: "Check Cache"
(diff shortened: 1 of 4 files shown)


ferferga:
*Saying all of this out loud so I remember to check all of this later*

I know we discussed this in Matrix but imo the ideal fix would be to find exactly which dependencies are causing issues and why, instead of relying on a version that it's already in maintenance state.

I wonder if the issue is only in the alpine image, as nobody else got around this problem. Maybe the Debian one is worth a try as well.

MrTimscampi (approved):

PR #639, opened 2021-01-31
Add server-controlled item data updates

MrTimscampi:
This adds a mixin to provide real-time updating of items based on notification from the server.

Some parts that can't use the mixin have their own implementation.

https://user-images.githubusercontent.com/19396809/106389561-2c307a80-63e4-11eb-87df-22a4542015ef.mp4

### DIFF
--- a/store/plugins/websocket.ts
+++ b/store/plugins/websocket.ts
@@ -1,3 +1,4 @@
+import { BaseItemDto } from '@jellyfin/client-axios';
 import { Plugin } from 'vuex';
 import { AppState } from '..';
 
@@ -27,7 +28,9 @@ export const websocketPlugin: Plugin<AppState> = (store) => {
     store.state.socket.instance?.send(JSON.stringify(msg));
   }
 
-  store.subscribe((mutation, state) => {
+  store.subscribe(async (mutation, state) => {
+    // Messages
+    // Keep the websocket alive
     if (
       mutation.type === 'SOCKET_ONMESSAGE' &&
       state.socket.message.MessageType === 'ForceKeepAlive'
@@ -36,7 +39,31 @@ export const websocketPlugin: Plugin<AppState> = (store) => {
       keepAliveInterval = window.setInterval(() => {
         sendWebsocketMessage('KeepAlive');
       }, state.socket.message.Data * 1000 * 0.5);
-    } else if (mutation.type === 'SOCKET_ONCLOSE') {
+    }
+    // Update items when changed
+    else if (
+      mutation?.type === 'SOCKET_ONMESSAGE' &&
+      state.socket.message.MessageType === 'LibraryChanged'
+    ) {
+      // @ts-expect-error -- The Data property doesn't describe its content
+      const itemsToUpdate: string[] = state.socket.message.Data.ItemsUpdated.filter(
+        (itemId: string) => {
+          return itemId in store.state.items.allIds;
+        }
+      );
+
+      const updatedItems: BaseItemDto[] =
+        (
+          await store.$api.items.getItems({
+            userId: store.$auth.user?.Id,
+            ids: itemsToUpdate
+          })
+        ).data.Items || [];
+
+      store.dispatch('items/addItems', { items: updatedItems });
+    }
+    // Socket disconnection
+    else if (mutation.type === 'SOCKET_ONCLOSE') {
       if (keepAliveInterval) {
         clearInterval(keepAliveInterval);
       }


ferferga:
Some comments:

* Isn't easier to create a new variable in data and in ``created`` set the prop to that variable and change that variable everytime there's a new item update instead of all the event mess? So the ``prop`` is just "set and forget" and all the references in the template are for that variable.

* Why create a mixin for this and not do it directly inside the card?

Additional thoughts, not for this PR are in comments

mixins/itemUpdate.ts
```diff
@@ -0,0 +1,61 @@
...
+      type: Boolean
+    }
+  },
+  created() {
+    if (!this.manualUpdate) {
+      this.$store.subscribe(async (mutation, state) => {
+        if (
+          this.item.Id &&
+          mutation?.type === 'SOCKET_ONMESSAGE' &&
+          state.socket.message.MessageType === 'LibraryChanged' &&
```
ferferga, line 44: We should create an enum and proper interfaces for mutation types and socket message types. We are relying more and more on socket stuff and this is getting dirty.

MrTimscampi:
> Isn't easier to create a new variable in data and in created set the prop to that variable and change that variable everytime there's a new item update instead of all the event mess? So the prop is just "set and forget" and all the references in the template are for that variable.

What if the parent component uses that data elsewhere? We'd have a difference in the data. Handling data at the deepest state and moving it back up makes the most sense, given how the API works right now.

In some places we'll have to handle it differently (Item details, places with virtual scrollers) but it makes sense to have the deepest item be the default (At least until we figure out a good way to use Vuex for everything)

>     * Why create a mixin for this and not do it directly inside the card?

So that it can be reused. Anything we want to pass an item prop to, we can use this instead and have updates handled for us.

ferferga:
@MrTimscampi This is how I would do it: 

* Add to the mixin ``data`` and ``watch`` properties: In data, set a new variable called ``it``or ``upditem`` (not sure what's the best name for it)
* Set an **immediate and deep** watcher for the item prop whose handler does this: ``this.upditem = this.item``
* Update in the store mutation's callback ``updItem`` instead of emitting anything.
* Remove the item prop, store subscriptions, etc... in each of the components that rely on a single ``BaseItemDto`` and replace references to item with references to ``updItem``.
* Add the mixin to each of those components so ``data``, ``watch`` and ``created`` are "injected".

For handling pages, I would do a check of ``this.$route.fullPath.includes(item.Id)`` in the mixin's and call the refresh there, so there's no need for additional code anywhere else.

Add full documentation about all of this to the mixing so it's clear where everything's coming from for newcomers. My approach is not extraordinarily easier either, but imo it makes it more difficult to miss updates across the chain, which might be problematic with the event way of things (aside from code duplication).

Let me know what do you think about this, with this I think we will propagate updates properly across all the components.

MrTimscampi:
> This is how I would do all of this instead

You're making things a lot more complex and duplicating things all over the place, though (Unless I misunderstand what you want to do)

We can't really do one store for items, since we have items in plenty of things (Home sections, libraries, individual items, playlists, as children of other items, etc).

So if we make one Vuex store for all the items, we'd have to do plenty of processing on it whenever we need something (Essentially replicating what the server does).

This way keeps things simple, allows us to override it if need be, and is flexible enough to be used everywhere we need it.

ferferga:
@MrTimscampi there's no need of a store, it's all inside the same mixin you created

MrTimscampi:
I feel like duplicating stuff and keeping two different versions of data is a bad idea.

Here, you'd keep using `item` everywhere, like we do currently, which makes a lot more sense and doesn't duplicate potentially different versions of things all over the place.

It's easier and more logical to handle updates the way it's done in this branch, imo.

ferferga:
@MrTimscampi I worded it wrong maybe. My point is that the fact that storing parts of the components in a mixin is already confusing (I wondered, for instance, why item was removed from card the first time I looked into this PR). There are no comments about that in your commits, only in the callbacks of each of the parent components (so we will need commenting either way). I wasn't referring to comments in my system alone, but also to yours, so **we will need comments either way**.

Additionally to all of this, I want to link this from Vue docs: https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-the-Parent-Component-Instance

Your system is not exactly the one they put as an example there, but imo is the same anti-pattern we should avoid: **data should flow as much as possible in one direction, not in a two-way one**. The whole fact that we will need to rely on events for this instead of Vue's "internals" is also another reason why I think this approach is a bad idea.

There also will never be duplication: there will be one variable that's mutable and one that isn't in each instance that relies on BaseItemDto, both of which will be in perfect sync as long as the parent also makes use of the mixin in the structure I proposed.

I can make a branch for demonstration purposes if wanted.

ThibaultNocchi:
This component could be handled in a mixin (global or not) instead, and still use events or something.

MrTimscampi:
Blocked, as refactoring depends on #787 and #764

MrTimscampi:
Reworked this to use the `items` store and the websocket plugin.

The plugin now listens to the `LibraryChanged` message from the server, filters out items we don't have (so we don't waste requests on things we're not showing anywhere anyway), then fetch all of them and update the store.

store/index.ts
```diff
@@ -78,6 +80,7 @@ export const actions: ActionTree<RootState, RootState> = {
     promises.push(dispatch('backdrop/clearAllBackdrop', { root: true }));
     promises.push(dispatch('deviceProfile/clearDeviceProfile', { root: true }));
     promises.push(dispatch('displayPreferences/resetState', { root: true }));
+    promises.push(dispatch('items/clearState', { root: true }));
```
ferferga, line 83: Maybe we can except this store, so if an user logs out the data it's loaded instantly for the next user (if the user it's in the same server)?

Although I guess it's better to keep it for consistency.

MrTimscampi: That's from the items store PR :)

Main issue with keeping items between sessions, though, is that they might have UserData stuck to them, which isn't good.

store/plugins/websocket.ts
```diff
@@ -36,7 +39,31 @@ export const websocketPlugin: Plugin<AppState> = (store) => {
       keepAliveInterval = window.setInterval(() => {
         sendWebsocketMessage('KeepAlive');
       }, state.socket.message.Data * 1000 * 0.5);
-    } else if (mutation.type === 'SOCKET_ONCLOSE') {
+    }
+    // Update items when changed
+    else if (
+      mutation?.type === 'SOCKET_ONMESSAGE' &&
+      state.socket.message.MessageType === 'LibraryChanged'
```
ferferga, line 46: There isn't an enum we can use with all the websocket message types?

MrTimscampi: There is one in the server, but it's not exposed to the API at the moment.

ferferga: We should raise an issue for this in server 👍🏻.

ferferga (approved):

ThibaultNocchi (approved):

PR #640, opened 2021-01-31
Implemented stream selection on item detail page and VideoPlayer

Sergeeeek:
Changes:
- Add mediaStreamPreferences Vuex store to store stream selection for each item in a session
- Populate url parameters for `VideoStreamIndex`, `AudioStreamIndex` and `SubtitleStreamIndex` when building video source url

Partially addresses #557

### DIFF
--- a/components/Item/TrackSelector.vue
+++ b/components/Item/TrackSelector.vue
@@ -1,6 +1,5 @@
 <template>
   <v-select
-    v-model="trackIndex"
     outlined
     filled
     flat
@@ -8,10 +7,12 @@
     single-line
     hide-details
     class="text-truncate"
+    :value="value === undefined ? defaultIndex : value"
     :items="selectItems"
     :placeholder="placeholder"
     :disabled="disabled"
     :clearable="clearable"
+    @input="onChange"
   >
     <template slot="selection" slot-scope="{ item: i }">
       {{ getTrackSelection(i.text) }}
@@ -50,6 +51,10 @@ import {
 
 export default Vue.extend({
   props: {
+    value: {
+      type: Number,
+      default: undefined
+    },
     /**
      * Media item
      */
@@ -74,9 +79,6 @@ export default Vue.extend({
         return ['Audio', 'Subtitle', 'Video'].includes(value);
       }
     }
-  },
-  data() {
-    return { trackIndex: undefined as number | undefined };
   },
   computed: {
     mediaSourceItem: {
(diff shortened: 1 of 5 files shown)


components/Item/TrackSelector.vue
```diff
@@ -251,6 +227,10 @@ export default Vue.extend({
         default:
           return 'mdi-surround-sound';
       }
+    },
+    onChange(event: any): void {
+      console.log(event);
```
MrTimscampi, line 232:
```suggestion
```

MrTimscampi (changes requested):
Thanks for the contribution :)

I'm unfortunately not too fond of the implementation of this. I don't think using another store is necessary for this.

What should be done is extending the `play` action in the playback manager to be able to take optional track indexed, which would then be set directly when starting playback.

If they're not passed to the play action, the playback manager automatically sets the default ones. Then simply use the indexes in the playbackManager store when generating the playback URL.

When switching to the next track, check if the indexes are still valid, if not switch back to the default ones.

I'm curious to know what the rest of the Vue team thinks, though.

@ferferga @Camc314 @ThibaultNocchi @dkanada

Camc314:
> What should be done is extending the `play` action in the playback manager to be able to take optional track indexes, which would then be set directly when starting playback.

This is probably an easier method.

dkanada:
@MrTimscampi just out of curiosity, does the play method already take time as an argument?

ferferga:
@dkanada Yes

I agree with what others said here: it seems overkill to me to create a single store just for stream selection. In some cases, decoupling stores are a good solution towards maintainability but I don't think this is the case with playback-related stuff: playback stuff makes sense to be as "single-managed" as possible.

I'm not sure if you used the multiple stores approach because dictionaries, as multidimensional objects, have some issues with Vue's reactivity. Imo, **selecting track indexes for each of the items in the queue is overkill**, normally nobody does that job manually and either expect:

• That the default tracks of the next item are used
• That the language of the currently playing show/movie track is chosen for the next one.

We could live perfectly with just integer variables that refer to the currently playing track indexes instead of a dictionary, so if that adds complexity for you, I would go that way. However, if you still want to keep the dictionary way **(which can also be useful, as we could add track selectors to the queue view where users can change those preferences individually)**, move all the logic into the playbackManager store.

Let us know if you need any help with anything 👍🏻.

Sergeeeek:
Thanks for your feedback everyone. 

I had created a separate store under an assumption that stream indexes for different media sources mean different things, e.g. audio stream 0 in one movie could be English, while in another movie it could be Spanish or something else. 

From your comments I see that maybe it was incorrect. Does Jellyfin server reorder streams based on user preferences? 

Anyway, I will remove the separate store and pass stream indexes into the play method as you suggested.

ThibaultNocchi:
No, your first assumption was correct. To my knowledge, JF just parses the tracks in the media in the order they are. But what happens usually though is that in a TV show, usually they have the same tracks between each episodes. That's why keeping the same tracks index usually works. :)

MrTimscampi:
> That's why keeping the same tracks index usually works

That said, there should still be a check if the track is defined and a fallback to the default one :) If the track ends up being different, there isn't much we can do realistically with the current API, but it's easy enough for the user to fix during playback.

ThibaultNocchi:
@Sergeeeek Hey, are you planning to improve the PR? If yes then no problem, take your time :)

Sergeeeek:
Yes, I am. Just been a little busy lately :)

ThibaultNocchi:
No problem then, do what you want :)

PR #644, opened 2021-02-01
Fix FMP4-HLS playback on Edge and Safari

nyanmisaka:
**Changes**
- Fix browser detection for Edge chromium
- Fix fmp4-hls playback on Edge and Safari
- Reduce OSD timeout from 5 seconds to 3 (https://github.com/jellyfin/jellyfin-web/pull/903)

### DIFF
--- a/plugins/browserDetection.ts
+++ b/plugins/browserDetection.ts
@@ -74,7 +74,17 @@ class BrowserDetector {
    * @memberof BrowserDetector
    */
   isEdge(): boolean {
-    return this.userAgentContains('Edge/');
+    return this.userAgentContains('Edg/') || this.userAgentContains('Edge/');
+  }
+
+  /**
+   * Check if the current platform is Chromium based.
+   *
+   * @returns {boolean} Determines if browser is Chromium based
+   * @memberof BrowserDetector
+   */
+  isChromiumBased(): boolean {
+    return this.userAgentContains('Chrome');
   }
 
   /**
--- a/utils/playbackProfiles/helpers/fmp4AudioFormats.ts
+++ b/utils/playbackProfiles/helpers/fmp4AudioFormats.ts
@@ -6,6 +6,7 @@ import {
   hasEac3Support,
   hasMp3AudioSupport
 } from './mp4AudioFormats';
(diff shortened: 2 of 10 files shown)


plugins/browserDetection.ts
```diff
@@ -74,7 +74,17 @@ class BrowserDetector {
    * @memberof BrowserDetector
    */
   isEdge(): boolean {
-    return this.userAgentContains('Edge/');
+    return this.userAgentContains('Edg/') || this.userAgentContains('Edge/');
```
ferferga, line 77: Good catch!

utils/playbackProfiles/helpers/fmp4AudioFormats.ts
```diff
@@ -35,7 +36,9 @@ export function getSupportedFmp4AudioCodecs(
   }
 
   if (getSupportedAudioCodecs('flac')) {
-    codecs.push('flac');
+    if (!browserDetector.isEdge()) {
+      codecs.push('flac');
+    }
```
ferferga, line 41: Why? Version 16 should support it: https://caniuse.com/flac. Maybe we should call ``canPlay`` instead of hardcoding this?

utils/playbackProfiles/transcodingProfile.ts
```diff
@@ -12,6 +12,7 @@ import {
   hasVp8Support
 } from './helpers/mp4VideoFormats';
 import { canPlayNativeHls, hasMkvSupport } from './helpers/transcodingFormats';
+import { canPlayHlsWithMSE } from './helpers/transcodingFormats';
```
ferferga, line 15: Merge this import please

ferferga (changes requested):

utils/playbackProfiles/helpers/fmp4AudioFormats.ts
nyanmisaka: It's indeed supported in MP4 and MKV container.
But I tested in Edge 88 and FLAC is not supported in shaka-player with FMP4-HLS.

ferferga (approved):

PR #645, opened 2021-02-01
fix: spacing in artist pages, translation and cards

ferferga:
* Better spacing between albums
* Remove the overlay from card
* Translatable tabs

**Spotify**
image

**Vue before**
image

**Vue after**
image

### DIFF
--- a/components/Item/Card/Card.vue
+++ b/components/Item/Card/Card.vue
@@ -70,12 +70,12 @@
           />
         </div>
         <div
-          v-if="overlay && !$browser.isMobile()"
+          v-if="!noOverlay && !$browser.isMobile()"
           class="card-overlay d-flex justify-center align-center"
         >
           <play-button fab :item="item" />
           <div
-            v-if="overlay"
+            v-if="!noOverlay"
             class="card-lower-buttons d-flex justify-center align-center"
           >
             <like-button v-if="canPlay(item)" :item="item" dark />
@@ -120,10 +120,10 @@ export default Vue.extend({
         return false;
       }
     },
-    overlay: {
+    noOverlay: {
       type: Boolean,
       default: (): boolean => {
-        return true;
+        return false;
       }
     },
     noText: {
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -1,6 +1,5 @@
 {
   "3DFormat": "3D format",
-  "about": "About",
   "actor": "Actor",
   "actors": "Actors",
   "addNewPerson": "Add a new person",
@@ -88,6 +87,10 @@
   },
   "incorrectUsernameOrPassword": "Incorrect username or password",
   "item": {
+    "artist": {
+      "discography": "Discography",
+      "information": "Information"
+    },
     "cast": "Cast",
     "composer": "Composer",
     "crew": "Crew",
--- a/pages/artist/_itemId/index.vue
+++ b/pages/artist/_itemId/index.vue
@@ -12,44 +12,42 @@
           />
         </v-col>
         <v-col cols="12" sm="7">
-          <div class="ml-sm-4 d-flex flex-column">
-            <div
-              class="text-subtitle-1 text--secondary font-weight-medium text-capitalize"
-            >
-              {{ $t('artist') }}
+          <v-row justify="space-between">
+            <div class="ml-sm-4 d-flex flex-column">
+              <div
+                class="text-subtitle-1 text--secondary font-weight-medium text-capitalize"
+              >
+                {{ $t('artist') }}
+              </div>
+              <h1 class="text-h4 text-md-h2 font-weight-light">
+                {{ item.Name }}
+              </h1>
             </div>
-            <h1 class="text-h4 text-md-h2 font-weight-light">
-              {{ item.Name }}
-            </h1>
-          </div>
+            <div class="d-flex align-center">
+              <item-menu :item="item" />
+              <like-button :item="item" />
+            </div>
+          </v-row>
         </v-col>
       </v-row>
       <v-row>
         <v-col>
           <v-tabs v-model="activeTab" background-color="transparent">
-            <v-tab :key="0">{{ $t('overview') }}</v-tab>
-            <v-tab :key="1">{{ $t('about') }}</v-tab>
+            <v-tab :key="0">{{ $t('item.artist.discography') }}</v-tab>
+            <v-tab :key="1">{{ $t('item.artist.information') }}</v-tab>
           </v-tabs>
           <v-tabs-items v-model="activeTab" class="transparent">
             <v-tab-item :key="0">
               <v-row no-gutters>
                 <v-col cols="12" class="ma-3">
-                  <v-row>
-                    <h2 class="text-h6 my-3">
-                      <span>{{ $t('albums') }}</span>
-                    </h2>
-                  </v-row>
                   <v-row v-for="appearance in appearances" :key="appearance.Id">
                     <v-col cols="12">
                       <div class="d-flex flex-column">
                         <v-row>
-                          <div
-                            style="width: 125px"
-                            class="ma-2 d-none d-md-block"
-                          >
+                          <v-col lg="2" sm="1">
                             <card :item="appearance" no-text no-margin />
-                          </div>
-                          <div class="py-2">
+                          </v-col>
+                          <v-col class="py-2">
                             <div
                               class="text-subtitle-1 text--secondary font-weight-medium"
                             >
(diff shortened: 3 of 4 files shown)


dkanada:
I can count six different left margins in a single photo. Even the previous state had five, although the bottom padding on the title is a welcome change.

MrTimscampi:
The lack of alignment between the buttons and the cover is annoying :/

I could see the card only having the Play button, but being larger, then have the two small icons aligned to the bottom of the card.... but then the card is big enough to have the buttons on it directly.

Either way, the card needs to be bigger :yum:

ferferga:
Page still needs a huge responsive pass, but after your comments I also tried to keep this somewhat responsive:

**Desktop**

image

**Mobile**
image

MrTimscampi:
> Page still needs a huge responsive pass

We should probably merge #616 before this, then, so we can adjust it properly with the few changes I made in there (I adjusted album title size based on size, among other things)

ferferga:
Final design:

**Mobile** *(As in Spotify, in mobile design the tracklist isn't shown. One can see the tracklist by clicking on an specific album)*

image


**Desktop**
image

Camc314:
Wouldn't it be better (like spotify) to have a grid of album covers, at the moment they look a bit too big imo.

neopc10:
Would also like to have a means of sorting these albums by year, or better yet, have an ascending sort by year as a default!

ferferga:
@Camc314 This is what I mean in Spotify. You mean like this but with smaller covers right?



@neopc10 I don't think this is the place to do so. You can already do that in library view. You will usually want to check albums in a recent > oldest fashion

Camc314:
Actually, It looks like Spotify doesn't do that.

What I was thinking of was using the item Grid on mobile devices, that way we could fit 2/3 widthwise

ferferga:
@Camc314 Can you post an screenshot of what you mean please? 

For reference, this is the latest version of Spotify in Android, I don't know if in iOS it's different as I don't use any iOS device, but I'm aware there are some subtle differences in the way the layout is organised.

Camc314:
Spotify doesn't actually do what I thought for artist pages. I though it had an area like this

image

.gitignore
```diff
@@ -88,3 +88,6 @@ sw.*
 
 # Vim swap files
 *.swp
+
+# Husky config files
+.husky
```
Camc314, line 93: This shouldn't be here

components/Item/Card/Card.vue
```diff
@@ -120,10 +120,10 @@ export default Vue.extend({
         return false;
       }
     },
-    overlay: {
+    noOverlay: {
```
Camc314, line 123: I would prefer we leave it as `overlay` and set the default to false

ferferga:
@Camc314 Oh I see what you mean now. That one isn't bad either. We already have the same layout in the library browsing pages and this makes me push towards the "row" design so there's a bit of differentiation. I don't mind either though, both are great ways to present the content tbh.

Thoughts @ThibaultNocchi ?

.gitignore
ferferga: Husky generated a husky.sh file in that folder, do we really want that to be committed?

pages/artist/_itemId/index.vue
```diff
@@ -12,44 +12,42 @@
...
+          </v-row>
         </v-col>
       </v-row>
       <v-row>
         <v-col>
           <v-tabs v-model="activeTab" background-color="transparent">
-            <v-tab :key="0">{{ $t('overview') }}</v-tab>
-            <v-tab :key="1">{{ $t('about') }}</v-tab>
+            <v-tab :key="0">{{ $t('item.artist.discography') }}</v-tab>
+            <v-tab :key="1">{{ $t('item.artist.information') }}</v-tab>
```
Camc314, line 37: Do we need to remove `overview` and `about` or are they used elsewhere

Camc314 (commented):
Few comments, rest looks fine to me.

.gitignore
ThibaultNocchi: Just rebase from master :)

components/Item/Card/Card.vue
ferferga: We're switching most of the props (see https://github.com/jellyfin/jellyfin-vue/pull/592#issuecomment-762050298)

neopc10:
> @neopc10 I don't think this is the place to do so. You can already do that in library view. You will usually want to check albums in a recent > oldest fashion

That's cool with a small library but isn't exactly practical with a large one.

This forces a ton of scrolling to see any relevant results for artists you have more than one album for. For example, I have something like 50 Elvis albums. The most recent is a 2019 compilation of his 1967 concerts, which has 8 discs. I'm looking for a 1950 album. That means I'm scrolling for quite some time before I finally get there. 

With most other streamers, the albums are presented in a horizontal scroller with the tracks obviously not expanded, so there's far less work involved.

Since the excessive vertical scrolling is forced here, having the ability to sort would at least allow you to choose from wihch end you want to start scrolling.

.gitignore
Camc314: The shell file shouldn't be committed, but there is a .gitignore inside husky that should tell git to ignore it

pages/artist/_itemId/index.vue
ferferga: Only ``about``, ``overview`` it's used. I need to clean this up furthermore so all the item pages and elements are in their own subkey, but that's for a further PR.

ThibaultNocchi:
@ferferga Yeah I like Cam's idea, having a grid of covers with only their name and year could be great. We don't have to duplicate the track listing as the album is only a click away, and we can still add a whole album to the queue in a click.

I also agree with neopc's idea about being able to sort them on the artist page, but I'd leave that as an idea for later and just get this first view of the artist page.

components/Item/Card/Card.vue
Camc314: I don't quite understand, The comment, on 592 is what I'm saying?

Camc314:
Shall we leave it as is for now, and merge, and then refactor over the next few weeks to use the grid?

pages/artist/_itemId/index.vue
Camc314: I've checked, about can be removed. Overview is used elsewhere

Perseusdehond:
I like both the grid view and a list view. Maybe an option in settings in the future? But I like it either way

.gitignore
ferferga: For some reason it committed to me. Now it doesn't, so I'm removing this. Strange!

Camc314: Ah no worries

components/Item/Card/Card.vue
ferferga: MrTimscampi says that, if a property is not specified, it's false by default. Before, we were defaulting overlay to ``true`` and, for disabling it, we needed to pass it like ``:overlay="false"``, which is counter-intuitive. 

Other props in card also follow the approach this prop will follow now: they are like ``noX`` and setting them means they're false, no need to do the ``:overlay="false"`` thing anymore, just ``no-overlay`` when using the component.

locales/en-US.json
```diff
@@ -1,6 +1,5 @@
 {
   "3DFormat": "3D format",
-  "about": "About",
```
ThibaultNocchi, line 3: This shouldn't be removed

components/Item/Card/Card.vue
ThibaultNocchi: I think Cam is just saying to keep the property as `overlay` even though it is at false by default (which I agree should be)

Camc314: Ah I didn't notice that all the others were `noX`, imho, like vuetify does, we should prefer the other method (instead of `noOverlay` have `overlay`). Since we do this elsewhere we can leave for now. Thoughts @ThibaultNocchi

ThibaultNocchi (changes requested):

components/Item/Card/Card.vue
ThibaultNocchi: I think Cam is just saying to keep the property as `overlay` even though it is at false by default (which I agree should be)

Camc314 (approved):

components/Item/Card/Card.vue
Camc314: So we will leave as is for now, and refactor later. @ThibaultNocchi one last approval before merge?

ThibaultNocchi:
LGTM, but we'll really need to do the responsive pass on this page :p

ThibaultNocchi (approved):

PR #672, opened 2021-02-09
refactor(user-card): fix user-image and refactor user-card

MrTimscampi:
Fixes UserImage essentially being broken and fairly useless, and refactors UserCard.

UserImage relied *heavily* on `$auth.user`, making it completely useless in a lot of situations (And breaking the client completely if a public user had an avatar set).

The structure of UserCard was also fairly nonsensical, having a `v-avatar` with a `v-image` in it, whose placeholder was another `v-avatar` with a `v-icon` in it...  
This cleans up the structure and reworks the component to look like this:
image

One remaining issue is that, since Vuetify is limited, there is no way to easily get these to have button-like hover effects while also being tied to Vuetify's styles without essentially rewriting a custom `v-btn`.

### DIFF
--- a/components/Layout/Images/UserImage.vue
+++ b/components/Layout/Images/UserImage.vue
@@ -1,27 +1,27 @@
 <template>
-  <v-avatar v-if="userImage">
-    <v-img :src="userImage" :alt="$auth.user.Name" class="userImage">
+  <v-avatar color="primary darken-3" :size="size">
+    <v-img :src="userImage" :alt="user.Name" class="userImage">
       <template #placeholder>
-        <v-avatar color="primary">
-          <v-icon dark>mdi-account</v-icon>
-        </v-avatar>
+        <v-icon :size="size - 32" dark>mdi-account</v-icon>
       </template>
     </v-img>
-  </v-avatar>
-  <v-avatar v-else color="primary">
-    <v-icon dark>mdi-account</v-icon>
   </v-avatar>
 </template>
 
 <script lang="ts">
+import { UserDto } from '@jellyfin/client-axios';
 import Vue from 'vue';
 
 export default Vue.extend({
   props: {
-    id: {
-      type: String,
+    user: {
+      type: Object as () => UserDto,
+      required: true
+    },
+    size: {
+      type: Number,
       required: false,
-      default: undefined
+      default: 64
     },
     quality: {
       type: Number,
@@ -32,14 +32,8 @@ export default Vue.extend({
   computed: {
     userImage: {
       get(): string | undefined {
-        if (
-          !this.id &&
-          this.$auth.user?.Id &&
-          this.$auth.user?.PrimaryImageTag
-        ) {
-          return `${this.$axios.defaults.baseURL}/Users/${this.$auth.user.Id}/Images/Primary/?tag=${this.$auth.user.PrimaryImageTag}&quality=${this.quality}`;
-        } else if (this.id) {
-          return `${this.$axios.defaults.baseURL}/Users/${this.id}/Images/Primary/?tag=${this.$auth.user.PrimaryImageTag}&quality=${this.quality}`;
+        if (this.user?.Id && this.user?.PrimaryImageTag) {
+          return `${this.$axios.defaults.baseURL}/Users/${this.user.Id}/Images/Primary/?tag=${this.user.PrimaryImageTag}&quality=${this.quality}`;
         } else {
           return undefined;
         }
@@ -48,9 +42,3 @@ export default Vue.extend({
   }
 });
 </script>
-
-<style lang="scss" scoped>
-.userImage {
-  image-rendering: crisp-edges;
-}
-</style>
--- a/components/Users/UserCard.vue
+++ b/components/Users/UserCard.vue
@@ -1,10 +1,11 @@
 <template>
-  <div
-    class="user-card mx-auto d-flex flex-column"
-    @click="$emit('connect', user)"
-  >
-    <user-image :size="128" :user="user" />
-    <span class="text-subtitle-1 text-center mt-2">{{ user.Name }}</span>
+  <div class="ma-2 d-flex flex-column pointer" @click="$emit('connect', user)">
+    <v-btn plain ripple :height="128" :width="128" class="rounded">
+      <user-image :size="128" :user="user" />
+    </v-btn>
+    <a class="text-subtitle-1 text-center mt-2 link">
+      {{ user.Name }}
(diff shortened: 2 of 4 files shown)


components/Layout/Images/UserImage.vue
```diff
@@ -48,9 +42,3 @@ export default Vue.extend({
   }
 });
 </script>
-
```
ferferga, line 51: Why this was removed?

MrTimscampi: Per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering):

> Suitable algorithms include nearest-neighbor and other non-smoothing scaling algorithms such as 2×SaI and hqx-family algorithms. This value is intended for pixel-art images, such as in browser games.

None of these algorithms are suited for reducing image sizes properly, and they are targeted more towards pixel art than towards proper images, which user images would generally be.

They would cause pixelization, which is not acceptable, in my opinion.

components/Users/UserCard.vue
```diff
@@ -1,29 +1,11 @@
...
-      </v-btn>
-    </v-card-actions>
-  </v-card>
+  <div
+    class="user-card mx-auto d-flex flex-column"
+    @click="$emit('connect', user)"
+  >
+    <user-image :size="128" :user="user" class="pointer" />
+    <span class="text-subtitle-1 text-center mt-2">{{ user.Name }}</span>
+  </div>
```
Camc314, line 8: Why not use a `<button>`?

components/Layout/Images/UserImage.vue
ferferga: What about swithcing to ``high-quality```?

components/Users/UserCard.vue
ThibaultNocchi: I can try that this afternoon

components/Layout/Images/UserImage.vue
ThibaultNocchi: Just a question, why is it needed?

ferferga: @ThibaultNocchi If I recall correctly I added it because Firefox image scalation added more pixelation than Chrome's.

components/Users/UserCard.vue
ThibaultNocchi: `<button>` works, but I can get a nice effect with the `plain` attribute of `v-btn`, also I'd need to rework it haha

https://user-images.githubusercontent.com/1619359/107372874-30485080-6ae6-11eb-904c-1c5ad8212fe4.mp4

components/Layout/Images/UserImage.vue
ThibaultNocchi: Here are FF (left) and Chromium (right) both zoomed in at 200%, so I'm not sure it's still needed :p
print_screen_2021-02-09-14-57-17

components/Users/UserCard.vue
Camc314: Looks good so far!

ThibaultNocchi: The video or the current state?

ferferga (changes requested):
image

Text is misaligned and there's no hover effect on the button.

ThibaultNocchi:
@ferferga Ok the misalignment came from the fact I used the `<button>` instead of the div to get at least a pointer hand, I reverted back to that. For the styling, as MrTim said in the OP we don't really know how to do it, feel free to try :)

components/Layout/Images/UserImage.vue
```diff
@@ -32,14 +32,8 @@ export default Vue.extend({
...
-          this.$auth.user?.Id &&
-          this.$auth.user?.PrimaryImageTag
-        ) {
-          return `${this.$axios.defaults.baseURL}/Users/${this.$auth.user.Id}/Images/Primary/?tag=${this.$auth.user.PrimaryImageTag}&quality=${this.quality}`;
-        } else if (this.id) {
-          return `${this.$axios.defaults.baseURL}/Users/${this.id}/Images/Primary/?tag=${this.$auth.user.PrimaryImageTag}&quality=${this.quality}`;
+        if (this.user?.Id && this.user?.PrimaryImageTag) {
+          return `${this.$axios.defaults.baseURL}/Users/${this.user.Id}/Images/Primary/?tag=${this.user.PrimaryImageTag}&quality=${this.quality}`;
         } else {
           return undefined;
```
Codex-, line 38: nit: this is probably extraneous, void functions return undefined

nit: since the first if ends with a return, the else doesn't need to be there, there is nothing further to be executed after the if block

components/Layout/Images/UserImage.vue
```diff
@@ -1,27 +1,27 @@
...
 <script lang="ts">
+import { UserDto } from '@jellyfin/client-axios';
 import Vue from 'vue';
 
 export default Vue.extend({
   props: {
-    id: {
-      type: String,
+    user: {
+      type: Object as () => UserDto,
```
Codex-, line 18: does this have to cast object as a function? Can't just be the function type I'm guessing? (genuinely curious lol)

ferferga: No, due to the bad TypeScript support that Vue 2 has. I suppose we won't need this for Vue 3 but, so far and personally, I have no idea as I haven't played with Vue 3 yet.

components/Layout/Images/UserImage.vue
ThibaultNocchi: The linter doesn't accept it though :p I could remove the else and just put the `return undefined` below the if, but it'll have to be there

ferferga: @Codex- I think you confused this with a function, this is a property 😁.

ferferga:
Converted user card to a button:


https://user-images.githubusercontent.com/10274099/107777354-ff188c00-6d42-11eb-89d5-b76773bc1e01.mp4

*Mouse doesn't appear in the recording for some awkward reason but you can imagine it :P*

ferferga (approved):

components/Layout/Images/UserImage.vue
ThibaultNocchi: Though `get()` is a function, isn't it? :p

ferferga: It's a getter: https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Funciones/get

ThibaultNocchi (approved):

PR #723, opened 2021-02-12
chore: new locales

ThibaultNocchi:


ferferga:
What are you having problems with specifically?

ThibaultNocchi:
I'm having troubles making some locales work with the date parser. For instance, I can't make the `es419` locale work with the date localizer
print_screen_2021-02-12-15-21-22

ferferga:
@ThibaultNocchi The problem is that the locales you added doesn't exist in datefns. Here you have a list of the available locales:

https://github.com/date-fns/date-fns/tree/master/src/locale

*With the dashes removed, as you can see in the enUS defined in our nuxt config*

i18n itself doesn't care about the codes you assign to the languages: you can create your own language and give it the code ``gqjgjq`` and it will work fine. That's why they provide the ``iso`` property as an optional one in their locale definition, so, if you ever need it, you can also check the standardised code instead of the one you want to use internally. But we only care about the internal one (anyway, define properly the iso one with those new languages you added just in case we need it in the future)

The idea to simplify our work is to match the i18n codes with datefns ones, so ``$i18n.locale`` always return the appropiate datefns one. If we use our custom codes (like we did in the past, using ``en`` instead of ``enUS``), we would need a mixin for that, which is less convenient.

If it wasn't for ``datefns`` we could call our locale codes as ``aaaaa`` or ``bbb`` perfectly :)

Hope this cleared out your doubts.

ThibaultNocchi:
Blocked until https://github.com/nuxt-community/date-fns-module/pull/51 gets released

ThibaultNocchi:
Just posting my regex to clean empty lines:
`^\s*".+":\s*"",?$`

Camc314:
@ThibaultNocchi could we move removing empty strings into a seperate PR, so we can get that merged?

ThibaultNocchi:
Yup I can, I'll do this this afternoon if you don't mind ;)

Camc314:
Sounds good to me 👍

ferferga:
@jellyfin rebase

ferferga:
@jellyfin rebase

MrTimscampi (changes requested):
Missing the following locales:

* am.json
* ar.json
* be_Latn.json
* eo.json
* fa.json
* fil.json
* he.json
* hi.json
* it.json
* ja.json
* ms.json
* nn.json
* pt.json
* sr_Latn.json
* sw.json
* uk.json
* ur.json
* zh_Hant.json

Otherwise, all seems good :+1:

MrTimscampi (approved):

ferferga (changes requested):
Why the rename from ``enUS`` to ``en-US``. It's exported without the dash:

https://github.com/date-fns/date-fns/blob/master/src/locale/index.js#L22

ThibaultNocchi:
@ferferga This is due to how nuxt date fns now imports the locales: https://github.com/nuxt-community/date-fns-module/commit/8a1673b713b154580b06dae909dc6da4de2c0f28#diff-11af61311ab29f43ed180a25e84cbae6a92c72a244b8c4f717d573ec930ff5acR6-R12

MrTimscampi (approved):

ferferga (approved):

ferferga:
@jellyfin-bot rebase

PR #725, opened 2021-02-13
perf(build): optimize build size

ferferga:
Quick pass for optimizing our bundle size.

We might need to do the same for all the imports of the axios client, however, unless we don't have the sources (https://github.com/jellyfin/jellyfin-meta/issues/10), I can't check if we can prepare it for better treeshaking

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -268,6 +268,8 @@ const config: NuxtConfig = {
         server: '#424242'
       }
     },
+    optimizeCSS: true,
+    extractCSS: true,
     babel: {
       // envName: server, client, modern
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/components/Forms/LoginForm.vue
+++ b/components/Forms/LoginForm.vue
@@ -62,7 +62,7 @@
 </template>
 
 <script lang="ts">
-import { isEmpty } from 'lodash';
+import isEmpty from 'lodash/isEmpty';
 import Vue from 'vue';
 import { mapActions } from 'vuex';
 import { UserDto } from '@jellyfin/client-axios';
--- a/components/Item/ItemGrid.vue
+++ b/components/Item/ItemGrid.vue
@@ -48,7 +48,7 @@
 </template>
 
 <script lang="ts">
-import { chunk } from 'lodash';
+import chunk from 'lodash/chunk';
 import Vue from 'vue';
 import { BaseItemDto } from '@jellyfin/client-axios';
 
--- a/components/Item/Metadata/MetadataEditor.vue
+++ b/components/Item/Metadata/MetadataEditor.vue
@@ -211,7 +211,8 @@
 
 <script lang="ts">
 import Vue from 'vue';
-import { pick, set } from 'lodash';
+import pick from 'lodash/pick';
+import set from 'lodash/set';
 import { BaseItemDto, BaseItemPerson } from '@jellyfin/client-axios';
 
 export default Vue.extend({
--- a/components/Players/AudioPlayer.vue
+++ b/components/Players/AudioPlayer.vue
@@ -12,7 +12,7 @@
 <script lang="ts">
 import Vue from 'vue';
 import { stringify } from 'qs';
-import { throttle } from 'lodash';
+import throttle from 'lodash/throttle';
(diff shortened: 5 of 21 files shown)


ThibaultNocchi (commented):
Nice! Do you have some stats? :)

Camc314 (commented):
I might be wrong, but I don't think the changing of is imports makes and difference?

ferferga:
## Bundle size
*All sizes are the total of JS bundles in modern build in client-side*

**Master**
4.41 MB

**lodash imports**
4.36 MB (5% reduction)

**lodash imports + CSS extraction**
4.34 MB (7% reduction)

## Webpack analysis
**Master**
image

**This PR**
image

*lodash is near vuei18n*

ferferga:
As you can see, there are a lot of modules we can still optimize, being the most preminent one the axios client, where you can see that we're bundling DLNA and stuff we're not using yet.

Axios client alone is ~1MB of the bundle size.

MrTimscampi:
The multiple lines of lodash imports are cumbersome, imo, and impact maintainability quite a bit (Think of a module using dozens of lodash functions. That's dozens of import lines).

Lodash themselves recommend [this Babel plugin](https://github.com/lodash/babel-plugin-lodash) and [this webpack plugin](https://github.com/lodash/lodash-webpack-plugin), which seem to produce by far the smallest bundles (at the expense of some build time), according to [this benchmark](https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark).

MrTimscampi:
Here is what the bundles look like for me with the two plugins above:
image

ferferga:
@MrTimscampi This is what I tried at first but the client didn't load at all with it. In fact I prefer yours.

Mind you amending my commit?

Camc314:
@ferferga I have updated to use webpack plugins

ThibaultNocchi:
I get an error when compiling about chunk order or something:

```
chunk pages/artist/_itemId/index/pages/genre/_itemId/index/pages/index/pages/item/_itemId/index/pages/libr/320f409c [extract-css-chunks-webpack-plugin]
Conflicting order. Following module has been added:
 * css ./node_modules/css-loader/dist/cjs.js??ref--7-oneOf-1-1!./node_modules/vue-loader/lib/loaders/stylePostLoader.js!./node_modules/postcss-loader/src??ref--7-oneOf-1-2!./node_modules/sass-loader/dist/cjs.js??ref--7-oneOf-1-3!./node_modules/@nuxt/components/dist/loader.js??ref--0-0!./node_modules/vue-loader/lib??vue-loader-options!./components/Item/Card/Card.vue?vue&type=style&index=0&id=42862b4e&lang=scss&scoped=true&
despite it was not able to fulfill desired ordering with these modules:
 * css ./node_modules/css-loader/dist/cjs.js??ref--3-oneOf-1-1!./node_modules/vue-loader/lib/loaders/stylePostLoader.js!./node_modules/postcss-loader/src??ref--3-oneOf-1-2!./node_modules/@nuxt/components/dist/loader.js??ref--0-0!./node_modules/vue-loader/lib??vue-loader-options!./components/Skeletons/SkeletonCard.vue?vue&type=style&index=0&id=fc90ba3e&scoped=true&lang=css&
   - couldn't fulfill desired order of chunk group(s) pages/genre/_itemId/index, pages/library/_viewId, pages/person/_itemId/index, pages/index
   - while fulfilling desired order of chunk group(s) pages/artist/_itemId/index, pages/item/_itemId/index, pages/musicalbum/_itemId/index, pages/series/_itemId/inde
```

nuxt.config.ts
```diff
@@ -270,6 +270,8 @@ const config: NuxtConfig = {
         server: '#424242'
       }
     },
+    optimizeCSS: true,
+    extractCSS: true,
```
ThibaultNocchi, line 274: So do we keep this?

ferferga: @ThibaultNocchi I'm not having any issues building locally. Have you tried in an perfectly clean environment? I wonder if you're getting conflicts with cache artifacts.

ThibaultNocchi (approved):

MrTimscampi (approved):

PR #726, opened 2021-02-13
chore(husky): fix husky hook permissions and command

MrTimscampi:
No idea how this was even merged in this state, but Husky is completely broken.

Neither of the scrips had executable permissions (preventing them from even running) and the `lint-staged` one didn't work at all, since tools need to be called through yarn or npx [as mentioned in the migration documentation](https://typicode.github.io/husky/#/?id=locally-installed-binaries)

This fixes both, making Husky actually work.


ferferga:
I'm on Windows and I have my git tooling separate from my node environment (in WSL) so I can't test this and, as nobody complained, I thought it was working all fine.

@Camc314 @ThibaultNocchi can you guys verify this please?

ThibaultNocchi (approved):

ThibaultNocchi:
Ok yeah that was the reason why I got weird bash outputs when committing lol

Camc314 (approved):

PR #764, opened 2021-02-18
Add an item store

ThibaultNocchi:
This PR will migrate our current different stores to a single store for the item storage

Part of #801

Added the HTML and lcov coverage reporters to Jest

### DIFF
--- a/store/items.ts
+++ b/store/items.ts
@@ -40,8 +40,8 @@ export const mutations: MutationTree<ItemsState> = {
       state.allIds.splice(idx, 1);
     }
   },
-  CLEAR_STATE() {
-    this.replaceState(defaultState());
+  CLEAR_STATE(state: ItemsState) {
+    Object.assign(state, defaultState());
   }
 };
 
--- a/store/items.ts
+++ b/store/items.ts
@@ -26,7 +26,7 @@ export const getters: GetterTree<ItemsState, ItemsState> = {
 export const mutations: MutationTree<ItemsState> = {
   ADD_ITEM(state: ItemsState, { item }: { item: BaseItemDto }) {
     if (!item.Id) {
-      throw new Error('No item ID');
+      throw new Error("No item ID in provided item, can't store it");
     }
     Vue.set(state.byId, item.Id, item);
     if (!state.allIds.includes(item.Id)) {
--- a/store/items.ts
+++ b/store/items.ts
@@ -1,0 +1,71 @@
+import Vue from 'vue';
+import { BaseItemDto } from '@jellyfin/client-axios';
+import { ActionTree, GetterTree, MutationTree } from 'vuex';
+import map from 'lodash/map';
+import some from 'lodash/some';
+import forEach from 'lodash/forEach';
+
+export interface ItemsState {
+  byId: Record<string, BaseItemDto>;
+  allIds: string[];
+}
+
+export const defaultState = (): ItemsState => ({
+  byId: {},
+  allIds: []
+});
+
+export const state = defaultState;
+
+export const getters: GetterTree<ItemsState, ItemsState> = {
+  getItem: (state) => (id: string): BaseItemDto | undefined => state.byId[id],
+  getItems: (state) => (ids: string[]): BaseItemDto[] =>
+    map(ids, (id) => state.byId[id])
+};
+
+export const mutations: MutationTree<ItemsState> = {
+  ADD_ITEM(state: ItemsState, { item }: { item: BaseItemDto }) {
+    if (!item.Id) {
+      throw new Error('No item ID');
+    }
+    Vue.set(state.byId, item.Id, item);
+    if (!state.allIds.includes(item.Id)) {
+      state.allIds.push(item.Id);
+    }
+  },
+  DELETE_ITEM(state: ItemsState, { id }: { id: string }) {
+    delete state.byId[id];
+    const idx = state.allIds.indexOf(id);
+    if (idx > -1) {
+      state.allIds.splice(idx, 1);
+    }
+  },
+  CLEAR_STATE() {
+    this.replaceState(defaultState());
+  }
+};
+
+export const actions: ActionTree<ItemsState, ItemsState> = {
+  addItem({ commit }, { item }: { item: BaseItemDto }) {
+    commit('ADD_ITEM', { item });
+  },
+  addItems({ commit }, { items }: { items: BaseItemDto[] }) {
+    if (some(items, (item) => !item.Id)) {
+      throw new Error('Item missing ID');
+    }
+    forEach(items, (item) => {
+      commit('ADD_ITEM', { item });
+    });
+  },
+  deleteItem({ commit }, { id }: { id: string }) {
+    commit('DELETE_ITEM', { id });
+  },
+  deleteItems({ commit }, { ids }: { ids: string[] }) {
+    forEach(ids, (id) => {
+      commit('DELETE_ITEM', { id });
+    });
+  },
+  clearState({ commit }) {
+    commit('CLEAR_STATE');
+  }
+};
--- a/jest-helpers.d.ts
+++ b/jest-helpers.d.ts
@@ -1,0 +1,8 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { ActionContext } from 'vuex';
+import { RootState } from '~/store';
+
+type ModuleAction<ModuleState> = (
+  injectee: ActionContext<ModuleState, RootState>,
+  payload: any
+) => any;
--- a/jest.config.js
+++ b/jest.config.js
@@ -25,5 +25,5 @@ module.exports = {
   ],
   testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$',
   setupFiles: ['jest-canvas-mock', './jest.setup.ts'],
-  coverageReporters: ['text', 'cobertura']
+  coverageReporters: ['text', 'cobertura', 'html', 'lcov']
 };
--- a/store/__tests__/items.spec.ts
+++ b/store/__tests__/items.spec.ts
@@ -1,0 +1,307 @@
+import Vuex, { Store, ActionContext } from 'vuex';
+import { BaseItemDto } from '@jellyfin/client-axios';
+import cloneDeep from 'lodash/cloneDeep';
+import {
+  state,
+  mutations,
+  ItemsState,
+  defaultState,
+  actions,
+  getters
+} from '../items';
+import type { ModuleAction } from '~/jest-helpers.d.ts';
+import { RootState } from '~/store';
+
+let store: Store<ItemsState>;
+let mockCommit: jest.Mock;
+let mockDispatch: jest.Mock;
+
+let addItem: ModuleAction<ItemsState>;
+let addItems: ModuleAction<ItemsState>;
+let deleteItem: ModuleAction<ItemsState>;
+let deleteItems: ModuleAction<ItemsState>;
+let clearState: ModuleAction<ItemsState>;
+
+const id1 = 'ID';
+const name1 = 'Movie 1';
+const item1: BaseItemDto = { Id: id1, Name: name1 };
+
+const id2 = 'ID2';
+const name2 = 'Movie 2';
+const item2: BaseItemDto = { Id: id2, Name: name2 };
+
+const name1Bis = 'Movie 1Bis';
+const item1Bis: BaseItemDto = { Id: id1, Name: name1Bis };
+
+const itemInvalid: BaseItemDto = {};
+
+beforeEach(() => {
+  store = new Vuex.Store(cloneDeep({ state, mutations, actions, getters }));
+  addItem = (actions.addItem as unknown) as ModuleAction<ItemsState>;
+  addItems = (actions.addItems as unknown) as ModuleAction<ItemsState>;
+  deleteItem = (actions.deleteItem as unknown) as ModuleAction<ItemsState>;
+  deleteItems = (actions.deleteItems as unknown) as ModuleAction<ItemsState>;
+  clearState = (actions.clearState as unknown) as ModuleAction<ItemsState>;
+
+  mockCommit = jest.fn();
+  mockDispatch = jest.fn();
+});
+
+afterEach((): void => {
+  mockCommit.mockReset();
+  mockDispatch.mockReset();
+});
+
+describe('Vuex: items', () => {
+  it('adds an item when ADD_ITEM is commited with valid ID', (): void => {
+    store.replaceState({ ...defaultState() });
+
+    store.commit('ADD_ITEM', { item: item1 });
+
+    expect(store.state.byId[id1]).toBe(item1);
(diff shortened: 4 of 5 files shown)


components/Layout/HomeHeader/HomeHeader.vue
```diff
@@ -63,14 +67,8 @@ export default Vue.extend({
...
 
-      const itemData = (
-        await this.$api.userLibrary.getItem({
-          userId: this.$auth.user?.Id,
-          itemId: id
-        })
-      ).data;
-
-      this.relatedItems[key] = itemData;
+      await this.fetchItem({ id });
```
ferferga, line 70: What about passing to this method a function as an argument? So we can pass it whatever Axios function and run it in the store instead of the component's context. That way we can query whole libraries or whatever we need instead of doing it item by item.

Also, we shouldn't await this so the component load isn't blocked.

ferferga: You can in the store then check if the response is an array or a single item and, if it's an array, add each of the inner objects.

You can do this easily with lodash

ThibaultNocchi: I don't wanna use lodash too much but yeah :p

MrTimscampi: > So we can pass it whatever Axios function and run it in the store instead of the component's context

I think a better way would be to make multiple actions for whatever we need. Most of the time, we're going to be touching this store from somewhere else anyway, so most of the API calls would be done elsewhere, I would think.

The Websocket update is really the only place where I can see us fetching things from the item store. All the other API calls would be triggered from somewhere else, then processed a bit to copy IDs as an array and `addOrUpdateItems` to the Items store.

store/items.ts
```diff
@@ -0,0 +1,61 @@
...
+export const defaultState = (): ItemsState => ({});
+
+export const state = defaultState;
+
+export const getters: GetterTree<ItemsState, ItemsState> = {
+  getItem: (state) => (id: string): BaseItemDto | undefined => state[id]
+};
+
+export const mutations: MutationTree<ItemsState> = {
+  INSERT_ITEM(state: ItemsState, { item }: { item: BaseItemDto }) {
```
MrTimscampi, line 15: This could be renamed to `ADD_ITEM`.

store/items.ts
```diff
@@ -0,0 +1,63 @@
...
+import reduce from 'lodash/reduce';
+
+export interface ItemsState {
+  items: Record<string, BaseItemDto>;
+  ids: string[];
+}
+
+export const defaultState = (): ItemsState => ({
+  items: {},
+  ids: []
```
MrTimscampi, line 13:
```suggestion
  byId: {},
  allIds: []
```

This removes some repetition in the access and is more explicit about what is what.

Instead of doing `$store.state.items.items` you'd do `$store.state.items.byId`.

store/items.ts
```diff
@@ -0,0 +1,63 @@
...
+export const getters: GetterTree<ItemsState, ItemsState> = {
+  getItem: (state) => (id: string): BaseItemDto | undefined => state.items[id],
+  getItems: (state) => (ids: string[]): Record<string, BaseItemDto> =>
+    reduce(
+      ids,
+      (res, el) => {
+        res[el] = state.items[el];
+        return res;
+      },
+      {} as Record<string, BaseItemDto>
```
MrTimscampi, line 27: We want the return of this to be `BaseItemDto[]`, not a record

store/items.ts
ThibaultNocchi: Though it's only `$store.state.items` or `$store.state.ids` currently, there's not an `items` object root :)

MrTimscampi: When you access `$store.state` in the app, you get the RootState, which is this:
image

So you'll have `state.items.items` and `state.items.ids` currently.

ThibaultNocchi: Noted

MrTimscampi (approved):
LGTM :)

store/items.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    Vue.set(state.byId, item.Id, item);
+    if (!state.allIds.includes(item.Id)) {
+      state.allIds.push(item.Id);
+    }
+  },
+  DELETE_ITEM(state: ItemsState, { id }: { id: string }) {
+    delete state.byId[id];
+    const idx = state.allIds.indexOf(id);
+    if (idx > -1) {
```
ferferga, line 39:
```suggestion
    if (idx !== undefined) {
```

store/items.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    forEach(items, (item) => {
+      commit('ADD_ITEM', { item });
+    });
+  },
+  deleteItem({ commit }, { id }: { id: string }) {
+    commit('DELETE_ITEM', { id });
+  },
+  deleteItems({ commit }, { ids }: { ids: string[] }) {
+    forEach(ids, (id) => {
```
ferferga, line 64: Why lodash's function instead of ids.forEach?

store/items.ts
```diff
@@ -0,0 +1,71 @@
...
+export const getters: GetterTree<ItemsState, ItemsState> = {
+  getItem: (state) => (id: string): BaseItemDto | undefined => state.byId[id],
+  getItems: (state) => (ids: string[]): BaseItemDto[] =>
+    map(ids, (id) => state.byId[id])
+};
+
+export const mutations: MutationTree<ItemsState> = {
+  ADD_ITEM(state: ItemsState, { item }: { item: BaseItemDto }) {
+    if (!item.Id) {
+      throw new Error('No item ID');
```
ferferga, line 29: I feel this is too generic

store/items.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    Vue.set(state.byId, item.Id, item);
```
ferferga, line 31: Why we need this? We shouldn't need it in Vuex as everything is reactive and we're doing a single level deep.

https://forum.vuejs.org/t/vuex-best-practices-for-complex-objects/10143/2

store/items.ts
```diff
@@ -0,0 +1,71 @@
+import Vue from 'vue';
+import { BaseItemDto } from '@jellyfin/client-axios';
+import { ActionTree, GetterTree, MutationTree } from 'vuex';
+import map from 'lodash/map';
+import some from 'lodash/some';
+import forEach from 'lodash/forEach';
+
+export interface ItemsState {
+  byId: Record<string, BaseItemDto>;
+  allIds: string[];
```
ferferga, line 10: What's the difference between this and Object.keys(byId)? Why do we need this specifically?

store/items.ts
ThibaultNocchi: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf#return_value though?

store/items.ts
ThibaultNocchi: They are a little bit quicker ^^

store/items.ts
ThibaultNocchi: Done

store/items.ts
ThibaultNocchi: Cause it creates a key, and Vuex doesn't react to key creation. See point 2 in screenshot https://vuex.vuejs.org/guide/mutations.html 
print_screen_2021-03-04-23-54-59

store/items.ts
ThibaultNocchi: This is inspired by Redux way of doing (cc @MrTimscampi). It allows both to save time instead of doing Object.keys and allows to store order.

MrTimscampi: More or less, Object.keys() makes a copy of the object, which is expensive. Objects also don't guarantee any sort of order is kept.

The reasoning is that doing this helps us keep the order, if we need it, and avoids doing potentially expensive copies every time we need to check if something is in the store.

store/items.ts
Codex-: Afaik that's only true for sparse arrays, otherwise just use Array.forEach

Unless the performance is significantly better than native forEach then I'd steer clear of lodash

store/items.ts
```diff
@@ -0,0 +1,71 @@
...
+  byId: {},
+  allIds: []
+});
+
+export const state = defaultState;
+
+export const getters: GetterTree<ItemsState, ItemsState> = {
+  getItem: (state) => (id: string): BaseItemDto | undefined => state.byId[id],
+  getItems: (state) => (ids: string[]): BaseItemDto[] =>
+    map(ids, (id) => state.byId[id])
```
Codex-, line 23: Why not native map? Targeting < es6?

store/items.ts
```diff
@@ -0,0 +1,71 @@
+import Vue from 'vue';
+import { BaseItemDto } from '@jellyfin/client-axios';
+import { ActionTree, GetterTree, MutationTree } from 'vuex';
+import map from 'lodash/map';
+import some from 'lodash/some';
+import forEach from 'lodash/forEach';
```
Codex-, line 6: Every lodash import here is provided natively in es6 on the array prototype, unless you're getting significant performance improvements from them but even then it seems very extraneous

store/items.ts
ferferga: @Codex- https://www.measurethat.net/Benchmarks/Show/3470/0/lodash-foreach-vs-native-foreach

I''ve been researching and It seems that Chrome JS engine does while loops faster than forEach, while in Firefox both of them are the same. Lodash already takes care of that and chooses the best method based on what's faster in each environment.

store/items.ts
ferferga: Could've said that I got an ``undefined`` from this function the other day. Good to know!

store/items.ts
ferferga: @Codex- See here https://github.com/jellyfin/jellyfin-vue/pull/725

Codex-: Sorry I don't quite understand what you're referring to with that PR link. The PR is in regards to the bundle size, and using direct imports to better optimize the package.

This has nothing to do with my question about why you're opting to use lodash functions for methods that are provided natively on the prototype in es6+

I understand the desire to use lodash methods for things that are not provided natively (although I do disagree with that too [when we have so many better options](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore), in addition to the countless tech blogs, articles, security reasons, maintainability reasons, and so on). Don't get me wrong, lodash served a great purpose back in the day, but we're living in a better time now. 

Using `lodash/forEach` over `Array.prototype.forEach` seems very extraneous. Same with not simply using `Array.prototype.map` and `Array.prototype.some`.

ferferga: @Codex- Oh, I understand now. I thought you were mentioning the import thing because you raised the topic in another comment as well.

See here: 

In general, all the lodash functions seems to perform better than the native counterparts.

store/items.ts
ThibaultNocchi: Honestly it's a pros and cons question. Lodash's implem is quicker, but yeah I also like using the maps. @MrTimscampi could explain that to you :)

MrTimscampi: Lodash is generally faster: https://www.measurethat.net/Benchmarks/Show/3518/5/lodash-map-vs-es6-map-v2

store/items.ts
Codex-: In the link you sent re the other thread: https://www.measurethat.net/Benchmarks/Show/3470/0/lodash-foreach-vs-native-foreach

It shows even there that lodash performs worse than native based on the last run:
image

When running that same benchmark myself for chrome stable:
image

Firefox stable:
image

In general, from the benchmarks i've found on that same benchmark site, lodash performs worse in nearly every case? Except map which is similar enough that including lodash also seems like a waste of time. 

Also using the native methods means that as v8 / spidermonkey move forward with their own optimizations you're not locked to lodash's explicit implementation of a given method.

In the early days Lodash absolutely outperformed the native implementations of these methods, and provided great value back with <=es5 but this just isn't the case anymore. Not to a strong enough extend that I would be using lodash for these.

store/items.ts
Codex-: https://github.com/jellyfin/jellyfin-vue/pull/764#discussion_r588967462

store/items.ts
MrTimscampi: Lodash is faster for me (Latest Firefox on Linux):
image

Same thing on the latest Chromium, where the difference is even bigger:
image

Codex-: Is it really worth using these for what is barely a 14% gain in inconsistent testing though? I really don't think it's worth it unless you're explicitly targeting ancient versions of spidermonkey / v8.

Ultimately it's up to you guys / the maintainers and what they're willing to put up with as far as a healthy & maintainable the code base is / will be. 

Lodash at this point is less sustainable / maintainable on almost every front, performance (negligible gains), spaghetti consumption (prototype native mixing / confusion), higher maintenance costs (i.e. [this was only one version prior](https://snyk.io/test/npm/lodash/4.17.9)), amongst many other things.

¯\_(ツ)_/¯

MrTimscampi: > I really don't think it's worth it unless you're explicitly targeting ancient versions of spidermonkey / v8.

Well, we're technically planning support for old versions of WebOS and Tizen, so Lodash makes sense there. You're also (afaik) the only one between the four of us who got better scores on native than lodash for these benchmarks, so I'm not sure what's up.

The TV UI is pretty far off, but the performance difference, from what we've seen, is pretty negligible (or in favor of Lodash, though not in your case) on evergreen browsers, while older ones will benefit from Lodash, so we may as well use it.

h1dden-da3m0n: Sorry to join in here, but I too got better/equal *native scores* (linux-arch, ff & brave).
However, this only proves the point that its incredibly system, OS and environment dependent and with the target of aincient platforms such as WebOS or Tizen (Browser support speaking) there likely is no way around *lodash*.

(Disclaimer: I am no Web dev, so my knowledge is very limited in this area)

Codex-: > we're technically planning support for old versions of WebOS and Tizen
If this is the official support case, then consider using shims too. 

Unless you're expecting like 20%+ to be using Tizen / WebOS on ancient versions (i'd be surprised if any actually being used by Jellyfin users were old enough to not support es6?) I still wouldn't do it.

I'd probably look into using analytics to determine the weak points in performance for users though, not sure what the state of analytics data in Jellyfin web/vue is though.

WebOS appears to be based on Android, which is an interesting thing. If it's uses the AOSP base without play services then the WebVew is locked to the chromium version of that base (i.e. if it's AOSP of 8.0, you'll get the webview at the time of that release specifically and no newer version ever). Tizen seems entirely proprietary so who knows whether or not the webview gets updated here lol.

> Sorry to join in here, but I too got better/equal native scores (linux-arch, ff & brave).
Yeah I tested on mobile (Android 10), Elementary OS with Chromium, macos 10.15.7, and Windows 10 to see how the results may vary. In nearly all of these tests lodash only ever outperformed native on Windows 10 and only for `map` but about ~10%.

ferferga (approved):

PR #767, opened 2021-02-19
chore: rework CI and bump nuxt and eslint-plugin-jsdoc

ferferga:
CI needed an update for working with nuxt 2.15

- Completely done in GitHub actions
- Easily structured and understandable in comparison with Azure
- Removed Docker image building on pull requests and only do it in pushes to master or releases
- Uses always latest LTS of node
- Better release schema with the following tags, automatically pushed on the relevant events, as follows:

### unstable (on each commit to master)
- SSR

**unstable-ssr**
**unstable-ssr.2021-02-19.``sha``**

- Static

**unstable**
**unstable.2021-02-19.``sha``**

***sha** being the short hash of the commit in master that triggered the build*

### rc (on prerelease releases)

- SSR

**stable-rc-ssr**
**stable-rc-ssr.``tag``**

- Static

**stable-rc**
**stable-rc.``tag``**

*Being tag the release tag that triggered the build*

### stable (on stable releases)

- SSR

**stable-ssr**
**stable-ssr.``tag``**

- Static

**latest**
**stable**
**stable.``tag``**

*Being tag the release tag that triggered the build*

## Other changes

- Removed .yarnclean as it was causing issues with Nuxt 2.15 auto component detection
- Moved types packages to devDeependencies


ThibaultNocchi (changes requested):

.github/workflows/main.yml
```diff
@@ -43,27 +41,30 @@ jobs:
             ${{ runner.os }}-yarn-
 
       - name: Install dependencies
-        run: yarn
+        run: yarn install --froze-lockfile
```
ThibaultNocchi, line 44: Typo here

.github/workflows/main.yml
```diff
@@ -43,27 +41,30 @@ jobs:
...
+          ref: 'master'
 
       - name: Setup node env
         uses: actions/setup-node@v2.1.4
         with:
-          node-version: ${{ matrix.node }}
+          node-version: 14
+
+      - name: Install dependencies
+        run: yarn install --froze-lockfile
```
ThibaultNocchi, line 67: Typo here

.github/workflows/docker.yml
```diff
@@ -0,0 +1,67 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: |
+            jellyfin/jellyfin-vue:unstable-ssr
+            jellyfin/jellyfin-vue:unstable-ssr.${{ steps.date.outputs.date }}.${{ github.sha }}
```
ThibaultNocchi, line 37: Is there a reason for tagging with the date?

.github/workflows/build.yml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/cache@v2.1.4
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --froze-lockfile
```
ThibaultNocchi, line 44: Typo :p

.github/workflows/build.yml
```diff
@@ -0,0 +1,83 @@
...
```
ThibaultNocchi, line 80: Typo :p

ThibaultNocchi (changes requested):

.github/workflows/docker.yml
ferferga: I thought that it gave more freedom when looking for images in docker hub, so one doesn't need to go back all the git history to check for an specific day. We have dates right now in our pushed tags, so I think it makes sense to keep them somehow.

With date and short hash, I think everyone will be pleased :D

h1dden-da3m0n: FYI: `github.sha` is the full hash not the short hash

MrTimscampi:
> Uses always latest LTS of node

This part is bound to cause issues, since tools and their dependencies don't always update as soon as a new LTS is released and the new versions might break things (See for example some of the issues we've had previously with the docker crashing on launch with Node 14 on the previous version of Nuxt).

.github/workflows/docker.yml
ThibaultNocchi: Maybe we can put the short one then @ferferga ? :p

Dockerfile
```diff
@@ -1,21 +1,20 @@
-FROM node:12-alpine AS build
+FROM node:lts-alpine AS build
```
MrTimscampi, line 1:
```suggestion
FROM node:14-alpine AS build
```

Could we change these? Manually controlling the version should avoid us a bunch of breakage down the line, as explained.

MrTimscampi (changes requested):
Looks mostly right, aside from the safety of using the LTS-tagged image of node and the short-SHA1 mentioned above.

I'd like to get @Shadowghost 's input on the Dockerfiles, as usual.

MrTimscampi:
We may also need intervention from Core for the Dockerhub info.

.github/workflows/docker.yml
h1dden-da3m0n: the only way to get a short SHA would be to build it yourself in a step before use. [Something I stumbled over](https://stackoverflow.com/a/59819441) while I searched for this functionality for another repo. 

(note: I ended up not using the SHA so not 100% sure It would work, but looked good to me)

ferferga: @h1dden-da3m0n Done!

.github/workflows/docker.yml
```diff
@@ -16,6 +15,40 @@ jobs:
         id: date
         run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
 
+      - name: Get commit hash
+        run: echo "SHORT_SHA=`echo ${GITHUB_SHA} | cut -c1-8`" >> $GITHUB_ENV
```
h1dden-da3m0n, line 19: Nitpick but if you have the power of a full bash you could use the following instead of using `cut`:

```bash
echo "SHORT_SHA=`echo ${GITHUB_SHA::8}`" >> $GITHUB_ENV
```

.github/workflows/docker.yml
```diff
@@ -45,6 +81,40 @@ jobs:
...
```
h1dden-da3m0n, line 85: same as the above, however this may be preference :wink:

ferferga:
Switched to npm registry as, in my experience, it was much better and faster while getting packages than the yarn registry

ferferga:
Images built from the workflow of my test repository are available at https://hub.docker.com/r/ferferga/jellyfin-vue

h1dden-da3m0n:
Curious question and likely unrelated, but have you considered hosting the Jellyfin Container Images on an additional Registries such as ghcr.io or quay.io yet? (feel free to ignore since it is not PR related)

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,153 @@
...
+      - name: Configure Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build docker images and push static stable
```
MrTimscampi, line 130:
```suggestion
      - name: Build and push stable static images
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,153 @@
...
+        with:
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm,linux/arm/v6
+          file: '.github/workflows/Dockerfile.ssr'
+          tags: |
+            jellyfin/jellyfin-vue:stable-ssr
+            jellyfin/jellyfin-vue:stable-ssr.${{ github.event.release.tag_name }}
+
+      - name: Build docker images and push SSR candidate
```
MrTimscampi, line 70:
```suggestion
      - name: Build and push release candidate SSR Docker images
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,153 @@
...
+      - name: Configure Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build docker images and push SSR stable
```
MrTimscampi, line 58:
```suggestion
      - name: Build and push stable SSR Docker images
```

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,153 @@
...
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm,linux/arm/v6
+          file: '.github/workflows/Dockerfile.nginx'
+          tags: |
+            jellyfin/jellyfin-vue:stable
+            jellyfin/jellyfin-vue:stable.${{ github.event.release.tag_name }}
+            jellyfin/jellyfin-vue:latest
+
+      - name: Build docker images and push static candidate
```
MrTimscampi, line 143:
```suggestion
      - name: Build and push release candidate static images
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,139 @@
+name: Docker unstable release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  ssr:
+    name: Build SSR docker image
```
MrTimscampi, line 10:
```suggestion
    name: Build SSR image
```

It's already mentioned in the workflow name that it's for Docker.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,139 @@
...
+      - name: Configure Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build docker images and push
```
MrTimscampi, line 64:
```suggestion
      - name: Build images and push
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,139 @@
...
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm,linux/arm/v6
+          file: '.github/workflows/Dockerfile.ssr'
+          tags: |
+            jellyfin/jellyfin-vue:unstable-ssr
+            jellyfin/jellyfin-vue:unstable-ssr.${{ steps.date.outputs.date }}.${{ env.SHORT_SHA }}
+
+  static:
+    name: Build static docker image
```
MrTimscampi, line 76:
```suggestion
    name: Build static image
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,139 @@
...
+      - name: Configure Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build docker images and push
```
MrTimscampi, line 130:
```suggestion
      - name: Build images and push
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,139 @@
...
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm,linux/arm/v6
+          file: '.github/workflows/Dockerfile.nginx'
+          tags: |
+            jellyfin/jellyfin-vue:unstable
```
MrTimscampi, line 138: SSR should be the default unstable as well.

.github/workflows/main.yml
```diff
@@ -1,87 +1,86 @@
...
+          ref: 'master'
 
-      - name: Setup node env
+      - name: Configure node environment
         uses: actions/setup-node@v2.1.4
         with:
-          node-version: ${{ matrix.node }}
+          node-version: 14
+
+      - name: Install dependencies
```
MrTimscampi, line 66: I believe this should be done after the cache step, since that step restores the cache from the previous run (if it matches)

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,153 @@
...
+
+on:
+  release:
+    types:
+      - released
+      - prereleased
+
+jobs:
+  ssr:
+    name: Build SSR docker image
```
MrTimscampi, line 11:
```suggestion
    name: Build SSR image
```

Docker is already part of the workflow, no need to repeat

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,153 @@
...
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm,linux/arm/v6
+          file: '.github/workflows/Dockerfile.ssr'
+          tags: |
+            jellyfin/jellyfin-vue:stable-rc-ssr
+            jellyfin/jellyfin-vue:stable-rc-ssr.${{ github.event.release.tag_name }}
+
+  static:
+    name: Build static docker image
```
MrTimscampi, line 83:
```suggestion
    name: Build static image
```

MrTimscampi (changes requested):
Some step names, an apparent error in step orders for one of the jobs and some changes needed for the Docker tags, but aside from that LGTM.

Once again asking for @Shadowghost 's input on the Dockerfiles, though they look okay to me.

MrTimscampi (approved):
LGTM.

Still waiting for the Docker Hub secrets (Likely for @anthonylavado ) and a double check on the Dockerfiles from @Shadowghost .

Shadowghost:
I'm currently having exams, maybe I find some time on the weekend to check everything in depth.
Some comments after a quick glance:
- Dockerfiles seem to be fine to me
- We're currently building the client once locally and once in stable/unstable docker workflows. Ideally we only build it once locally and copy those generated clients to the images. This would also prevent us from caching node modules between the different workflows
- The above obviously does not work, if the workflows run in parallel (not talking about the parallel docker builds using buildx)

MrTimscampi:
>     * The above obviously does not work, if the workflows run in parallel (not talking about the parallel docker builds using buildx)

I think the workflow here works a bit differently, since we can hook to Github things like Pull Requests and releases.

If I'm reading it right, the images are never built on PRs. Unstables get built on commits to master, while stables get built on releases/pre-releases.  
And normal builds don't really happen on master, since the containers get built.

ThibaultNocchi (commented):
Am I missing something or we're not building on PR anymore?

ferferga:
@MrTimscampi Exactly as you mention.

@Shadowghost I also tried to reproduce the "cache parallelism" problem in my private testing repo and I reached only two conclusions:

• Only chance of conflicts can happen when cache doesn't exist and it needs to be populated.

• In those cases, actions fail gracefully and just ignore, unless there are conflicting changes. As we're installing with the lockfile in a frozen state, this shouldn't happen ever.

Shadowghost:
Okay, if we never build images and normal in one go, the current setup seems fine to me.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,140 @@
...
+    name: Build SSR images
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+
+      - name: Get commit hash
+        run: echo "SHORT_SHA=${GITHUB_SHA::8}" >> $GITHUB_ENV
```
h1dden-da3m0n, line 19: mind me asking why you expose the values in 2 different ways here?
once as a step output (date) and the other as a env (short_sha)

Works perfectly fine just noticed that and was wondering why

.github/workflows/docker.yml
```diff
@@ -0,0 +1,140 @@
...
+    name: Build static images
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+
+      - name: Get commit hash
+        run: echo "SHORT_SHA=${GITHUB_SHA::8}" >> $GITHUB_ENV
```
h1dden-da3m0n, line 86: ditto

h1dden-da3m0n: theoretically you could unify them to one step, however you would sacrifice the clear decouple of the actions

h1dden-da3m0n (approved):
I may have no power, however here my seal of approval :wink: :+1:

.github/workflows/docker.yml
ferferga: At first, this step action was here for creating the appropiate tag based on the release type, and I couldn't have duplicated IDs, so I needed to use env vars. Finally, it seemed cleaner to me to decouple it in another file, so just took the original approach, changed the name and leave it! :D

I can revisit this in another PR where I will integrate this with GH Container Registry.

h1dden-da3m0n: fair enough, was just curious and can definitely be revised on later :wink:

MrTimscampi:
The needed tokens have been added by Anthony.

When Thibault has some time to re-review this and approves it, this is good to got :+1:

ThibaultNocchi (approved):

PR #768, opened 2021-02-19
fix(home sections): fix titles being localized too early to be reactive

ThibaultNocchi:
Home section titles are now reactive to locale changes

### DIFF
--- a/components/Layout/HomeSection.vue
+++ b/components/Layout/HomeSection.vue
@@ -1,7 +1,11 @@
 <template>
   <client-only>
     <swiper-section
-      :title="section.name"
+      :title="
+        section.libraryName === undefined
+          ? $t(section.name)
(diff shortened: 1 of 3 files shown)


components/Layout/HomeSection.vue
```diff
@@ -1,7 +1,11 @@
 <template>
   <client-only>
     <swiper-section
-      :title="section.name"
+      :title="
+        section.libraryName === undefined
```
ferferga, line 5: Don't we need $t here for translating?

ThibaultNocchi: That's the two lines below ;)

ferferga: @ThibaultNocchi *Late night* reviewing tales 🥱

MrTimscampi (approved):

ferferga (approved):

PR #769, opened 2021-02-20
imageUtils rewrite

MrTimscampi:
Rewrites the utility functions to get image URLs, based on Jellyfin-web.

* Allows for image-type preference and proper fallback (For example, Thumbs will be preferred for thumb-shaped cards, but fallback to backdrops, then primary images)
* Properly handles parent items
* Adds blurhash + fallback for artist and person details page

### Thumbs with Backdrop fallback
image

### Proper episode images
image

### DIFF
--- a/client/components/Item/PersonList.vue
+++ b/client/components/Item/PersonList.vue
@@ -11,7 +11,7 @@
           <v-img
             v-if="item.PrimaryImageTag"
             ref="personImg"
-            :src="getImageUrl(item.Id)"
+            :src="getPersonImage(item).url"
           />
           <v-icon v-else class="grey darken-3">mdi-account</v-icon>
         </v-list-item-avatar>
@@ -37,8 +37,8 @@
 
 <script lang="ts">
 import Vue from 'vue';
-import { BaseItemPerson, ImageType } from '@jellyfin/client-axios';
-import imageHelper from '~/mixins/imageHelper';
+import { BaseItemDto, BaseItemPerson } from '@jellyfin/client-axios';
+import imageHelper, { ImageUrlInfo } from '~/mixins/imageHelper';
 import itemHelper from '~/mixins/itemHelper';
 
 export default Vue.extend({
@@ -56,13 +56,14 @@ export default Vue.extend({
     }
   },
   methods: {
-    getImageUrl(itemId: string): string | undefined {
+    getPersonImage(item: BaseItemDto): ImageUrlInfo {
       const element = this.$refs.personImg as HTMLElement;
 
-      return this.getImageUrlForElement(ImageType.Primary, {
-        itemId,
-        element
+      const image = this.getImageUrl(item, {
+        width: element?.clientWidth
       });
+
+      return image;
     }
   }
 });
--- a/client/components/Item/PersonList.vue
+++ b/client/components/Item/PersonList.vue
@@ -9,11 +9,7 @@
       >
         <v-list-item-avatar>
           <v-avatar color="card">
-            <blurhash-image :item="item">
-              <template #placeholder>
-                <v-icon size="16">mdi-account</v-icon>
-              </template>
-            </blurhash-image>
+            <blurhash-image :item="item" icon-size="16" />
           </v-avatar>
         </v-list-item-avatar>
         <v-list-item-content>
--- a/client/components/Item/PersonList.vue
+++ b/client/components/Item/PersonList.vue
@@ -8,12 +8,13 @@
         :to="getItemDetailsLink(item, 'Person')"
       >
         <v-list-item-avatar>
-          <v-img
-            v-if="item.PrimaryImageTag"
-            ref="personImg"
-            :src="getPersonImage(item).url"
-          />
-          <v-icon v-else class="grey darken-3">mdi-account</v-icon>
+          <v-avatar color="card">
+            <blurhash-image :item="item">
+              <template #placeholder>
+                <v-icon size="16">mdi-account</v-icon>
+              </template>
+            </blurhash-image>
+          </v-avatar>
         </v-list-item-avatar>
         <v-list-item-content>
           <v-list-item-title>{{ item.Name }}</v-list-item-title>
@@ -37,8 +38,8 @@
 
 <script lang="ts">
 import Vue from 'vue';
-import { BaseItemDto, BaseItemPerson } from '@jellyfin/client-axios';
-import imageHelper, { ImageUrlInfo } from '~/mixins/imageHelper';
+import { BaseItemPerson } from '@jellyfin/client-axios';
+import imageHelper from '~/mixins/imageHelper';
 import itemHelper from '~/mixins/itemHelper';
 
 export default Vue.extend({
@@ -53,17 +54,6 @@ export default Vue.extend({
     skeletonLength: {
       type: Number,
       default: 0
-    }
-  },
-  methods: {
-    getPersonImage(item: BaseItemDto): ImageUrlInfo {
-      const element = this.$refs.personImg as HTMLElement;
-
-      const image = this.getImageUrl(item, {
-        width: element?.clientWidth
-      });
-
-      return image;
     }
   }
 });
--- a/client/components/Layout/Images/BlurhashImage.vue
+++ b/client/components/Layout/Images/BlurhashImage.vue
@@ -1,6 +1,6 @@
 <template>
   <div ref="imageElement">
-    <div v-if="!error" ref="img" class="absolute">
+    <div v-if="!error" ref="img">
       <v-fade-transition>
         <blurhash-canvas
           v-if="hash"
@@ -23,17 +23,11 @@
         />
       </v-fade-transition>
     </div>
-    <div v-if="!$slots.placeholder" class="absolute icon img">
-      <v-icon
-        class="absolute text--disabled"
-        :size="iconSize"
-        color="white"
-        dark
-      >
+    <slot v-else name="placeholder">
+      <v-icon class="absolute text--disabled" :size="iconSize">
         {{ getItemIcon(item) }}
       </v-icon>
-    </div>
-    <slot v-else name="placeholder" />
+    </slot>
   </div>
 </template>
 
@@ -179,9 +173,5 @@ export default Vue.extend({
 .img {
   color: transparent;
   object-fit: cover;
-}
-
-.icon {
-  z-index: -5;
 }
 </style>
--- a/client/components/Layout/Images/BlurhashImage.vue
+++ b/client/components/Layout/Images/BlurhashImage.vue
@@ -1,16 +1,14 @@
 <template>
   <div ref="imageElement">
     <div v-if="!error" ref="img">
-      <v-fade-transition>
-        <blurhash-canvas
-          v-if="hash"
-          :hash="hash"
-          :width="width"
-          :height="height"
-          :punch="punch"
-          class="absolute"
-        />
-      </v-fade-transition>
+      <blurhash-canvas
+        v-if="hash"
+        :hash="hash"
+        :width="width"
+        :height="height"
+        :punch="punch"
+        class="absolute"
+      />
       <v-fade-transition>
         <img
           v-show="!loading"
@@ -73,7 +71,6 @@ export default Vue.extend({
   data() {
     return {
       image: '' as string | undefined,
-      tag: '' as string,
       loading: true,
       error: false,
       resetting: false
@@ -130,7 +127,6 @@ export default Vue.extend({
         });
 
         this.image = imageInfo.url;
-        this.tag = imageInfo.tag ? imageInfo.tag : 'no-image-tag';
 
         if (!this.image) {
           this.onError();
--- a/client/components/Layout/Images/BlurhashImage.vue
+++ b/client/components/Layout/Images/BlurhashImage.vue
@@ -120,7 +120,7 @@ export default Vue.extend({
       this.$nextTick(() => {
         const element = this.$refs.imageElement as HTMLImageElement;
 
-        const imageInfo = this.getImageUrl(this.item, {
+        const imageInfo = this.getImageInfo(this.item, {
           preferThumb: this.type === ImageType.Thumb,
           preferBanner: this.type === ImageType.Banner,
           preferLogo: this.type === ImageType.Logo,
--- a/client/components/Layout/Images/BlurhashImage.vue
+++ b/client/components/Layout/Images/BlurhashImage.vue
@@ -1,19 +1,19 @@
 <template>
-  <div class="absolute">
+  <div ref="imageElement">
     <div v-if="!error" ref="img" class="absolute">
-      <blurhash-canvas
-        v-if="hash"
-        key="canvas"
-        :hash="hash"
-        :width="width"
-        :height="height"
-        :punch="punch"
-        class="absolute"
-      />
+      <v-fade-transition>
+        <blurhash-canvas
+          v-if="hash"
+          :hash="hash"
+          :width="width"
+          :height="height"
+          :punch="punch"
+          class="absolute"
+        />
+      </v-fade-transition>
       <v-fade-transition>
         <img
           v-show="!loading"
-          :key="`blurhashimage-${item.Id}`"
           class="absolute img"
           :src="image"
           v-bind="$attrs"
@@ -79,6 +79,7 @@ export default Vue.extend({
   data() {
     return {
       image: '' as string | undefined,
+      tag: '' as string,
       loading: true,
       error: false,
       resetting: false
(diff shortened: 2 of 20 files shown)


ferferga:
@jellyfin rebase

client/pages/artist/_itemId/index.vue
```diff
@@ -3,13 +3,20 @@
...
+              color="card"
+              width="100%"
+              height="100%"
+              class="elevation-2"
+            >
+              <blurhash-image
+                v-if="item.ImageTags && item.ImageTags.Primary"
+                :item="item"
+              />
+              <v-icon v-else size="128" dark>mdi-account</v-icon>
```
ferferga, line 17: Why not do this fallback on BlurhashImage like all the items does?

BlurhashImage also has slots now, no need for ``v-else``

client/pages/artist/_itemId/index.vue
```diff
@@ -120,8 +127,10 @@ import imageHelper from '~/mixins/imageHelper';
 import timeUtils from '~/mixins/timeUtils';
 import itemHelper from '~/mixins/itemHelper';
 import { isValidMD5 } from '~/utils/items';
+import BlurhashImage from '~/components/Layout/Images/BlurhashImage.vue';
```
ferferga, line 130: This is unnecessary

client/pages/person/_itemId/index.vue
```diff
@@ -4,14 +4,20 @@
...
+                color="card"
+                width="100%"
+                height="100%"
+                class="elevation-2"
+              >
+                <blurhash-image
+                  v-if="item.ImageTags && item.ImageTags.Primary"
+                  :item="item"
+                />
+                <v-icon v-else size="128" dark>mdi-account</v-icon>
```
ferferga, line 18: Same as above

client/utils/items.ts
```diff
@@ -9,6 +11,26 @@ export const validLibraryTypes = [
...
+  | 'square-card'
+  | 'banner-card';
+
+/**
+ * Determines if the item is a person
+ *
+ * @param {*} item - The item to be checked.
+ * @returns {boolean} Whether the provided item is of type BaseItemPerson.
+ */
+export function isPerson(item: unknown): item is BaseItemPerson {
```
ferferga, line 26: This is already on imageHelper, either remove it from there (preferred as it makes more sense) or remove it from here

client/components/Layout/Images/BlurhashImage.vue
```diff
@@ -1,19 +1,19 @@
...
+          :width="width"
+          :height="height"
+          :punch="punch"
+          class="absolute"
+        />
+      </v-fade-transition>
       <v-fade-transition>
         <img
           v-show="!loading"
-          :key="`blurhashimage-${item.Id}`"
```
ferferga, line 16: How does this work when the component is not remounted? (I.e item update or audio controls)

ferferga (changes requested):

client/mixins/imageHelper.ts
```diff
@@ -6,37 +6,16 @@
...
-  tag?: string;
-  itemId?: string;
-  maxWidth?: number;
-  maxHeight?: number;
-  quality?: number;
-  limitByWidth?: boolean;
-  backdropIndex?: number;
-  checkParent?: boolean;
-};
+export interface ImageUrlInfo {
```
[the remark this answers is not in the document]
MrTimscampi: > EDIT: After reviewing this completely, I don't see where are you using anything else that's not url?

That would be because more refactoring is needed, but I wanted to keep it to images only (More is coming for blurhashes in a follow-up)

It's all combined into one function, because we usually need all of them:

* URL to display
* Blurhash for placeholder
* Tag to check if there's an image

This gives us all three in one method, with the same logic for everything. This prevents us from having to copy-paste the image priority logic 3 different times (like it was done before) and has less risk of them getting different if we make change to one and not the other two.

client/components/Layout/Images/BlurhashImage.vue
MrTimscampi: It works just fine for me.

client/pages/artist/_itemId/index.vue
MrTimscampi: Mostly because this happens:
image

MrTimscampi: 576fdf1 fixes this by using slots properly in `blurhash-image` (There's no need to make a condition, you define the default content by putting it into the slot)

client/pages/artist/_itemId/index.vue
MrTimscampi: Automatic import by VSCode. Removed in 576fdf1

client/pages/person/_itemId/index.vue
MrTimscampi: Fixed in 576fdf1

client/utils/items.ts
MrTimscampi: Removed the other one in 576fdf1.

client/components/Item/PersonList.vue
```diff
@@ -11,7 +11,7 @@
           <v-img
             v-if="item.PrimaryImageTag"
             ref="personImg"
-            :src="getImageUrl(item.Id)"
+            :src="getPersonImage(item).url"
```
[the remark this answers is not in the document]
MrTimscampi: Fixed in 576fdf1

neopc10:
Would this PR be a good opportunity to include usage of Primary Poster images for the latest media in a mixed library? I know those libraries have been largely untouched but for cosmetics this would go a long way.

image

MrTimscampi:
> Would this PR be a good opportunity to include usage of Primary Poster images for the latest media in a mixed library? I know those libraries have been largely untouched but for cosmetics this would go a long way.

The logic in this is the exact same as what the web client does, so whatever it uses for images is what this uses too. If that's not the case for mixed libraries, there may be changes needed elsewhere, which would probably fall under another PR :/

neopc10:
Web is using thumb images. Movies have always used Primary and TV Shows in the past few versions moved to Primary but the mixed were left behind.

client/pages/artist/_itemId/index.vue
ferferga: I don't have a clue why this is happening. Maybe we can make the ``v-if`` [here](https://github.com/jellyfin/jellyfin-vue/blob/master/client/components/Layout/Images/BlurhashImage.vue#L26) an ``v-if-else``? I don't have a idea why this is happening, we're already getting artist images with proper icons in the home header.

I originally placed the condition so BlurhashImage is smart enough to provide icons for fallbacks but we're still able to pass our own custom slots in case we needed a different component or other needs. 

You *shouldn't* need those slots in any case, I'm testing your changes locally to see what could be going on.

ferferga: Okay, something strange iis going on. Master is showing me the ``mid-account`` icon for artists that don't have an image in the Home Header. Your PR shows the ``mdi-disc`` icon instead.

Artists that have images don't have any overlapping icon, so I think you should remove the passed slots altogether and let BlurhashImage handle it like we do in HomeHeader's master and artist cards.

client/mixins/imageHelper.ts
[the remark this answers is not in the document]
ferferga: This leaves us two options I guess:

* Rename the function to ``getImageInfo`` or something like that. ``getImageUrl`` is unrealistic.
* Split the logic of getting tags, aspect ratios, etc in different functions, like it was done before. This might also simplify the arguments that need to be passed.

I'm fine with either.

MrTimscampi:
@neopc10 I fixed the shape issue in 80d2bdf. The images all use the Primary, like other sections do (See the screenshot below)

image

client/pages/artist/_itemId/index.vue
MrTimscampi: > I don't have a clue why this is happening. Maybe we can make the v-if here an v-if-else? I don't have a idea why this is happening, we're already getting artist images with proper icons in the home header.

That part of the code is completely gone.

The way you define a slot's default content isn't with v-if and v-else, but by putting the default inside of the slot.

client/mixins/imageHelper.ts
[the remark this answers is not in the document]
MrTimscampi: I'd rather avoid splitting the logic, since it'll lead to a ton of headaches later on.

client/pages/artist/_itemId/index.vue
MrTimscampi: > You shouldn't need those slots in any case, I'm testing your changes locally to see what could be going on.

I removed the ones for the main artist and person images.

*However* one still remains in PersonList. In order to remove it, we should rework `getShapeFromItemType` to take `BaseItemDto | BaseItemPerson`. But that touches stuff outside of this PR's scope, so I'd rather do it in a followup.

MrTimscampi: > Okay, something strange iis going on. Master is showing me the `mid-account` icon for artists that don't have an image in the Home Header. Your PR shows the `mdi-disc` icon instead.

It has always showed `mdi-disc` in the home header for me, since the home header lists albums, not persons.

ThibaultNocchi (approved):
LGTM

client/components/Layout/Images/BlurhashImage.vue
```diff
@@ -1,19 +1,19 @@
...
-      <blurhash-canvas
-        v-if="hash"
-        key="canvas"
-        :hash="hash"
-        :width="width"
-        :height="height"
-        :punch="punch"
-        class="absolute"
-      />
+  <div ref="imageElement">
```
ferferga, line 2: Shouldn't we use better the ``img`` element to pass it to the function?

client/components/Layout/Images/BlurhashImage.vue
```diff
@@ -1,19 +1,19 @@
...
+      <v-fade-transition>
+        <blurhash-canvas
+          v-if="hash"
+          :hash="hash"
+          :width="width"
+          :height="height"
+          :punch="punch"
+          class="absolute"
+        />
+      </v-fade-transition>
```
ferferga, line 13: We don't need a v-fade transition here. On fast connections the images load so fast that the transition is longer than the time that it takes to load the connection, so we want to show the blurhash as soon as possible without any fancy thing. The image is another thing though.

With your branch, the transition when switching tracks is very "abrupt" if that makes sense.

client/components/Layout/Images/BlurhashImage.vue
```diff
@@ -23,17 +23,11 @@
...
-        color="white"
-        dark
-      >
+    <slot v-else name="placeholder">
+      <v-icon class="absolute text--disabled" :size="iconSize">
         {{ getItemIcon(item) }}
       </v-icon>
-    </div>
-    <slot v-else name="placeholder" />
+    </slot>
```
ferferga, line 30: What if somewhere we don't want an icon placeholder but something else instead?

ferferga (changes requested):
I confirm that there aren't issues while switching between tracks in audiocontrols (with a caveat, see review comments).

The issues with people (only tested music artists) I commented in my other review persist though:

**Master**
image

**This branch**
image

*(We expect album covers now or artist images?)*

Also, in some places artists has images and in some other they don't:
image

However, in other item "You may also like":
image

There are some other cases where the image is present in the library cards but not in the artist page:

image
image

**I found this happens when you're navigating from a place where the image doesn't exist as well**. So, if I click in the library card, I can see the image. However, if I find that artist through "You may also like", there are no images. Same applies with home header and I guess to every place where the image isn't displayed correctly. **Following links in albums to artist also reproduce this problem**

client/components/Item/PersonList.vue
```diff
@@ -8,12 +8,13 @@
...
-            v-if="item.PrimaryImageTag"
-            ref="personImg"
-            :src="getImageUrl(item.Id)"
-          />
-          <v-icon v-else class="grey darken-3">mdi-account</v-icon>
+          <v-avatar color="card">
+            <blurhash-image :item="item">
+              <template #placeholder>
+                <v-icon size="16">mdi-account</v-icon>
+              </template>
```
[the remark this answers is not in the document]
MrTimscampi: Because this doesn't use BaseItemDto, but BaseItemPerson.

client/components/Layout/Images/BlurhashImage.vue
```diff
@@ -122,16 +117,25 @@ export default Vue.extend({
...
+          preferBackdrop: this.type === ImageType.Backdrop,
+          width: element?.clientWidth,
+          ratio: window.devicePixelRatio || 1
+        });
 
-      if (!this.image) {
-        this.onError();
-      }
+        this.image = imageInfo.url;
+        this.tag = imageInfo.tag ? imageInfo.tag : 'no-image-tag';
```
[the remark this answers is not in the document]
MrTimscampi: Holdover from jf-web, thanks for catching it :yum:

client/components/Players/PlayerManager.vue
```diff
@@ -523,50 +522,44 @@ export default Vue.extend({
           artwork: [
             {
               src:
-                this.getImageUrlForElement(ImageType.Primary, {
-                  item: this.getCurrentItem,
-                  maxWidth: 96
-                }) || '',
+                this.getImageInfo(this.getCurrentItem, {
+                  width: 96
```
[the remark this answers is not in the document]
MrTimscampi: The server won't cut them if we specify the height too, as far as I'm aware. It will just try to make the image fit in the box you send it.

client/components/Item/PersonList.vue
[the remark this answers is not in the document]
ferferga: @MrTimscampi Shouldn't we then instruct BlurhashImage to handle those? Right now they don't have a blurhash as well because this. As you moved everything to a single function in imageHelper I guess this should be possible without much hassle.

I'd rather we get it merged without this as the placeholder mess is very difficult to maintain and having everything centered (like image rescaling on resize, the fallbacks and so on) in blurhashimage turned out to be great in the long term (this PR would have been massive for instance :P)

client/components/Layout/Images/BlurhashImage.vue
ferferga: After testing what I comment on https://github.com/jellyfin/jellyfin-vue/pull/769#discussion_r608576319 both in your branch and in master, I see that the Cast & Crew images have a far superior resolution in your branch than in master, which makes me suspect that I was indeed right about this change.

Maybe, being the child absolute positioned, the size can't be determined correctly and it's loading full resolution?

MrTimscampi: That wasn't working, probably because the image doesn't contain anything and is absolute.

This element has the same size anyway, and is technically our image component.

client/components/Layout/Images/BlurhashImage.vue
MrTimscampi: You just use the slot and set whatever you want inside of it.

This is how slots work. What is inside the slot definition is the default. Then you override it by defining a template for that slot when calling the component.

MrTimscampi:
> The issues with people (only tested music artists) I commented in my other review persist though

If fixed this locally, but it's important to note a few things:
* We're listing albums and not artists, so the disc makes more sense
* It's only showing a person because HomeHeaderItem is currently fetching the parent items. With this PR, this isn't necessary at all (as pointed out before, items have all the necessary tags and blurhashes), so the next step after this is completely removing the related items thing, which will lead to having discs there anyway.

> Also, in some places artists has images and in some other they don't

I am not able to reproduce this using a 10.7.1 server and this branch.

> There are some other cases where the image is present in the library cards but not in the artist page

Same thing, I'm not able to reproduce it.

> I found this happens when you're navigating from a place where the image doesn't exist as well. So, if I click in the library card, I can see the image. However, if I find that artist through "You may also like", there are no images. Same applies with home header and I guess to every place where the image isn't displayed correctly. Following links in albums to artist also reproduce this problem

Doing this gives me images all the time:

* Home header => View details => By { artist }
  **Image shows up**
* Library => Click on card => By { artist }
  **Image shows up**
* Library => Click on artist link on card
  **Image shows up**
* On artist page => You may also like
  **Image shows up**

client/mixins/imageHelper.ts
```diff
@@ -264,82 +248,248 @@ const imageHelper = Vue.extend({
...
+    },
     /**
-     * Returns the URL of an item's image:
-     * · When 'element' parameter is passed, size of the image will be determined by the element's width & height
-     * · When 'maxWidth' and 'maxHeight' parameters are passed, size of the image will be as requested
-     * · When no 'element' or 'maxWidth' or 'maxHeight' is provided, image will have the original size.
+     * Generates the image information for a BaseItemDto or a BasePersonDto according to set priorities.
+     *
+     * @param {(BaseItemDto | BaseItemPerson)} item - Item to get image information for
+     * @param {Object} [options] - Options object
```
ferferga, line 276:
```suggestion
     * @param {object} options - Optional parameters for the function.
```

I think the original description was more suitable and it also fixes the lint problem :P.

ferferga (approved):

PR #780, opened 2021-02-22
Improve PWA experience

MrTimscampi:
This does a few things:

* Fixes the backdrop overlay not over-scrolling properly
* Use a transparent system bar on iOS
* Fix the app bar to display properly with the transparent system bar
* Add safe zones to the various player controls

**iOS 14.4 (iPhone 12)**
*Before*
image
*after*
image
image
image

### DIFF
--- a/assets/global.scss
+++ b/assets/global.scss
@@ -97,6 +97,27 @@ body {
   flex: 0;
 }
 
+.pa-s {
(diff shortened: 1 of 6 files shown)


ThibaultNocchi:
Maybe we can split out the safe zone classes to our global CSS with 5 classes being like "safezone-bottom/top/left/right" and "safezone-all"?

ferferga (approved):

PR #781, opened 2021-02-22
Server card infos & UX

MrTimscampi:
* Adds a tooltip on the info button with the ProductName, Version and OperatingSystem
* [WIP] Add a confirmation dialog when removing a server


ferferga:
@jellyfin rebase

PR #800, opened 2021-02-26
ci(github): fix fetch-depth for sonarcloud test results

MrTimscampi:


.github/workflows/main.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@master
         with:
-          fetch-depth: 1
+          fetch-depth: 0
```
ferferga, line 57: It's not strictly necessary right now, but maybe we should set this to 0 in all the workflows so this is consistent and reproducible across all the workflows?

MrTimscampi: Apparently, the default is 1 (So you line wasn't needed), but Sonarcloud has an auto-issue creator which looks at blames to assign people automatically.

Setting the fetch-depth to 0 removes this, which gets rid of the issue. I don't think it's necessary to set this to something other than 1, unless explicitly needed, but we should probably get rid of the lines that just repeat the default.

ferferga: @MrTimscampi Now that you're mentioning this it's true about the default thing. I just copy-pasted this from my README profile generator repo (which rebases and ammends the commits instead) and that one needeed full history ^^'.

ferferga (approved):

ferferga (approved):

ThibaultNocchi (approved):

PR #805, opened 2021-02-27
chore: replace nuxt-ts with nuxt

ferferga:
Nuxt now is built on TypeScript so nuxt-ts is no longer necessary

This removes the warnings at build time.


MrTimscampi (changes requested):
We can probably remove the dependency itself, too.

MrTimscampi (approved):
Nevermind, we don't depend on it directly, it's a sub-dependency.

MrTimscampi (changes requested):
Third time's the charm, we can actually remove `@nuxt/typescript-runtime`

ThibaultNocchi (approved):

PR #815, opened 2021-03-01
chore: switch to npm

ferferga:
*This needs further discussion across all the client devs, as this has been discussed more thoroughly between the Vue contributors, but not as much with the rest of the client developers and, ideally, we should use the same package manager everywhere*
Major reasons for switching that were pointed:

* All the yarn's advantages over npm are non-existent today
* Yarn introduced some breaking changes in 2.x that caused controversies and avoid us to upgrade directly, which means we are stuck in a version in maintenance mode. So far, all the npm major updates could be updated like a charm, and it's likely they will commit to it. The same can't be said about yarn.
* Npm comes with the node installation, so it's frictionless for the user
* Npm is more mainstream and all the package maintainers add instructions with how to use it to install their packages. The same doesn't apply to yarn. Although adapting the commands to yarn is trivial, for new contributors it might cause some friction.

## Extra

* Added ``prod`` and ``prod:static`` commands as an shortcut for ``build && start`` and ``build:static && start:static`` respectively.


.github/workflows/build.yml
```diff
@@ -20,24 +20,24 @@ jobs:
         with:
           node-version: 14
 
-      - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Get npm config get cacheectory path
```
MrTimscampi, line 23:
```suggestion
      - name: Get npm cache directory path
```

.github/workflows/build.yml
```diff
@@ -52,21 +52,21 @@ jobs:
...
```
MrTimscampi, line 55:
```suggestion
      - name: Get npm cache directory path
```

.github/workflows/docker.yml
```diff
@@ -26,24 +26,24 @@ jobs:
         with:
           node-version: 14
 
-      - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Get npm config get cacheectory path
```
MrTimscampi, line 29:
```suggestion
      - name: Get npm cache directory path
```

.github/workflows/docker.yml
```diff
@@ -89,24 +89,24 @@ jobs:
...
```
MrTimscampi, line 92:
```suggestion
      - name: Get npm cache directory path
```

.github/workflows/main.yml
```diff
@@ -23,24 +23,24 @@ jobs:
         with:
           node-version: 14
 
-      - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Get npm config get cacheectory path
```
MrTimscampi, line 26:
```suggestion
      - name: Get npm cache directory path
```

.github/workflows/main.yml
```diff
@@ -57,24 +57,24 @@ jobs:
...
```
MrTimscampi, line 60:
```suggestion
      - name: Get npm cache directory path
```

.husky/commit-msg
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname $0)/_/husky.sh"
 
-yarn commitlint --edit
+npm commitlint --edit
```
MrTimscampi, line 4:
```suggestion
npx commitlint --edit
```

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn lint-staged
+npm lint-staged
```
MrTimscampi, line 4:
```suggestion
npx lint-staged
```

README.md
```diff
@@ -41,7 +41,7 @@ This is an experimental web client for Jellyfin based on Vue.js. We welcome all
 
 ## Requirements
 
-This project uses [Yarn](https://yarnpkg.com/getting-started/install) as a package manager.
+This project uses [npm](https://docs.npmjs.com/getting-started) as a package manager.
```
MrTimscampi, line 44: We can probably just change this to mention we require Node 14, since NPM comes standard with it.

Dockerfile
```diff
@@ -28,8 +28,8 @@ COPY --from=build /app/.nuxt ./.nuxt
 COPY --from=build /app/static ./static
 
 # Install runtime dependencies
-RUN yarn install --production --no-lockfile
+RUN npm install --production --no-save --no-audit
```
ThibaultNocchi, line 31: Why not go `npm ci` in that case? :)

.github/workflows/Dockerfile.ssr
```diff
@@ -11,8 +11,8 @@ COPY .nuxt ./.nuxt
 COPY static ./static
 
 # Install runtime dependencies
-RUN yarn install --production --no-lockfile
+RUN npm install --production --no-save --no-audit
```
ThibaultNocchi, line 14: Why not go with `npm ci` in that case? :)

Dockerfile
ferferga: ``npm ci`` fetches packages from the lockfile (``pacakge-lock.json``). Here, we're installing the runtime dependencies for SSR, so there's no lockfile, we're doing a fresh installation from ``package.json``

.github/workflows/Dockerfile.ssr
ferferga: Same as above

ThibaultNocchi (approved):

ferferga:
@ThibaultNocchi After merging #861, it seems we need a package-lock file here so dependabot catches updates for the dependencies properly, so we're going to use ``npm ci`` in the end :)

ThibaultNocchi (approved):

ThibaultNocchi (approved):

.github/workflows/build.yml
```diff
@@ -51,22 +52,23 @@ jobs:
         uses: actions/setup-node@v2.1.5
         with:
           node-version: 14
+          check-latest: true
```
MrTimscampi, line 55: From the docs:

> Setting check-latest to true has performance implications as downloading versions of Node is slower than using cached versions

They also mention that for maximum stability, it should be left to the default.

It's likely best to keep this to the default, since we don't need the absolute latest version of Node to run these. I'd rather have the workflows run as fast as possible so we don't have to wait too much when rebasing or something.

MrTimscampi (changes requested):
One small thing about check-latest, otherwise this is good to go.

MrTimscampi (approved):

Issue #801, opened 2021-02-26
Refactor the way items are handled in Vuex stores


heyhippari:
Multiple design issues plague our current Vuex stores.  
Among them is the way we store items coming from the API ("items" here refers to BaseItemDto, UserDto and PublicSystemInfo, the basic item types of the API)

We currently store multiple copies of these in various stores, which makes no sense (It creates more noise, makes store testing more complex and makes updates through the Websocket completely nonsensical, since they have to be done everywhere...)

## Proposal

The proposed schema (containing only the relevant fields for each store) is as follows:
![image](https://user-images.githubusercontent.com/19396809/109303831-a0233e80-783b-11eb-96d7-6533a625d20e.png)

Some corrections apply to this schema, however:

* In the `Users` store, `publicUsers` should be of the type `string[]`
* ~~The `HomeSections` store should be renamed to `HomeScreen`, and~~ In the `HomeSections` store, `sections` should be of the type `string[]` and another property `sectionContent` of type `Record<number, string[]>` should be added. An `enableHeader` flag should also be part of the store.

### Core stores

The `Servers`, `Users` and `Items` stores are the central stores of the client. They store "raw" objects from the API in their respective types: `ServerInfo`, `UserDto` and `BaseItemDto` (Note: `ServerInfo` is built in the Vue client, but it acts as a core data type here).

These stores get referenced from other stores as the main source of data.

Related:
- #764 

### Session

This store is a combination of a few of our current stores: `userViews`, `user` and part of `server`. It's goal is, as the name implies, to represent the current user's session, with everything needed for it (It's sort of a "fit-all" store for things that don't really need their own stores)

It currently contains:

* `currentServer` (`string`) - Contains the ID of the currently connected server
* `accessToken` (`string`) - Contains the "raw" access token for the API
* `userViews` (`string`) - Contains the ID of the BaseItemDto representing the user's libraries

Both `currentServer` and `userViews` are primarily accessed through getters which consolidate the list of IDs with the contents of the respective stores.

### HomeSection

In order to be more modular, this store is completely refactored.

It now features a list of strings (`sections`), representing the various section types that make up the user's home screen.

Another property, `sectionContent`, contain the index of the section, as well as a list of IDs for the items in that section.

~~Finally, an `enableHeader` boolean sets the visibility of the HomeHeader component.~~ An `enableHeader` getter checks if the first index of `sections` is the HomeHeader.

### ShowSeasons

[...]

PR #821, opened 2021-03-02
perf(memory): unsubscribe from mutation watchers on destroy

ferferga:
Unsubscribes the stores of PlaybackManager, AudioPlayer and VideoPlayer, in an attempt to fix the memory leak the client is facing right now. The rest of the components that contain store watchers won't need those as soon as #801 and #764 are done, so I didn't change those.

Although normally the components where I added unsubscribe are not destroyed and recreated (like cards, which are the problematic ones), they are recreated on logout and a relogin, causing the memory leak to appear, which will be a problem for an user that swaps between accounts frequently.


ThibaultNocchi (approved):
Tested locally :+1:

MrTimscampi (approved):

PR #833, opened 2021-03-04
fix(snackbar): snackbar is now passed the locale key instead of the a…

### LINKED ISSUE #765: Snackbar message passing
In multiple places in the app, we use a snackbar associated to its store to display messages.

A long time ago, we used to directly pass an arbitrary string as the message. But with localization, it made more sense to pass the localization key and let the component display the localized version.

But a few instances of arbitrary or already localized strings passed to the snackbar still exist, like [here](https://github.com/jellyfin/jellyfin-vue/blob/master/components/Buttons/FilterButton.vue#L322)

Those strings should be fixed by only giving the locales keys, such as `filtersNotFound` in that case

ThibaultNocchi:
…rbitrary string

Snackbar's dispatch was called with already localized strings, so it kept them as is. Now the key is
passed and the snackbar comp does the localization with the given key.

fix #765

### DIFF
--- a/components/Buttons/FilterButton.vue
+++ b/components/Buttons/FilterButton.vue
@@ -319,7 +319,7 @@ export default Vue.extend({
         }
       } catch (error) {
         this.pushSnackbarMessage({
-          message: this.$t('filtersNotFound'),
+          message: 'filtersNotFound',
           color: 'error'
         });
       }
--- a/components/Buttons/LikeButton.vue
+++ b/components/Buttons/LikeButton.vue
@@ -75,7 +75,7 @@ export default Vue.extend({
         }
       } catch {
         this.pushSnackbarMessage({
-          message: this.$t('unableToToggleLike'),
+          message: 'unableToToggleLike',
           color: 'error'
         });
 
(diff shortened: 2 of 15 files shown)


ferferga (approved):

MrTimscampi (changes requested):
I think it makes more sense the other way around.

That allows us to keep support for extensions like i18n-ally, to show us if a key exists or not easily, and allows us to send exception messages directly in extreme situations.

ThibaultNocchi:
In that case, maybe I should pass to the lambda a function that returns a string, and the snackbar calls it? That way we keep i18n ally and the arbitrary string passing?

MrTimscampi:
> In that case, maybe I should pass to the lambda a function that returns a string, and the snackbar calls it? That way we keep i18n ally and the arbitrary string passing?

Why not simply have the snackbar component expect a pre-translated string?

If something isn't translated, we'll see it fairly easily anyway, since we'll have the key show up (This mirrors how jf-web handles things, usually)

ThibaultNocchi:
Cause in the edge case where you change language mid-snackbar, well it won't be translated. But initially I think it was done because the snackbar component uses `$t(msg)`, a change you've introduced here ;p https://github.com/jellyfin/jellyfin-vue/commit/7585079a1479256916c2d58b0dfb1eee7b75d7bf#diff-76b59f82295864ac40ecb7a034111d0e1a8245ce4815660434d69a0898e6aba2

But yeah, we shouldn't bother. So the PR should be reverted: directly use the message in the component and check for snackbar calls with the localizaton key :p

PR #874, opened 2021-03-13
Improve page navigation and refactor home screen

MrTimscampi:
* Adds `keep-alive` prop to the Nuxt root component, in order to keep some pages in memory, which improved the navigation experience.
  This should eventually be coupled with using `fetch()` for all the data fetching around the app, and the use of the `activated` hook with a timestamp check, in order to refresh stale data upon navigation after a certain time (likely after 30s).

https://user-images.githubusercontent.com/19396809/111021546-720e3480-83cd-11eb-8464-4d342838da74.mp4

* Move the data fetching logic of the home page to the `fetch()` hook.

### DIFF
--- a/store/homeSection.ts
+++ b/store/homeSection.ts
@@ -10,7 +10,7 @@ import { AppState } from './index';
 
 export interface HomeSection {
   name: string;
-  libraryName: string | undefined;
+  libraryName?: string | null | undefined;
   libraryId: string;
   shape: string;
   type: string;
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -98,7 +98,7 @@
     </v-app-bar>
     <v-main>
       <div class="pa-s">
-        <nuxt />
+        <nuxt keep-alive :keep-alive-props="{ max: 10 }" />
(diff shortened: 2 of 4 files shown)


MrTimscampi:
This will need more refactoring (Redoing the store completely, using different components for different sections, etc) but these can already be merged as-is, both to already benefit users and to allow more work to be done (moving the other pages to the `fetch()` hook).

store/homeSection.ts
```diff
@@ -10,7 +10,7 @@ import { AppState } from './index';
 
 export interface HomeSection {
   name: string;
-  libraryName: string | undefined;
+  libraryName?: string | null | undefined;
```
ThibaultNocchi, line 13: What does `?` add more than the union?

MrTimscampi: In an interface, it marks the property as optional.

ThibaultNocchi:
@jellyfin rebase

ferferga (approved):

ThibaultNocchi (approved):

PR #875, opened 2021-03-13
ci: use the jellyfin-bot token for Github Actions

MrTimscampi:
A few of the actions (labeler and the merge conflicts one) show up as `github-actions` while they should be `jellyfin-bot` for consistency.

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,4 +1,4 @@
-name: "Label pull requests"
+name: 'Label pull requests'
 on:
   - pull_request_target
 
@@ -9,4 +9,4 @@ jobs:
       - uses: actions/labeler@main
         continue-on-error: true
         with:
-          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          repo-token: '${{ secrets.GH_TOKEN }}'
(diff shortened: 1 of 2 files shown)


.github/workflows/merge-conflicts.yml
```diff
@@ -11,5 +11,5 @@ jobs:
     steps:
       - uses: mschilde/auto-label-merge-conflicts@master
```
ferferga, line 12: We should use the fixed version for this

.github/workflows/label.yml
```diff
@@ -9,4 +9,4 @@ jobs:
       - uses: actions/labeler@main
```
ferferga, line 9: Same

ferferga (changes requested):

.github/workflows/merge-conflicts.yml
h1dden-da3m0n: I second this (as it is best practice) + your dependabot config will already auto update the actions when there are new versions

h1dden-da3m0n (approved):

ferferga (approved):

ThibaultNocchi (approved):

PR #877, opened 2021-03-13
feat(pages): validate item IDs when navigating to the pages

MrTimscampi:
Our item IDs are MD5 hashes in v10, so this isn't doing proper UUID checks but a basic MD5 validity
check.

### DIFF
--- a/pages/artist/_itemId/index.vue
+++ b/pages/artist/_itemId/index.vue
@@ -114,13 +114,18 @@
 import Vue from 'vue';
 import { mapActions } from 'vuex';
 import { BaseItemDto, ImageType } from '@jellyfin/client-axios';
+import { Context } from '@nuxt/types';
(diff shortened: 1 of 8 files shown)


ferferga:
Wouldn't this be better in a middleware? This seems a bit hacky

MrTimscampi:
I don't get how this would be hacky since the validate hook is made specifically for that?

ferferga:
You're indeed right, thought you implemented this from scratch and it wasn't a nuxt hook.

Anyway, what's the difference with a middleware?

MrTimscampi:
> Anyway, what's the difference with a middleware?

Middlewares are very generic and meant more as "navigation guards" than validation. For example, blocking routes per user roles, which is something generic that doesn't really depend on anything in the route itself.

ferferga (approved):

ThibaultNocchi (approved):

PR #883, opened 2021-03-15
build(docker): restore dockerignore

MrTimscampi:
The Dockerfile was wrongly removed in #767 .

This makes building the docker images locally unnecessarily long (due to copying node_modules then removing it by running `npm ci`) and copies unnecessary stuff on the CI.

This restores the original file, and cleans it up a bit.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,37 @@
+# Dockerfile
+Dockerfile
+Dockerfile.static
+
+# Coverage directory used by tools like istanbul
+coverage
+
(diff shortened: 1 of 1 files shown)


ferferga (approved):
Wasn't wrongly removed in #767 because in yarn, the ``-frozen-lockfile`` argument didn't remove the ``node_modules`` directory, so we could still reuse it in the Docker context. It should've been added in the #815 though :)

ThibaultNocchi (approved):

PR #884, opened 2021-03-15
fix(docker): fix dockerfile not building the SSR version properly

MrTimscampi:
For some reason, npm wasn't passing down the standalone flag to nuxt-build. As a result, dependencies weren't bundled anymore, leading to errors when trying to run it.

This adds a dedicated standalone build command to avoid having to pass down flags.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -37,7 +37,7 @@ jobs:
         run: npm ci --no-audit
 
       - name: Build client
-        run: npm run build --production --standalone
+        run: npm run build:standalone
 
       - name: Configure QEMU
         uses: docker/setup-qemu-action@v1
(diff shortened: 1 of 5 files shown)


ThibaultNocchi (approved):
LGTM

.github/workflows/docker.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: npm ci --no-audit
 
       - name: Build client
-        run: npm run build --production --standalone
+        run: npm run build:standalone
```
ferferga, line 47:
```suggestion
        run: "npm run build:standalone"
```

.github/workflows/docker-release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         run: npm ci --no-audit
 
       - name: Build client
-        run: npm run build --production --standalone
+        run: npm run build:standalone
```
ferferga, line 40:
```suggestion
        run: "npm run build:standalone"
```

ferferga (changes requested):
This won't work unless you quote it.

ferferga (approved):

Issue #836, opened 2021-03-05
[tweak] Title and Original Title are both displayed when values are the same


neopc10:
![image](https://user-images.githubusercontent.com/1365322/110048345-b2aae580-7d1d-11eb-9ebc-5be94ab9c7fc.png)

A check will be useful on item details pages (movies, TV series) that determines if the title and original title have the same value--if so, then hide the original title.

neopc10:
This appears to still be an issue with TV Shows.

PR #890, opened 2021-03-17
check if OriginalTitle and Name values are same

jakobkukla:
Don't display OriginalTitle if Name has same value. fix https://github.com/jellyfin/jellyfin-vue/issues/836

### DIFF
--- a/pages/item/_itemId/index.vue
+++ b/pages/item/_itemId/index.vue
@@ -13,7 +13,7 @@
             {{ item.Name }}
           </h1>
           <h2
-            v-if="item.OriginalTitle"
+            v-if="item.OriginalTitle && item.OriginalTitle !== item.Name"
             class="text-subtitle-1"
             :class="{ 'text-center': !$vuetify.breakpoint.mdAndUp }"
           >


pages/item/_itemId/index.vue
```diff
@@ -13,7 +13,7 @@
             {{ item.Name }}
           </h1>
           <h2
-            v-if="item.OriginalTitle"
+            v-if="item.OriginalTitle != item.Name"
```
ferferga, line 16:
```suggestion
            v-if="item.OriginalTitle !== item.Name"
```

ferferga (changes requested):
Unrelated to why I'm requesting changes, but imo, we should trim and lowercase the strings first as well to avoid any trailing space, capitalization differences or some other quirks that providers might have.

Thoughts? @MrTimscampi @ThibaultNocchi @Camc314 

PS: Also, if you want your contribution to be parsed successfully in GitHub, you need to set the same email your use in GitHub in your local git environment, rebase and force push. You can ping me in Matrix if you need help with this.

ThibaultNocchi:
I agree to @ferferga's comment, it could be a nice QOL addition to trim and lowercase first the strings :)

Also, you should add `fix #836` in your PR body, that way GH will automatically close the issue when the PR is merged ;)

pages/item/_itemId/index.vue
```diff
@@ -13,7 +13,7 @@
             {{ item.Name }}
           </h1>
           <h2
-            v-if="item.OriginalTitle"
+            v-if="item.OriginalTitle !== item.Name"
```
MrTimscampi, line 16: This changes the way the condition works.

Example. Before:
```
OriginalTitle = undefined => not shown
OriginalTitle = null => not shown
OriginalTitle = "Lorem Ipsum" => Shown
```

Now:
```
OriginalTitle = null && Name = "Lorem Ipsum" => Shown
OriginalTitle = undefined && Name = "Lorem Ipsum" => Shown
```

The condition should actually be:
```
item.OriginalTitle && item.OriginalTitle !== item.Name
```

MrTimscampi (changes requested):
I also don't really agree with the string cleanup that's been proposed above.

It's not the client's job to work around provider weirdness. That's done on the server. We shouldn't do any trimming or anything on our side. If there are discrepancies, it's the server's job or the provider's job to fix that.

ferferga:
@MrTimscampi I don't think so, because they're different fields and server doesn't care how we display the information, only about returning the needed data, and this is all about how we display the information.

pages/item/_itemId/index.vue
jakobkukla: Thanks, I was thinking about that myself, but forgot later haha :)

MrTimscampi:
> @MrTimscampi I don't agree, because they're different fields and server doesn't care how we display the information, only about returning the needed data, and this is all about how we display the information.
> 
> We can skip all the lowercase and trim altogether, but I'm against delegating that to server.

The trim is something the server should do. Its job is also to sanitize data for us, which means trimming and cleaning up the data it sends to clients.

As for doing lowercase, jf-web doesn't do it and it hasn't caused any issues up to now, so I'm not really seeing much of a case for it. See [here for how it's currently done over there](https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/itemDetails/index.js#L519-L527).

MrTimscampi (approved):
LGTM, thanks :)

ThibaultNocchi:
If you could squash your commits that'd be great please. If you don't know how to do, just ask us ;)

jakobkukla:
@ThibaultNocchi Sure can do! Thanks for all the help getting into this stuff :)

ThibaultNocchi (approved):

ThibaultNocchi:
Np, we may be pretty picky even for a line change haha

ferferga (approved):

Issue #835, opened 2021-03-04
Can't click on the artists on the music library page


ThibaultNocchi:
**Describe The Bug**

On the music library page, it'd be great to be able to click on the artist to directly be redirected to their page.

![print_screen_2021-03-04-19-51-36](https://user-images.githubusercontent.com/1619359/110014408-0e2ba200-7d23-11eb-9a06-0a930ec0d98d.png)

Here I'd love to click on "C2C" to go to the group page, instead of currently being redirected to the "Tetra" page.

neopc10:
Tangentially related to this, I'd like to be able to click the artist name/logo on the home screen "recently added" banner and be redirected to their page.

ThibaultNocchi:
Well, it's using the same component so it should work everywhere :p

MrTimscampi:
> Well, it's using the same component so it should work everywhere :p

The artist name/logo on the home header is a completely different component, though :yum:

PR #897, opened 2021-03-19
feat(clickable-artist): add link to card-subtitle

jakobkukla:
Make the artist name in the item subtitle clickable. fix https://github.com/jellyfin/jellyfin-vue/issues/835. 

I'm not sure if this is the most elegant solution though. The display: block is necessary, since the nuxt-link element sets it to inline and it gets shifted ever so slightly down.

### DIFF
--- a/components/Item/Card/Card.vue
+++ b/components/Item/Card/Card.vue
@@ -71,7 +71,15 @@
       </div>
       <div v-if="text" class="card-text">
         <div class="card-title mt-1 text-truncate">{{ cardTitle }}</div>
-        <div class="card-subtitle text--secondary text-truncate">
+        <nuxt-link
+          v-if="item.Type === 'MusicAlbum'"
+          tag="div"
+          class="card-subtitle text--secondary text-truncate link"
+          :to="getItemDetailsLink(item.AlbumArtists[0], 'MusicArtist')"
+        >
+          {{ cardSubtitle }}
+        </nuxt-link>
+        <div v-else class="card-subtitle text--secondary text-truncate">
           {{ cardSubtitle }}
         </div>
       </div>


components/Item/Card/Card.vue
```diff
@@ -71,7 +71,14 @@
...
       <div v-if="text" class="card-text">
         <div class="card-title mt-1 text-truncate">{{ cardTitle }}</div>
-        <div class="card-subtitle text--secondary text-truncate">
+        <nuxt-link
+          v-if="item.Type === 'MusicAlbum'"
+          class="card-subtitle text--secondary text-truncate link card-subtitle-link"
+          :to="getItemDetailsLink(item.AlbumArtists[0], 'MusicArtist')"
+        >
+          {{ cardSubtitle }}
+        </nuxt-link>
```
MrTimscampi, line 80:
```suggestion
        <nuxt-link
          v-if="item.Type === 'MusicAlbum'"
          tag="div"
          class="card-subtitle text--secondary text-truncate link"
          :to="getItemDetailsLink(item.AlbumArtists[0], 'MusicArtist')"
        >
          {{ cardSubtitle }}
        </nuxt-link>
```

Maybe this would fix it? We technically lose the link display on the bottom left, but I'm not really sure that matters.

Otherwise, no need to make a custom class, you can use [Vuetify's display utility classes](https://vuetifyjs.com/en/styles/display/). In this case, `d-block` should do the trick :wink: 

I'm also wondering if this is really usable on mobile :thinking: I can test it later today with my fat fingers to see if it's usable or not. If not, it's pretty easy to disable it on mobile.

ThibaultNocchi (approved):

MrTimscampi (approved):
LGTM

PR #905, opened 2021-03-23
ci(stalebot): add stalebot config

MrTimscampi:
Adds stalebot config to this repository.

The 90 days limit was moved down to 60 days, since 90 days feels way too long for this repository (Deep structural changes happen pretty often still, so 3 months basically means a PR usually has to be rewritten. 

I feel like 60 days feels a bit long too (I'd go with 30), but it's probably a nice middle ground for now.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - regression
+  - confirmed
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  Issues go stale after 60 days of inactivity. Mark the issue as fresh by adding a comment or commit. Stale issues close after an additional 14 days of inactivity.
+  If this issue is safe to close now please do so.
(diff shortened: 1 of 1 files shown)


ThibaultNocchi (approved):
Yeah, though for now it's not like we're not keeping track of what's around so it works

MrTimscampi:
> Yeah, though for now it's not like we're not keeping track of what's around so it works

It avoids us having to check old seemingly abandoned PRs by external contributors :)

We'll just let them be, like we do on other repositories, and let stalebot handle things for us. If they get closed, then it means we can go back in ourselves and finish them up, if needed/wanted.

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - regression
+  - confirmed
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
```
ferferga, line 12: We should maybe mention somehow that labelling as confirmed stops stale bot for that specific PR/issue?

MrTimscampi: I think we're the only ones who can label stuff, if I'm not mistaken, so that's not really a concern for the people who opened the issue or PR.

ferferga (approved):

MrTimscampi:
@jellyfin rebase

PR #911, opened 2021-03-24
Cleanup playback profiles

MrTimscampi:
* Adds typings for the Tizen Web APIs
* Simplifies the global video bitrate thing (Remove variables that are just aliases of the global video bitrate)
* Remove some casting (`as`) and properly declare variable types instead
* Use `createProfileCondition` to create profile conditions instead of doing it manually (It's already done in most of the profile creation, but some forward ports from jf-web didn't include them).
* Removed a bit of Tizen code that depends on NativeShell, since we don't support that
* Improved the HEVC profiles comments
* Enabled fMP4 by default for Chrome, Firefox, Safari (Desktop and mobile), Edge, WebOS and Tizen

### DIFF
--- a/client/plugins/playbackProfiles/transcodingProfile.ts
+++ b/client/plugins/playbackProfiles/transcodingProfile.ts
@@ -86,6 +86,9 @@ export function getTranscodingProfiles(
     videoTestElement
   );
 
+  console.dir(hlsInFmp4VideoCodecs);
+  console.dir(hlsInFmp4AudioCodecs);
+
   const hlsInTsVideoCodecs = getSupportedTsVideoCodecs(videoTestElement);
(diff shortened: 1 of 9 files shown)


MrTimscampi:
@jellyfin-bot rebase

client/plugins/playbackProfiles/transcodingProfile.ts
```diff
@@ -86,6 +86,9 @@ export function getTranscodingProfiles(
     videoTestElement
   );
 
+  console.dir(hlsInFmp4VideoCodecs);
+  console.dir(hlsInFmp4AudioCodecs);
```
ThibaultNocchi, line 90: I think you forgot them :p

ThibaultNocchi (approved):
LGTM without the console outputs

PR #912, opened 2021-03-24
fix(card artist subtitle): crashed with artist-less albums

ThibaultNocchi:


MrTimscampi (approved):

client/components/Item/Card/Card.vue
```diff
@@ -72,7 +72,7 @@
       <div v-if="text" class="card-text">
         <div class="card-title mt-1 text-truncate">{{ cardTitle }}</div>
         <nuxt-link
-          v-if="item.Type === 'MusicAlbum'"
+          v-if="item.Type === 'MusicAlbum' && item.AlbumArtists.length > 0"
```
ferferga, line 75: Just a fyi the > 0 is not necessary, we use both though :)

ferferga (approved):

client/components/Item/Card/Card.vue
ThibaultNocchi: Noted :)

MrTimscampi:
@jellyfin rebase

PR #913, opened 2021-03-24
Move sources to a dedicated folder

MrTimscampi:
With the client growing and us potentially moving in a direction where we would wrap the client inside of Electron or other frameworks for a more native experience on some platforms, we need the client's source to live in its own folder.

This makes the tree a little bit less intimidating for newcomers, while also ensuring we have a clean base to start adding code for dedicated platforms.


MrTimscampi:
This was a bit of a mess to get working properly, but it should be good to go.

Some things:

* It increases coverage a lot? Not sure why.
* I had to move the Nuxt root into the client folder, since otherwise Jest was having a seizure while trying to resolve imports during coverage and such.
* I split ESLint and Jest configs up, so we have a clean-ish base for everything, with overrides for Nuxt.

client/store/clientSettings.ts
```diff
@@ -1,5 +1,5 @@
 import { MutationTree, ActionTree } from 'vuex';
-import nuxtConfig from '~/nuxt.config';
+import nuxtConfig from '../nuxt.config';
```
ferferga, line 2: Why? It's in the root of client folder, which not refers to ~

tsconfig.json
```diff
@@ -16,17 +16,7 @@
       "~/*": ["./*"],
```
ferferga, line 16: I think we should be able to change the ``target`` and ``module`` options to ESNext just for the tests, that should speed things up as TypeScript won't transpile anything. Worth a try.

Can you also test which of these options are necessary for the global file as well? So it's not a blanket copy of the one in the package.

client/components/Layout/__tests__/SettingsPage.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { mount } from '@vue/test-utils';
-import SettingsPage from '~/components/Layout/SettingsPage.vue';
+import SettingsPage from '../SettingsPage.vue';
```
ThibaultNocchi, line 2:
```suggestion
import SettingsPage from '~/components/Layout/SettingsPage.vue';
```
You can put that back, tested and approved!

client/jest-helpers.d.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { ActionContext } from 'vuex';
-import { RootState } from '~/store';
+import { RootState } from './store';
```
ThibaultNocchi, line 3: This should also work as the previous version

client/mixins/__tests__/timeUtils.spec.ts
```diff
@@ -1,5 +1,5 @@
 import Vue from 'vue';
-import timeUtils from '~/mixins/timeUtils';
+import timeUtils from '../timeUtils';
```
ThibaultNocchi, line 2:
```suggestion
import timeUtils from '~/mixins/timeUtils';
```

client/store/__tests__/items.spec.ts
```diff
@@ -9,8 +9,8 @@ import {
   actions,
   getters
 } from '../items';
```
ThibaultNocchi, line 11:
```suggestion
} from '~/store/items';
```

client/store/__tests__/items.spec.ts
```diff
@@ -9,8 +9,8 @@ import {
...
-import type { ModuleAction } from '~/jest-helpers.d';
-import { RootState } from '~/store';
+import type { ModuleAction } from '../../jest-helpers.d';
+import { RootState } from '../../store';
```
ThibaultNocchi, line 13: You can put them back

client/components/Layout/__tests__/SettingsPage.spec.ts
ferferga: I think he did the change not because it didn't work, but because it's more "correct". 

The tests should always be alongside the components that are being tested, so if at some point in the future we move the component, the test should be alongside it and the import will work regardless.

tsconfig.json
MrTimscampi: > I think we should be able to change the target and module options to ESNext just for the tests

We should keep these the same per sub-project imo, with the root tsconfig only giving a baseline for config files and such, unless the sub-project doesn't need a special configuration.

I also don't think tests run through Babel, since they run directly in Node.

ferferga: You didn't understand my point I think :)

Basically, what I didn't want is to do unnecessary transpilations for linter/jest/nuxt config files. In Chromecast for instance, we do a transpilation for webpack config and another for the whole client. That's what I wanted to avoid here.

However, I checked out the branch locally and I noticed that there are no TS files in the root folder, so, as you mention, this file is just unused (but I like the idea of having it as a template) :)

client/components/Layout/__tests__/SettingsPage.spec.ts
MrTimscampi: > I think he did the change not because it didn't work, but because it's more "correct".

That's in part true :yum: It didn't work, so it was half an attempt to make it work, and also to make it easier to move tests around in the future.

However, I reverted them all back for now, since I think it's better to do that in a followup PR, if we want to do it.

MrTimscampi:
@jellyfin rebase

ferferga (approved):
Tested locally and Vetur works right and everything is looking good.

However, it seems that the i18n extension is not taking the locale files correctly. Non blocking as I don't think this interrupt the development flow as much, but please take it into account to make a follow-up PR.

MrTimscampi:
> However, it seems that the i18n extension is not taking the locale files correctly.

This is actually fixed as of 8dd30c4. I re-generated the paths for locale files in the settings.

ThibaultNocchi (approved):

PR #843, opened 2021-03-06
Refactor PlayButton and remove GlobalPlayButton

MrTimscampi:
GlobalPlayButton essentially duplicated what PlayButton was doing already, while adding shuffle
support.

This removes that component and adds the same functionality to PlayButton, to avoid having
multiple components with the same purpose.

Blocked by #841 since the tests need some changes (and added cases) but that PR changes the spec file a lot.

PR #917, opened 2021-03-25
feat(play-button): bring back the loading states of the play button

ferferga:
This was removed accidentally in #843

### DIFF
--- a/client/components/Buttons/PlayButton.vue
+++ b/client/components/Buttons/PlayButton.vue
@@ -4,13 +4,15 @@
       v-if="canPlay(items[0]) && fab"
       fab
       color="primary"
+      :loading="loading"
       @click.prevent="playOrResume"
     >
       <v-icon size="36">mdi-play</v-icon>
     </v-btn>
     <v-btn
-      v-if="!fab"
+      v-else-if="!fab"
       :disabled="!canPlay(items[0])"
+      :loading="loading"
       class="mr-2"
       color="primary"
       min-width="8em"
@@ -32,9 +34,10 @@
 <script lang="ts">
 import { BaseItemDto } from '@jellyfin/client-axios';
 import Vue from 'vue';
-import { mapActions } from 'vuex';
+import { mapActions, mapState } from 'vuex';
 import itemHelper from '~/mixins/itemHelper';
 import timeUtils from '~/mixins/timeUtils';
+import { PlaybackStatus } from '~/store/playbackManager';
 
 export default Vue.extend({
   mixins: [itemHelper, timeUtils],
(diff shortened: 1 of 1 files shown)


components/Buttons/PlayButton.vue
```diff
@@ -43,32 +47,57 @@ export default Vue.extend({
...
+    };
+  },
+  computed: {
+    ...mapState('playbackManager', ['status'])
+  },
+  watch: {
+    status(): void {
+      if (this.status === PlaybackStatus.playing) {
+        this.loading = false;
+      }
```
MrTimscampi, line 74: I don't see why this is there, as it doesn't make sense (hence why it was originally removed). Why would you disable the loading state when something is playing?

Plus, as-is, it's broken. Example: I'm loading a library page, with music in the queue, but paused. During loading, I press play. This watcher now triggers loading to be false, while the library still isn't loaded.

components/Buttons/PlayButton.vue
```diff
@@ -43,32 +47,57 @@ export default Vue.extend({
...
-        // We force playback from the start when shuffling, since you wouldn't resume AND shuffle at the same time
-        this.play({
-          items: this.items,
-          startShuffled: true
-        });
-      } else {
-        this.play({
-          items: this.items
-        });
+      if (!this.disabled) {
```
MrTimscampi, line 80: This isn't necessary. If the buttons are disabled, the click event is never sent (Tested on master).

components/Buttons/PlayButton.vue
```diff
@@ -43,32 +47,57 @@ export default Vue.extend({
...
     fab: {
       type: Boolean
     },
     shuffle: {
       type: Boolean
     }
   },
+  data() {
+    return {
+      loading: false
```
MrTimscampi, line 64: I think this is a bit superfluous (Hence why it was removed)

When loading a page, the button should simply be disabled (Which is already the case in master, since there's no item to play). There's no need to have two indications (the loading and the disabled state), one is enough and being disabled makes the most sense.

components/Buttons/PlayButton.vue
```diff
@@ -43,32 +47,57 @@ export default Vue.extend({
       type: Array as () => BaseItemDto[],
       required: true
     },
+    disabled: {
```
MrTimscampi, line 50: I don't see a case where we would need to disable this manually. It either can play items, or it can't.

Do you have any example of where and why we might use this?

MrTimscampi (changes requested):

components/Buttons/PlayButton.vue
ferferga: On library view while it's loading, like that old global-playback buttons

MrTimscampi: It's already disabled while loading, since we pass it the items directly (in order to only play what's shown, even with filters on). See: https://github.com/jellyfin/jellyfin-vue/blob/master/pages/library/_viewId.vue#L37-L38

components/Buttons/PlayButton.vue
ferferga: No, because this shows the loading circle just on presses. So it doesn't care about the previous stages.

If I enter a library and I'm playing something, the default will still be ``false``. After clicking, PlaybackStatus will be ``Playing``, but that will be skipped as it didn't mutate since the component was initialized. Then, it will go to ``Stopped`` while the queue is being loaded (so, the condition isn't met). Once everything is loaded, playback status will be ``Playing``, so the watcher trigger this time and removes the loading indicator, just like we want.

components/Buttons/PlayButton.vue
ferferga: Read below

ThibaultNocchi (approved):

MrTimscampi (approved):

MrTimscampi:
@jellyfin rebase

PR #920, opened 2021-03-26
Swiper paginators

### LINKED ISSUE #864: Home menu slideshow item buttons should have a bigger hitbox
On the home menu slideshow, the small bars representing the flow of time and different items are really small.
What would be great is to:
1. Increase their hitbox height
2. If possible, make them visually animate and increase to that increased hitbox

ferferga:
* Added pagination indicators on hover in the progress bar (desktop only):

image

Closes #864 

* Hide drawer when entering into the ``mobile`` breakpoint
* Fix margin issues with the swiper.

### DIFF
--- a/client/assets/styles/HomeHeader.scss
+++ b/client/assets/styles/HomeHeader.scss
@@ -38,9 +38,11 @@
   background-size: contain;
   background-repeat: no-repeat;
   box-sizing: border-box;
-  mask-image: linear-gradient(180deg,
-      rgba(37, 18, 18, 0.75) 0%,
-      rgba(0, 0, 0, 0) 100%);
+  mask-image: linear-gradient(
+    180deg,
+    rgba(37, 18, 18, 0.75) 0%,
+    rgba(0, 0, 0, 0) 100%
+  );
   z-index: 1;
 }
 
@@ -61,7 +63,8 @@
     margin-right: 0;
     padding-bottom: (9 / 16) * 80%;
     background-position: right center;
-    mask-image: linear-gradient(to right,
+    mask-image: linear-gradient(
+        to right,
         hsla(0, 0%, 0%, 0) 0%,
         hsla(0, 0%, 0%, 0.182) 5.6%,
         hsla(0, 0%, 0%, 0.34) 9.9%,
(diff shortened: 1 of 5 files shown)


MrTimscampi:
I'm not quite fan of the look of this. I don't think the numbers are needed or useful, and the circles makes for a somewhat confusing and awkward hit box.

I would see more something like this (The red boxes being the hit box for every segment):

image

Then we could have the growing animation from this PR on individual segment (vertically centered, currently it expands towards the bottom) as a hover indicator, to give feedback, but the hitbox would always remain the same (Currently, it seems like the hit box for the hover to trigger is smaller than the one for it to disappear and smaller than the one to click on segments).

ferferga:
@MrTimscampi I did the expand thing only the first, so bringing that back should be fairly easy 👍.

For reference, which final sizes you tested?

I also vertically centered them with flex, but moving all the progresses bar to center them when a single one was hovered looked really awkward: that's why I went the "hover one, hover all" route. It also forced a reflow.

ThibaultNocchi:
I also like the idea of increasing the height of the bars on hover. Though if you can't increase a specific one I wouldn't bother and just increase them all on hover

MrTimscampi:
> For reference, which final sizes you tested?

This is just a mockup :yum: That looks somewhat finger-sized and is big enough to not be frustrating with a mouse either, but it very likely would need adjustment with a proper implementation.

> Though if you can't increase a specific one I wouldn't bother and just increase them all on hover

Hm, imo then we should avoid increasing the size at all, and rely on another indicator like opacity. Set the opacity to like 0.8 by default, move it to 1.0 on hover with a short transition.

dkanada:
Changing the color is always an option instead of opacity, since that's already used for the swiper progress.

ThibaultNocchi:
I'd reduce the border radius on hover :)

MrTimscampi:
The size of the hit box is great now, imo. Like Thibault said, I'd reduce the border-radius a bit when the segments expand, so they look a bit more like rounded rectangles.

Aside from that though, work well, looks good :+1:

ferferga:
@jellyfin-bot rebase

ThibaultNocchi (approved):

MrTimscampi (approved):

MrTimscampi:
@jellyfin-bot rebase

PR #923, opened 2021-03-26
fix(keepalive): migrated sensible early lifecycle and destroyed metho…

### LINKED ISSUE #922: Homepage backdrop isn't cleared when going to a library
**Steps To Reproduce**

1. Go to the home page and wait for the carousel to pop up
2. Navigate to a library
3. See that the home backdrop is still there

**Expected Behavior**

The background should go back to its default state

ThibaultNocchi:
…ds to activated/deactivated

fix #922

### DIFF
--- a/client/components/Buttons/LikeButton.vue
+++ b/client/components/Buttons/LikeButton.vue
@@ -23,7 +23,9 @@ export default Vue.extend({
   },
   data() {
     return {
-      isFavorite: false
+      isFavorite: false,
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      unsubscribe(): void {}
     };
   },
   watch: {
@@ -34,8 +36,8 @@ export default Vue.extend({
       }
     }
   },
-  mounted() {
-    this.$store.subscribe((mutation) => {
+  activated() {
+    this.unsubscribe = this.$store.subscribe((mutation) => {
       if (
         mutation?.type === 'SOCKET_ONMESSAGE' &&
         mutation?.payload?.MessageType === 'UserDataChanged'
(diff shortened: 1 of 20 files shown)


pages/fullscreen/playback/index.vue
```diff
@@ -50,8 +50,7 @@ export default Vue.extend({
         keyboard: true,
         a11y: true
       } as SwiperOptions,
-      swiper: undefined as Swiper | undefined,
-      previousAppBarOpacity: 0
```
ferferga, line 54: What if I'm accessing this page from an item details page, for instance? (As it has transparency)

ThibaultNocchi: It seems every page that needs it set their transparency at activation, so it the player (or any page) shouldn't have to keep what was previously set imo :)

client/components/Buttons/LikeButton.vue
```diff
@@ -23,7 +23,9 @@ export default Vue.extend({
   },
   data() {
     return {
-      isFavorite: false
+      isFavorite: false,
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      unsubscribe(): void {}
```
ferferga, line 28: We won't need this as soon as the item store is properly implemented, that's why I intentionally skipped them in my PR. Doesn't harm to have this during the transition period.

ferferga (approved):

client/components/Buttons/LikeButton.vue
ThibaultNocchi: Ok that's why :P

MrTimscampi (approved):

MrTimscampi:
@jellyfin rebase

MrTimscampi:
@jellyfin rebase

PR #924, opened 2021-03-26
Add electron wrapper

### LINKED ISSUE #910: Explore the feasability of an Electron version
It should be [fairly easy](https://github.com/michalzaq12/electron-nuxt) to embed Nuxt inside of an Electron app.

This would potentially open the door to adding players with better format support, such as [WebChimera.js](https://github.com/RSATom/WebChimera.js) and distributing a desktop version of the client for some platforms.

In order to be extendable to other "host technologies" (for lack of a better term), this would require the use of a NativeShell-like API (But not NativeShell itself, since it's pretty bad) and the implementation of said API inside of the main client.

MrTimscampi:
**Work in progress**

* Add electron to the project
* Setup Nuxt programmatically and load in Electron window

image

Closes #910


MrTimscampi:
Tentatively closing in favor of #925

PR #925, opened 2021-03-26
Add tauri wrapper

crobibero:
Add [Tauri](https://tauri.studio/) for desktop app.

This uses the native WebView so it's much lighter weight then electron.
This should also allow us to use https://crates.io/crates/libmpv as it's rust based.

image


ThibaultNocchi:
Do you think the dir could be named "tauri" and drop the "src" prefix?

crobibero:
> 
> 
> Do you think the dir could be named "tauri" and drop the "src" prefix?

I just tested this change, looks like the path is hardcoded.

MrTimscampi:
> I just tested this change, looks like the path is hardcoded.

There's apparently these two options, which I think should allow changing it?

```
    --directory, -d   Set target directory for init
    --tauriPath, -t   Path of the Tauri project to use (relative to the cwd)
```

crobibero:
> 
> 
> > I just tested this change, looks like the path is hardcoded.
> 
> There's apparently these two options, which I think should allow changing it?
> 
> ```
>     --directory, -d   Set target directory for init
>     --tauriPath, -t   Path of the Tauri project to use (relative to the cwd)
> ```

I just went looking in the source code and only saw one reference to `tauriPath`, where did you find the arguments?

MrTimscampi:
@crobibero Found here: https://tauri.studio/en/docs/api/cli/#init

MrTimscampi:
Though it seems like it's indeed hardcoded and these parameters do something else. Not ideal, but eh.

MrTimscampi (commented):
I'm getting this on the home page, when connecting to an HTTPS server: 
image

Seems like something's up with the CSP and the Websocker, according to the console? (I am running through Nginx, but don't have any issue with the site when accessing it directly).

Some thoughts:

* I don't seem to have any icons in dev mode on Gnome.
* Building the production version fails with `error: no such subcommand: "tauri-bundler"`. The build script should probably check and install it if building prod.
* I'm not sure how I feel about it using native web views.

For this last one, here's what the cargo package Tauri uses reports, for engines:
image

Some issues with this:
* MSHTML is deprecated (It's Internet Explorer, we don't want that). EdgeHTML is deprecated as well and was replaced by Chromium.
* Gtk-webkit2 is super janky and veeeeery slow. For me, it goes blank whenever I resize the window and it lags a lot.

On one hand, using the native webview for each platform gives us a smaller binary, but on the other hand I feel like it impacts the experience a lot depending on which platform you're on... I can see why Electron uses Chromium and Chromium only: it provides a stable base for every OS, with the exact same features available and the same kind of speed everywhere.

So I'm kind of hesitant on Tauri as a result... :/

crobibero:
> Seems like something's up with the CSP and the Websocker, according to the console? (I am running through Nginx, but don't have any issue with the site when accessing it directly).

I didn't have any issues accessing my production Jellyfin server behind https in both dev & prod builds, so unsure where to look for that.

> Building the production version fails with error: no such subcommand: "tauri-bundler". The build script should probably check and install it if building prod.

I didn't run into this, afaik the Tauri prereqs should have installed this

> MSHTML is deprecated (It's Internet Explorer, we don't want that). EdgeHTML is deprecated as well and was replaced by Chromium.

The webview on windows should be using WebView2, which is available from "Edge Canary".

> Gtk-webkit2 is super janky and veeeeery slow. For me, it goes blank whenever I resize the window and it lags a lot.

@iwalton3 said the same thing when I originally proposed using Tauri, so maybe the tech is too new for actual use and we should just use Electron until a better alternative arises?  


tldr; I agree that Electron is the way to go (for now). I'll keep track of Tauri's progress, hopefully the performance of everything gets better :)

GrabbenD:
Got any updates?

ferferga:
@GrabbenD Every commit pushed to master has Tauri builds as artifacts: https://github.com/jellyfin/jellyfin-vue/actions/workflows/unstable.yml

PR #926, opened 2021-03-28
Plugin cleanup & SSR fixes

MrTimscampi:
* Move browser and device profile detection to Nuxt plugins
  This fixes server issues with SSR and general code design by moving everything about browser and profile detection to Nuxt plugins.  
  It also fixes several issues with SSR in these by removing assumptions about execution context and properly handling cases where we run on the server.
* Fix refreshing on pages with backdrops
  On some pages (like item details), we were setting the backdrop (a Blurhash) in hooks that are executed on the server.
  To solve this, the Backdrop component of the default layout is now set to be rendered only on the client.
* Cleanup item pages
  This also removes the weird lCol/rCol object that was in all the item pages. It's way nicer to do this via non-bound classes (Which also saves on some reactive properties)
* Remove link and overlay on main card in item details
  It doesn't make sense to link to the same page and there's a play button right next to it.


client/components/Item/Card/Card.vue
```diff
@@ -55,7 +55,7 @@
           />
         </div>
         <div
-          v-if="overlay && !$browser.isMobile()"
```
ferferga, line 58: Why? I guess you wanted to switch to vuetify breakpoint but forgot about it

client/plugins/browserDetection.ts
```diff
@@ -277,7 +297,15 @@ class BrowserDetector {
...
-    if (/(?:iPhone|iPad|iPod|Android)/.test(navigator.userAgent)) {
+    let userAgent = '';
+
+    if (process.client && navigator.userAgent) {
+      userAgent = navigator.userAgent;
+    } else if (this.context.req.headers['user-agent']) {
+      userAgent = this.context.req.headers['user-agent'];
+    }
+
+    if (/(?:iPhone|iPad|iPod|Android)/.test(userAgent)) {
```
ferferga, line 308: We can remove this altogether and use Vuetify breakpoint

ferferga (changes requested):
A surface review

client/components/Item/Card/Card.vue
MrTimscampi: It's because using `$browser.isMobile()` here completely breaks refreshing with SSR on some pages.

client/plugins/browserDetection.ts
MrTimscampi: They serve different purposes.

This is for platform detection, in order to know what we're running on, which we need in some cases (Like to fix some codec inconsistencies).

The Vuetify breakpoint only cares about viewport size, which isn't what we want here.

client/plugins/browserDetection.ts
```diff
@@ -50,7 +56,13 @@ class BrowserDetector {
    * @memberof BrowserDetector
    */
   private userAgentContains(key: string): boolean {
-    const userAgent = navigator.userAgent || '';
+    let userAgent = '';
```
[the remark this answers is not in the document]
MrTimscampi: I'm still trying to thing of a good way to have `userAgentContains` work with both strings and Regex.

client/plugins/browserDetection.ts
```diff
@@ -291,11 +319,11 @@ class BrowserDetector {
...
-    //
-    // As of January 2020, this is mainly used to adjust the default UI config
-    // for mobile devices, so it's low risk if something changes to break this
-    // detection.
-    return this.isApple() && navigator.maxTouchPoints > 1;
+    if (process.client) {
+      return this.isApple() && navigator.maxTouchPoints > 1;
+    } else {
+      return false;
+    }
```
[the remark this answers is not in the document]
MrTimscampi: Imo it's better to be explicit with this and make them easy to read blocks.

We might want to adjust the SSR-context behavior at some point, and this makes it easy to read and modify. Your proposal looks like the kind of things we've spent years refactoring away in jf-web :yum:

client/components/Item/Card/Card.vue
ThibaultNocchi: So maybe use $vuetify?

client/plugins/browserDetection.ts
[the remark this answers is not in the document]
ThibaultNocchi: Ok np

client/components/Item/Card/Card.vue
MrTimscampi: No, because we don't want small desktop windows to hide the overlay. Same this as below, this had a different purpose.

I added possible suitable replacement in the latest commit, using `window.matchMedia('(pointer:fine)').matches`.

ThibaultNocchi (approved):

MrTimscampi:
@jellyfin rebase

client/pages/item/_itemId/index.vue
```diff
@@ -218,17 +194,13 @@
               </v-row>
               <v-row align="center">
                 <v-col
-                  :cols="twoColsInfoColumn.lCols"
-                  :sm="twoColsInfoColumn.lSm"
-                  :class="twoColsInfoColumn.lClass"
+                  :cols="12"
+                  :sm="2"
+                  class="mt-sm-3 py-sm-0 px-0 text-truncate"
```
dkanada, line 199: I actually think the previous way was more readable. I've always found the flexbox sizing extremely annoying to read.

client/plugins/playbackProfiles/helpers/fmp4AudioFormats.ts
```diff
@@ -25,23 +27,23 @@ export function getSupportedFmp4AudioCodecs(
     codecs.push('mp3');
   }
 
-  if (hasAc3Support(videoTestElement)) {
-    if (hasAc3InHlsSupport(videoTestElement)) {
+  if (hasAc3Support(context, videoTestElement)) {
+    if (hasAc3InHlsSupport(context, videoTestElement)) {
```
dkanada, line 31: Passing context around everywhere is starting to remind me of Android development haha.

client/pages/item/_itemId/index.vue
MrTimscampi: Eh, it's a habit to take, imo. This way is similar to what you'd have with Tailwind or other CSS utilities.

I generally find having to refer to code when I want to change more annoying. Plus if I want to change one, I'd have to refactor everything.

client/plugins/playbackProfiles/helpers/fmp4AudioFormats.ts
MrTimscampi: Indeed, but there isn't much else we can do :smile:

client/pages/item/_itemId/index.vue
dkanada: I think the biggest problem I have here (and with lots of other implementations using class-based css frameworks) is that these won't be standardized across all the page layouts. There isn't a good solution for pages like there is for components besides making them as generic as possible, but I figured it was worth mentioning.

ferferga (approved):

PR #937, opened 2021-03-30
fix(ui): margin and layout changes for audio controls

dkanada:
- item menu button was likely made small for TrackList but doesn't match any other situations with normal sized icons
- album art in AudioControls had no bottom padding and the Name - Artist labels next to it had odd vertical alignment
- explicitly setting elevation to zero will cause a strange border on Firefox and should be avoided
- moved the fullscreen button to the end since this is where most people would expect a fullscreen button
- removed left padding on fullscreen button

### DIFF
--- a/client/components/Item/ItemMenu.vue
+++ b/client/components/Item/ItemMenu.vue
@@ -10,7 +10,6 @@
       <template #activator="{ on, attrs }">
         <v-btn
           icon
-          small
           :outlined="outlined"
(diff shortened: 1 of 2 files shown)


MrTimscampi (approved):

ThibaultNocchi (approved):

MrTimscampi:
@jellyfin rebase

PR #944, opened 2021-04-01
fix: items plugin was unregistered

ThibaultNocchi:
With the item PR, I forgot to `git add` my changes to nuxt.config.ts where I registered the plugin. It's now done


MrTimscampi (approved):

MrTimscampi:
Forcing this through with one review, since it fixes a major issue that prevents using the client and is one line.

PR #949, opened 2021-04-02
fix(ui): miscellaneous string and layout fixes

dkanada:
This is a bit of a vague pull request but none of these changes seemed large enough to split into a separate branch.

### DIFF
--- a/client/components/Item/Metadata/MetadataEditor.vue
+++ b/client/components/Item/Metadata/MetadataEditor.vue
@@ -19,7 +19,7 @@
             <v-text-field
               v-model="metadata.Name"
               outlined
-              :label="$t('title')"
+              :label="$t('item.tracklist.title')"
             />
             <v-text-field
               v-model="metadata.OriginalTitle"
--- a/client/components/Item/Metadata/MetadataEditor.vue
+++ b/client/components/Item/Metadata/MetadataEditor.vue
@@ -19,7 +19,7 @@
             <v-text-field
               v-model="metadata.Name"
               outlined
-              :label="$t('item.tracklist.title')"
+              :label="$t('metadata.title')"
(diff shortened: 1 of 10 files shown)


ThibaultNocchi:
Idk how I feel about removing the `ml-1` from the items, it puts the play button really close on the left to the artwork

### master
print_screen_2021-04-02-09-30-58

### PR
print_screen_2021-04-02-09-31-17

client/components/Item/Metadata/MetadataEditor.vue
```diff
@@ -19,7 +19,7 @@
             <v-text-field
               v-model="metadata.Name"
               outlined
-              :label="$t('title')"
+              :label="$t('item.tracklist.title')"
```
MrTimscampi, line 22: I think the metadata editor shouldn't depend on the track list. This thing needs to essentially be entirely rewritten anyway.

I know you did this because currently the key doesn't exist, but something like `item.metadata.title` is probably better, if only because the term used in the track list and the metadata editor might be different between languages depending on context.

Plus we can group translations together here (which we can't in jf-web) so it's less of an issue for organization. Weblate can also help a lot if we handle rules and the glossary properly.

MrTimscampi (approved):

MrTimscampi:
> it puts the play button really close on the left to the artwork

I think there's something wrong with CSS somewhere in general, because a bunch of padding and margins are just completely ignored. They show up in the dev tools, but none of the items inside take them into account. This seems to happen on most pages.

MrTimscampi (approved):

ThibaultNocchi (approved):

MrTimscampi:
@jellyfin-bot rebase

PR #950, opened 2021-04-02
fix(ui): update global style for tables and server info element

dkanada:
There were certain changes I made global to apply to all table elements, and others I kept scoped to this specific info element like the bottom border removal.


dkanada:
I had no good ideas for the custom variable name, but track base seemed too specific considering its possible use cases. It's basically some kind of secondary background color used for tons of elements.

ThibaultNocchi (approved):

MrTimscampi (approved):

MrTimscampi:
@jellyfin rebase

PR #953, opened 2021-04-04
fix(ux): always update loading status and update server data on login

### LINKED ISSUE #906: Branding text doesn't get reset when changing between servers
**Describe The Bug**

When connecting to a server with branding text (Like the demo server) and switching to another server without branding, the branding is never reset, leading to the wrong branding shown.

**Steps To Reproduce**

1. Connect to `demo.jellyfin.org`
2. Switch servers to a server without branding

**Expected Behavior**

The branding should be reset properly when switching servers.

dkanada:
Fixes #906


ThibaultNocchi:
Idk how it was before, but the branding appearing after the page load is kinda janky

MrTimscampi:
> Idk how it was before, but the branding appearing after the page load is kinda janky

We can remove the keep-alive prop in `fullpage.vue` (since it's only used for the login flow anyway) and revert the fetch bits to fix this.

client/pages/server/login.vue
```diff
@@ -57,23 +57,6 @@ export default Vue.extend({
...
-  async asyncData({ store, redirect, $api }) {
-    try {
-      await store.dispatch(
-        'servers/connectServer',
-        store.state.servers.serverUsed.address
-      );
-
-      const publicUsers = (await $api.user.getPublicUsers({})).data;
-
-      const brandingData = (await $api.branding.getBrandingOptions()).data;
```
MrTimscampi, line 69: As @ThibaultNocchi mentioned, moving the branding to `fetch()` makes it a bit jarring (Same for public users).

As a proper fix for them not being refreshed, while keeping the nicer experience, we can add an exclude list to the `keep-alive` here: https://github.com/jellyfin/jellyfin-vue/blob/master/client/layouts/fullpage.vue#L4

Just exclude the server pages (We might use fullpage for other things in the future) and that should fix these not being re-fetched when in `asyncData`.

MrTimscampi (changes requested):
Looks good, but there's a nicer UX solution for the fetching issue :yum:

client/pages/server/login.vue
MrTimscampi: More info on the available keep-alive props: https://v3.vuejs.org/api/built-in-components.html#keep-alive

MrTimscampi (approved):

PR #956, opened 2021-04-04
feat(blurhash-image): use web assembly version of the blurhash decoder

MrTimscampi:
For speed and glory, this replaces the web worker decoding with the (way faster) Rust-based one.

The import is a bit janky, but we have to force a split in the Webpack bundles by using a dynamic import, hence this madness.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,5 @@ dist
 .vscode
 .devcontainer
 sw.*
+client/.nuxt
+client/dist
--- a/client/components/Item/Card/Card.vue
+++ b/client/components/Item/Card/Card.vue
@@ -208,8 +208,8 @@ export default Vue.extend({
       }
     },
     getImageType(): ImageType {
-      if (this.shape === 'thumb-card' && this.item.Type === 'Movie') {
-        return ImageType.Backdrop;
+      if (this.shape === 'thumb-card') {
+        return ImageType.Thumb;
       } else {
         return ImageType.Primary;
       }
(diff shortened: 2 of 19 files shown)


ferferga:
Some interesting findings:

### WASM in the main thread (this PR)
wasm

### JS in Worker (master)
worker

### JS in the main thread
mainJS

### WASM in Worker
worker+wasm

Branch that has the test implementation live at [blurhash-benchmark](https://github.com/jellyfin/jellyfin-vue/tree/blurhash-benchmark)

Some conclusions:

* Clearly the speed advantage of the WASM implementation in this PR is just placebo as it doesn't have the overhead that workers introduce.
* JS actually seems to win over WASM on non JS heavy scenarios (it only loses on the first decoded hash, where all the page content is still being loaded so the main thread is quite busy).

Even if WASM is consistently faster, even in JS-heavy scenarios, I would still run it on a worker. [WASM introduces its own overhead to the main tread](https://www.sitepen.com/blog/using-webassembly-with-web-workers) and the time the function it's working is still essentially blocked, as WASM is not async by any means, it's just compiled code. 

Workers also queue stuff, which is something that wasm doesn't: If 10 blurhash-canvas components are mounted and the first's hash isn't decoded yet, the worker will queue all those calls, but **it will never block new canvases for calling it**. WASM functions are still classic JS function in that regard. More info about this [here](https://developer.mozilla.org/es/docs/Web/API/Web_Workers_API) and [here](https://stackoverflow.com/a/53471808)

MrTimscampi:
I accidentally killed the branch (don't name two branch the same thing...) and this might not be worth it anyway, so let's close it for now.

PR #964, opened 2021-04-05
feat(audio): gapless playback

ferferga:
Marking as draft as I didn't have the chance to test proper gapless tracks, but I checked some tracks that start rightaway and it's absolutely seamless. Once I test with gapless files I will label as ready.

* Loads 3 audio players to prefetch next tracks (easily changeable in the future) and switches seamlessly between them.
* Disables progress bar when fetching playback info (loading them into Shaka doesn't show any progress either)

The multiple audio player approach is not only helpful for prefetching but also for crossfade features once that's worked on as well.

### DIFF
--- a/client/components/Players/AudioPlayer.vue
+++ b/client/components/Players/AudioPlayer.vue
@@ -1,12 +1,15 @@
 <template>
-  <audio
-    ref="audioPlayer"
-    autoplay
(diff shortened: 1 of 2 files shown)


sapphonie:
errr... was this merged?

PR #967, opened 2021-04-06
refactor: use mapState instead of direct store access everywhere

ferferga:


ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

client/components/Layout/AudioControls.vue
```diff
@@ -213,31 +213,26 @@ export default Vue.extend({
...
         return 'mdi-repeat-once';
       }
 
       return 'mdi-repeat';
     },
-    isShuffling(): boolean {
-      return this.$store.state.playbackManager.isShuffling;
-    },
+    // TODO: Checking for route is faster to switch the controls than checking for the store
+    // Remove this as soon as we can use the fullpage layout in music pplayer
```
MrTimscampi, line 234:
```suggestion
    // TODO: Remove this as soon as we can use the fullpage layout in music pplayer
```

Makes it appear in todolists, so it's easier to fix in the future :)

client/components/Players/DraggableQueue.vue
```diff
@@ -80,8 +80,9 @@ export default Vue.extend({
   methods: {
     ...mapGetters('playbackManager', ['getCurrentItem']),
     ...mapActions('playbackManager', ['setNewQueue', 'setCurrentIndex']),
+    ...mapState('playbackManager', ['currentItemIndex']),
     isPlaying(index: number): boolean {
-      return index === this.$store.state.playbackManager.currentItemIndex;
+      return index === (this.currentItemIndex as unknown);
```
MrTimscampi, line 85: This should probably mirror the type defined in the store, instead.

MrTimscampi (approved):

client/components/Players/DraggableQueue.vue
ferferga: The problem is that mapState returns ``Computed`` as type, so it doesn't have the proper typings. I hope this is fixed in Vuex 4.

MrTimscampi:
```suggestion
      return index === (this.currentItemIndex as number | null);
```

What I mean is doing this, so we get the proper types, just in case :)

ferferga: Needs conversion to unknown first as the types don't overlap, will try later ``(this.currentItemIndex as unknown) as number``

client/pages/fullscreen/playback/index.vue
```diff
@@ -55,33 +55,20 @@ export default Vue.extend({
   },
   computed: {
     ...mapGetters('playbackManager', ['getCurrentItem']),
-    currentItemIndex: {
-      get(): number {
-        return this.$store.state.playbackManager.currentItemIndex;
-      }
-    },
-    currentQueue: {
```
ThibaultNocchi, line 63: You haven't changed a reference to currentQueue on L5 :)

ThibaultNocchi (changes requested):

client/components/Forms/AddServerForm.vue
```diff
@@ -58,14 +58,17 @@ export default Vue.extend({
       previousServerLength: 0
     };
   },
+  computed: {
+    ...mapState('servers', ['serverList'])
+  },
```
dkanada, line 63: What happened to the nice TypeScript mapState we were talking about ages ago?

ferferga: This is another project I discovered while working in this PR: https://github.com/maoberlehner/vuex-map-fields

Right now we can't use ``get/set`` computed properties as well, which leaves the ``DraggableQueue`` component with the old way of doing things.

iirc all the typescript stuff will be sorted out in Vuex 4

ferferga: I've been looking further for this and I confirm it's right around the corner:

https://github.com/vuejs/vuex/releases/tag/v4.0.0-beta.1

MrTimscampi: Not sure that'll work with Vue 2, though. It mentions pretty specifically Vue 3 as the main thing.

ferferga: Yeah, I was talking specifically about when we switch to Vue 3, in fact I added it to the comments of the ``this.currentItemIndex as unknown as number`` thing.

It's a long way to go still but imo not worth it to do an overhaul now when we will need to do it later with Vuex as well, just for the same purpose.

ThibaultNocchi (approved):

ferferga:
@jellyfin-bot rebase

PR #970, opened 2021-04-06
feat(baseurl): automatically prepend baseUrl when routing

crobibero:
This mostly works, as mentioned in chat loading fonts doesn't currently work.
Attempts to load `/_nuxt/css/_nuxt/fonts/{font}.woff2` instead of `/_nuxt/fonts/{font}.woff2`


MrTimscampi:
Closing this, since it's moot with the Vue 3 migration. We'll have a lot more control on bundling and routing too, so this should get fixed then.

PR #972, opened 2021-04-06
fix(library): fixed mixed library item fetching

MrTimscampi:
Fixes item listing in mixed libraries.

It was fetching recursively before, when it shouldn't have been.


hamburglar2160:
Mixed libraries work!

I only have one music album in my library now, apparently.

MrTimscampi:
> Mixed libraries work!

They don't, the current implementation lists recursively, so you get seasons and such in the list.

ThibaultNocchi (approved):

dkanada (approved):

PR #973, opened 2021-04-06
feat(websocket): handle UserDataChanged messages to update the item store

ferferga:
Implements the item store for the like button and markplayed button

### DIFF
--- a/client/store/plugins/websocket.ts
+++ b/client/store/plugins/websocket.ts
@@ -28,40 +28,64 @@ export const websocketPlugin: Plugin<AppState> = (store) => {
     store.state.socket.instance?.send(JSON.stringify(msg));
   }
 
-  store.subscribe(async (mutation, state) => {
-    // Messages
-    // Keep the websocket alive
-    if (
-      mutation.type === 'SOCKET_ONMESSAGE' &&
-      state.socket.message.MessageType === 'ForceKeepAlive'
-    ) {
-      sendWebsocketMessage('KeepAlive');
-      keepAliveInterval = window.setInterval(() => {
-        sendWebsocketMessage('KeepAlive');
-      }, state.socket.message.Data * 1000 * 0.5);
+  /**
+   * Updates the items in the store after the websocket informs of it
+   *
+   * @param {string[]} itemIds - Ids of the items to update
+   */
+  async function updateStoreItems(itemIds: string[]): Promise<void> {
+    if (itemIds.length) {
+      const updatedItems: BaseItemDto[] =
+        (
+          await store.$api.items.getItems({
+            userId: store.$auth.user?.Id,
+            ids: itemIds
+          })
+        ).data.Items || [];
+
+      store.dispatch('items/addItems', { items: updatedItems });
     }
-    // Update items when changed
-    else if (
-      mutation?.type === 'SOCKET_ONMESSAGE' &&
-      state.socket.message.MessageType === 'LibraryChanged'
-    ) {
-      // @ts-expect-error -- The Data property doesn't describe its content
-      const itemsToUpdate: string[] = state.socket.message.Data.ItemsUpdated.filter(
-        (itemId: string) => {
-          return store.state.items.allIds.includes(itemId);
-        }
-      );
+  }
 
-      if (itemsToUpdate) {
-        const updatedItems: BaseItemDto[] =
-          (
-            await store.$api.items.getItems({
-              userId: store.$auth.user?.Id,
-              ids: itemsToUpdate
-            })
-          ).data.Items || [];
+  store.subscribe((mutation, state) => {
+    const message = state.socket.message;
+    const storeIds = store.state.items.allIds;
+    let itemsToUpdate: string[];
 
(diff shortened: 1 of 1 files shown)


ferferga:
I have no idea why tests are failing

client/components/Buttons/LikeButton.vue
```diff
@@ -23,39 +23,22 @@ export default Vue.extend({
...
-  watch: {
-    item: {
-      immediate: true,
-      handler(): void {
-        this.isFavorite = this.item.UserData?.IsFavorite || false;
-      }
+  computed: {
+    ...mapGetters('items', ['getItem']),
+    observedItem(): BaseItemDto {
+      return this.getItem(this.item.Id as string);
```
ThibaultNocchi, line 32: Why not watch this.item?

client/components/Buttons/MarkPlayedButton.vue
```diff
@@ -32,33 +32,42 @@ export default Vue.extend({
...
+    }
+  },
   watch: {
-    item: {
-      immediate: true,
-      handler(): void {
-        this.isPlayed = this.item.UserData?.Played || false;
-      }
+    observedItem(): void {
+      this.isPlayed = this.observedItem?.UserData?.IsFavorite || false;
```
ThibaultNocchi, line 43: I don't think this should be `IsFavorite` :p

client/components/Buttons/LikeButton.vue
```diff
@@ -23,39 +23,22 @@ export default Vue.extend({
...
-    });
+  watch: {
+    observedItem(): void {
+      this.isFavorite = this.observedItem?.UserData?.IsFavorite || false;
+    }
   },
-  deactivated() {
-    this.unsubscribe();
+  async created() {
+    await this.$userLibrary.getItem(this.item.Id as string);
```
ThibaultNocchi, line 41: This forces fetching the item, so it kinda defeats the purpose of the store. Maybe check for its presence first.

client/components/Buttons/MarkPlayedButton.vue
```diff
@@ -32,33 +32,42 @@ export default Vue.extend({
...
     }
   },
-  created() {
-    this.isPlayed = this.item.UserData?.Played || false;
+  async created() {
```
ThibaultNocchi, line 46: Same concern about force fetching

client/components/Buttons/LikeButton.vue
MrTimscampi: I think it's there because most of the places don't use the store at the moment. We should finish implementing the store everywhere before refactoring this component, imo.

ThibaultNocchi: Then I'd say just check for presence and fetch if needed :p

MrTimscampi: Ideally, they'd be fetched outside of this, though, hence why I think doing all the rest before refactoring this is the better route :yum:

ThibaultNocchi: I agree. But I think it doesn't cost too much to add the if for now :)

ferferga:
@MrTimscampi I've been looking for this  deeply and I completely agree with your comment that this should be done from top -> bottom way from the start, instead of an stopgap solution like this. I can start doing that.

I'll rebase this PR to keep the refactoring of the websocket plugin so it also gets the items whose UserData has been updated.

Also, I think maybe it's a good idea to make the plugin fetch from the store instead if the item already exists in the store? We're going to repeat a lot of ``if``s otherwise.

MrTimscampi:
> Also, I think maybe it's a good idea to make the plugin fetch from the store instead if the item already exists in the store? We're going to repeat a lot of `if`s otherwise.

This was discussed in length with @ThibaultNocchi when he implemented the store.

The idea of keeping the if outside is because we only want to wrap the API, and we might want different checks or different logic around the fetch (or even just overriding the check in some cases).

client/store/plugins/websocket.ts
```diff
@@ -28,40 +28,65 @@ export const websocketPlugin: Plugin<AppState> = (store) => {
...
     // Messages
     // Keep the websocket alive
-    if (
-      mutation.type === 'SOCKET_ONMESSAGE' &&
-      state.socket.message.MessageType === 'ForceKeepAlive'
-    ) {
-      sendWebsocketMessage('KeepAlive');
-      keepAliveInterval = window.setInterval(() => {
+    if (mutation.type === 'SOCKET_ONMESSAGE') {
+      if (message.MessageType === 'ForceKeepAlive') {
```
MrTimscampi, line 57: Maybe a switch would make more sense here, now? :thinking:

MrTimscampi (approved):

client/store/plugins/websocket.ts
ferferga: Good idea 👍.

MrTimscampi (approved):

ThibaultNocchi (approved):

PR #974, opened 2021-04-06
docs: add link to Cloudflare pages to README

ferferga:


hamburglar2160:
Should https://jellyfin-vue.pages.dev/ be updating automatically when master is built? As far as I can tell, it's still on a build from at least a week ago.

ferferga:
It should. How do you know it's from a week?

EDIT: I actually checked and I can confirm the commits of today are not included.

README.md
```diff
@@ -36,6 +36,8 @@
 
 This is an experimental web client for Jellyfin based on Vue.js. We welcome all contributions and pull requests! If you have a larger feature in mind please open an issue so we can discuss the implementation before you start.
 
+An always up-to-date build from master branch is available at https://jellyfin-vue.pages.dev/
```
MrTimscampi, line 39: We could maybe use [this](https://img.shields.io/badge/pages.dev-master-success) instead?

I'm mainly not quite sure about the formulation of this and the place it should live in inside the readme.

ThibaultNocchi (approved):

README.md
ferferga: I'm fine with either but I think the direct mention is *slightly* better for users that don't know how to use GitHub and are not used to understand/see those badges.

Most projects have always a "View Demo" or "Online Demo" mention in their readmes rightaway, no badges in the middle. 

Perhaps I should change the formulation to "Demo" as well.

MrTimscampi: Yeah, I think a "View demo" link would fit better :)

dkanada: I agree with @ferferga here, the badges aren't very good for newcomers, an explicit link would probably be best.

MrTimscampi: I agree too, but a big "View demo" link feels better, then, as proposed by ferferga :yum:

ThibaultNocchi (approved):

MrTimscampi (approved):

PR #977, opened 2021-04-07
fix(layout): exclude fullscreen-playback from the pages kept alive

### LINKED ISSUE #976: Cannot select a specific song from an album
**Describe The Bug**

Seletcing a song from an album just plays the first one, and kicks me out of the album.

**Steps To Reproduce**

1. Open an album, and select a song thats not 1
2. See that it plays the first track, and youre now back in album view.

**Expected Behavior**
To play the song I chose and not kick me out.

**System (please complete the following information):**

- Platform: Docker
- Browser: Brave
- Jellyfin Version: 1.70 stable.

MrTimscampi:
Fixes #976 (at least it should)

### DIFF
--- a/client/layouts/default.vue
+++ b/client/layouts/default.vue
@@ -100,7 +100,10 @@
     </v-app-bar>
     <v-main>
       <div class="pa-s">
-        <nuxt keep-alive :keep-alive-props="{ max: 10 }" />
+        <nuxt
+          keep-alive
+          :keep-alive-props="{ max: 10, exclude: ['fullscreen-playback'] }"
+        />
       </div>
     </v-main>
     <audio-controls />


client/layouts/default.vue
```diff
@@ -100,7 +100,10 @@
     </v-app-bar>
     <v-main>
       <div class="pa-s">
-        <nuxt keep-alive :keep-alive-props="{ max: 10 }" />
+        <nuxt
+          keep-alive
+          :keep-alive-props="{ max: 10, exclude: ['fullscreen-playback'] }"
```
ferferga, line 105: Non blocking and super nitpicky, but maybe it's not cleaner to set the passed object in ``data`` (now that we're expanding it) for maintainability?

Also, I think you meant fullscreen/playback instead?

MrTimscampi: > Non blocking and super nitpicky, but maybe it's not cleaner to set the passed object in data (now that we're expanding it) for maintainability?

Setting it in data kind of implies a level of reactivity, which this doesn't have. I don't think it's necessary here.

> Also, I think you meant fullscreen/playback instead?

No, this is correct. It's the *router name*, not the route itself. You can guess them fairly easily if you know how they're build, otherwise you can look them up in the Vue Dev Tools, in Routing > Routes.

ferferga (approved):

ThibaultNocchi (approved):

PR #982, opened 2021-04-09
feat(robots): disallow all crawling

MrTimscampi:
As reported by Lighthouse, we don't have a robots.txt:
image

The one in this disallows all crawling, as a default. It is easily changeable if, for whatever reason, someone wants to make their server public and searchable through Google or something (Though there is no auto-login for a public user at this point in time).

### DIFF
--- a/client/static/robots.txt
+++ b/client/static/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /


ferferga:
Can't we simply copy the one we have in jf-web under static? 

We don't need to change it programatically at any point to deserve a module imo. It should also make this change easier for people that don't want to rebuild the client.

MrTimscampi:
> Can't we simply copy the one we have in jf-web under static?
> 
> We don't need to change it programatically at any point to deserve a module imo. It should also make this change easier for people that don't want to rebuild the client.

It has to be in the root directory of the built client. Putting it in static wouldn't work, I believe.

ferferga:
@MrTimscampi [Yes](https://nuxtjs.org/docs/2.x/directory-structure/static)

MrTimscampi:
> @MrTimscampi [Yes](https://nuxtjs.org/docs/2.x/directory-structure/static)
> 
> That's exactly how we're bundling our images now as well.

Oh, good. I thought it was mapped to like /static or something :yum: Replaced it with a static file then, that's one less dependency.

ThibaultNocchi (approved):

ferferga (approved):

PR #986, opened 2021-04-09
Feat/episode serie link

### LINKED ISSUE #985: Make the show clickable on episode card
**Describe The Feature**
On Continue Watching or Next Up screen, what's being displayed is an episode card that links to the episode page. At its bottom, the episode and the show names are displayed.

It'd be great to be able to click on the show name to get redirected to its page.

ThibaultNocchi:
* Moved the surrounding nuxt-link tag to only the image part
* Moved the hover effect to only trigger when hovering the image part
* Removed the nuxt-link attribute "event" as it seems to work without (please tell me if it was needed @ferferga, I blamed you to find you :p)
* Added the same nuxt-link as the image to the title
* Moved the title and subtitle link logics to a computed method
* Added the subtitle link to the show under episodes
* Edited the current subtitle link to an <a> tag to get the browsers feature of the URL tooltip and URL copy
* Changed surrounding nuxt-link to a variable component: a nuxt-link or div whether `link` is true
* Removed ineffective classes 

You can review by commit to be easier, due to the first commit moving the surrounding nuxt-link inside and moving every tabulation around

Fixes #985

### DIFF
--- a/client/components/Item/Card/Card.vue
+++ b/client/components/Item/Card/Card.vue
@@ -229,6 +229,8 @@ export default Vue.extend({
           this.item.AlbumArtists[0],
           'MusicArtist'
         );
+      } else if (this.item.Type === 'Episode') {
+        return this.getItemDetailsLink(this.item);
       }
 
       return undefined;
--- a/client/components/Item/Card/Card.vue
+++ b/client/components/Item/Card/Card.vue
@@ -78,10 +78,9 @@
       </nuxt-link>
       <!-- CARD SUBTITLE -->
       <nuxt-link
-        v-if="item.Type === 'MusicAlbum' && item.AlbumArtists.length > 0"
-        tag="div"
-        class="card-subtitle text--secondary text-truncate link"
-        :to="getItemDetailsLink(item.AlbumArtists[0], 'MusicArtist')"
+        v-if="cardSubtitleLink"
+        class="link d-block card-subtitle text--secondary text-truncate"
+        :to="cardSubtitleLink"
       >
         {{ cardSubtitle }}
       </nuxt-link>
@@ -216,6 +215,23 @@ export default Vue.extend({
       }
 
       return this.getItemDetailsLink(this.item);
+    },
+    /*
+     * @returns {string|undefined} A link to be applied to the subtitle
+     */
+    cardSubtitleLink(): string | undefined {
+      if (
+        this.item.Type === 'MusicAlbum' &&
+        this.item.AlbumArtists &&
+        this.item.AlbumArtists.length > 0
+      ) {
+        return this.getItemDetailsLink(
+          this.item.AlbumArtists[0],
+          'MusicArtist'
+        );
+      }
+
+      return undefined;
     },
     progress: {
       get(): number | false {
--- a/client/components/Item/Card/Card.vue
+++ b/client/components/Item/Card/Card.vue
@@ -1,7 +1,6 @@
 <template>
   <div :class="{ 'card-margin': margin, 'link-disabled': !link }">
     <nuxt-link
-      :event="link ? 'click' : null"
       :to="getItemDetailsLink(item)"
       :class="link ? null : 'link-disabled'"
       class="card-box nuxt-link"
--- a/client/components/Item/Card/Card.vue
+++ b/client/components/Item/Card/Card.vue
@@ -1,14 +1,12 @@
 <template>
-  <nuxt-link
-    :event="link ? 'click' : null"
-    :to="getItemDetailsLink(item)"
-    :class="link ? null : 'link-disabled'"
-    class="nuxt-link"
-  >
-    <div
-      class="card-box"
-      :class="{ 'card-margin': margin, 'link-disabled': !link }"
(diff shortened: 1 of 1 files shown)


ferferga:
Iirc the event was there to also disable the click event, as CSS alone didn't work for me. If hover and clicking on the card doesn't redirect you to the page in the current state, then it's good to remove.

MrTimscampi:
> Iirc the event was there to also disable the click event, as CSS alone didn't work for me. If hover and clicking on the card doesn't redirect you to the page in the current state, then it's good to remove.

Would setting `:to="null"` maybe achieve the same effect, maybe?

MrTimscampi (approved):

neopc10:
Individual episodes the TV Show/Season details screen would benefit from this treatment, too. I think after this is merged they'd be the only items missing links.

client/components/Item/Card/Card.vue
```diff
@@ -69,22 +66,29 @@
...
-          class="card-subtitle text--secondary text-truncate link"
-          :to="getItemDetailsLink(item.AlbumArtists[0], 'MusicArtist')"
-        >
-          {{ cardSubtitle }}
-        </nuxt-link>
-        <div v-else class="card-subtitle text--secondary text-truncate">
-          {{ cardSubtitle }}
-        </div>
+    </component>
+    <!-- CAPTIONS -->
```
dkanada, line 70: I don't think these title comments help anything personally.

client/components/Item/Card/Card.vue
```diff
@@ -202,6 +206,35 @@ export default Vue.extend({
 
       return '';
     },
+    /**
+     * @returns {string} A link to be applied to the title
+     */
```
dkanada, line 211: Do we want full sentences for the docs or just capital letters at the start? Honestly not sure what the plan is for this project specifically.

ThibaultNocchi: @mrtimscampi

client/components/Item/Card/Card.vue
MrTimscampi: If there's the need to add comments to a template, it's time to split that part of the template into a component, imo.

ThibaultNocchi: Yeah no worry, it was there to help me visualise where the stuff was, but it's not needed anymore

client/components/Item/Card/Card.vue
MrTimscampi: We seem to be using capital letters almost everywhere, so that seems to be the style we're going for.

For sentences or not, it seems to vary a bit. I'm leaning towards just describing what it returns, instead of what it should be used for. So here, something like `A router link to the item's page` or something similar, maybe?

ThibaultNocchi: Though it's not always the item's page ^^

MrTimscampi: True, but you get the idea :yum:  Basically describe the thing it returns instead of how it's used :smile:

client/components/Item/Card/Card.vue
```diff
@@ -214,7 +216,9 @@ export default Vue.extend({
       return this.getItemDetailsLink(this.item);
     },
     /*
-     * @returns {string|undefined} A link to be applied to the subtitle
+     * Gets a link to be applied to the card subtitle
+
+     * @returns {string|undefined} A router link to the item parent or related item
```
MrTimscampi, line 221:
```suggestion
     * @returns {string|undefined} A router link to the parent item or a related item
```

client/components/Item/Card/Card.vue
```diff
@@ -214,7 +216,9 @@ export default Vue.extend({
       return this.getItemDetailsLink(this.item);
     },
     /*
```
MrTimscampi, line 218: This needs to be `/**` otherwise it's not valid JSDOC.

ThibaultNocchi: I was wondering why the coloration didn't work

MrTimscampi (approved):

ferferga:
@jellyfin-bot rebase

client/components/Item/Card/Card.vue
```diff
@@ -202,6 +203,39 @@ export default Vue.extend({
...
+      ) {
+        return this.getItemDetailsLink(
+          this.item.AlbumArtists[0],
+          'MusicArtist'
+        );
+      } else if (this.item.Type === 'Episode') {
+        return this.getItemDetailsLink(this.item);
+      }
+
+      return undefined;
```
ferferga, line 237: This is unnecessary. Or return ``null`` if we explicitly want to know that this returned something.

ferferga (changes requested):

client/components/Item/Card/Card.vue
ThibaultNocchi: The return is necessary here cause it's an if else-if condition, and ESLint throws if it itsn't here

ThibaultNocchi: Returns null now

ferferga (approved):

Issue #980, opened 2021-04-08
Implement the new Items store in all the client


MrTimscampi:
This issue is meant to track the implementation of the items store in the rest of the app.

Please claim something if you're starting to work on it, to prevent multiple people working on it.

See #801 for more information behind the idea for the items store and how stuff should be organized around it.

* [x] RelatedItems (#994)
* [ ] MetadataEditor
* [ ] HomeHeader (@MrTimscampi)
* [ ] TrackList
* [ ] library/_viewId
* [ ] metadata
* [ ] store/homeSections (@MrTimscampi)
* [ ] store/tvShows (Has to be refactored according to #801)
* [ ] store/userViews (Has to be refactored according to #801)
* [x] translateItemsForPlayback (This should include a refactor of the playbackManager store according to #801) (@MrTimscampi #1008)

PR #994, opened 2021-04-10
feat(relateditems): implement store and fix reactivity

### LINKED ISSUE #990: related items are updated too frequently
The related items are reloaded every single time their base item changes, which also includes some unwanted updates. I propose the watch property is removed to solve it for now.

- favoriting an album
- skipping around while playing several songs in an album

ferferga:
Fix #990 and completes #980

### DIFF
--- a/client/components/Item/RelatedItems.vue
+++ b/client/components/Item/RelatedItems.vue
@@ -96,19 +96,17 @@ export default Vue.extend({
     'item.Id': {
       immediate: true,
       async handler(): Promise<void> {
-        if (this.item.Id) {
-          await this.refreshItems();
-        }
+        await this.refreshItems();
       }
     }
   },
   methods: {
     ...mapActions('snackbar', ['pushSnackbarMessage']),
     async refreshItems(): Promise<void> {
-      try {
-        this.loading = true;
+      if (this.item.Id) {
+        try {
+          this.loading = true;
 
-        if (this.item.Id) {
           const response = await this.$api.library.getSimilarItems({
             itemId: this.item.Id,
             userId: this.$auth.user?.Id,
@@ -130,14 +128,14 @@ export default Vue.extend({
 
             this.relatedItems = this.getItems(itemIds);
           }
+        } catch (error) {
+          this.pushSnackbarMessage({
+            message: this.$t('unableGetRelated'),
+            color: 'error'
+          });
+        } finally {
+          this.loading = false;
         }
-      } catch (error) {
-        this.pushSnackbarMessage({
-          message: this.$t('unableGetRelated'),
-          color: 'error'
-        });
-      } finally {
-        this.loading = false;
       }
     }
   }
--- a/client/components/Item/RelatedItems.vue
+++ b/client/components/Item/RelatedItems.vue
@@ -52,8 +52,8 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
-import { mapActions } from 'vuex';
+import Vue, { PropType } from 'vue';
+import { mapActions, mapGetters } from 'vuex';
 import { BaseItemDto } from '@jellyfin/client-axios';
 import imageHelper from '~/mixins/imageHelper';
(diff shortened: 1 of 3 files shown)


client/store/items.ts
```diff
@@ -19,8 +19,29 @@ export const state = defaultState;
...
+  getMissingIds: (state) => (ids: string[]): string[] => {
+    const missingIds = [] as string[];
+
+    ids.forEach((id) => {
+      if (!state.allIds.includes(id)) {
+        missingIds.push(id);
+      }
+    });
+
+    return missingIds;
```
ThibaultNocchi, line 31:
```suggestion
    return ids.filter((id) =>  !state.allIds.includes(id))
```

I think this should be better, wdyt?

client/components/Item/RelatedItems.vue
```diff
@@ -89,39 +89,56 @@ export default Vue.extend({
...
+      try {
+        this.loading = true;
 
-      if (this.item.Id) {
-        const response = await this.$api.library.getSimilarItems({
-          itemId: this.item.Id,
-          userId: this.$auth.user?.Id,
-          limit: this.vertical ? 5 : 12
-        });
+        if (this.item.Id) {
```
ThibaultNocchi, line 111: Quick question, what happens when there's no item ID? Is it caught by the trycatch? :)

ferferga: Nothing of the function runs and the function is called again in the watcher.

I noticed though that I had this both in this function and in the watcher, I'll remove it in the watcher to have all the proper type checking.

ThibaultNocchi (approved):

client/components/Item/RelatedItems.vue
```diff
@@ -89,39 +89,54 @@ export default Vue.extend({
...
-  beforeMount() {
-    try {
-      this.refreshItems();
-    } catch (error) {
-      this.pushSnackbarMessage({
-        message: this.$t('unableGetRelated'),
-        color: 'error'
-      });
+  watch: {
+    'item.Id': {
```
MrTimscampi, line 96: Is this really necessary? Afaik, this component isn't used anywhere where the item prop would change (When going between pages, a new one is created).

client/components/Item/RelatedItems.vue
```diff
@@ -89,39 +89,54 @@ export default Vue.extend({
...
+              }
+            );
+
+            const missingIds = this.getMissingIds(itemIds);
+
+            if (missingIds.length > 0) {
+              await this.$items.getItems({ ids: missingIds });
+            }
+
+            this.relatedItems = this.getItems(itemIds);
```
MrTimscampi, line 129: This entire section seems a bit convoluted (Mainly because of the API, but still).

We're essentially fetching the items twice here: once via `getSimilarItems` and then re-fetching the ones not in the store via `getItems`.

So we fetch a bunch of items... just to discard them completely. At that point, we may as well add them all to the store without any filtering.

MrTimscampi (changes requested):

client/components/Item/RelatedItems.vue
ThibaultNocchi: Aren't the items from the first query incomplete?

MrTimscampi: You can pretty much always specify which fields you want.

ThibaultNocchi: In that case, can't we in the first step only ask for IDs and then do the second step as here?

MrTimscampi: No, because you always get some info on the item. Plus the server will still have to run the queries to get what you want.

The issue here is doing 2 API calls, with all that implies: two handshakes, two requests, two DB queries, etc.

MrTimscampi:
Removing the confirmed tag, since that's not its use.

If this isn't abandoned, it should get finished, otherwise it should get closed.

MrTimscampi:
This has been untouched since June, so manually marking as stale and closing it.

PR #995, opened 2021-04-11
fix: genre chips margin

### LINKED ISSUE #991: genre chips are misaligned and disappear during playback
I noticed this issue on the album page, but it might apply to other pages with the genre chips. The margin issue is probably an easy fix but I have no clue why they disappear yet.

- [x] Misalignement #995
- [ ] Disappearing

ThibaultNocchi:
Partially corrects #991

### DIFF
--- a/client/pages/item/_itemId/index.vue
+++ b/client/pages/item/_itemId/index.vue
@@ -42,11 +42,7 @@
               v-if="item && item.GenreItems && item.GenreItems.length > 0"
               align="center"
             >
-              <v-col
-                :cols="12"
-                :sm="2"
-                class="mt-sm-3 py-sm-0 px-0 text-truncate"
-              >
+              <v-col :cols="12" :sm="2" class="px-0 text-truncate">
                 <label class="text--secondary">{{ $t('genres') }}</label>
               </v-col>
               <v-col class="px-0" :cols="12" :sm="10">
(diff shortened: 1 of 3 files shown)


MrTimscampi:
It looks better in the current master, imo. This PR makes the genres misaligned with everything else on the page, and create a lot of blank space.

Before
image
After
image

Before:
image
After:
image

ThibaultNocchi:
Nice catch, I went a little bit overboard by factorizing the margin class. Will fix it tomorrow

dkanada:
I don't know if it was mentioned, but the label should line up with the description below. It also looks like the director chips aren't aligned with the genre chips in the before image, perhaps that can be fixed as well. Maybe a component makes sense here.

ThibaultNocchi:
@MrTimscampi @dkanada fixed, I also removed `py-sm-0` as it didn't seem to change the design. But again feel free to say otherwise haha

client/pages/item/_itemId/index.vue
```diff
@@ -41,12 +41,9 @@
             <v-row
               v-if="item && item.GenreItems && item.GenreItems.length > 0"
               align="center"
+              class="mt-sm-3"
```
MrTimscampi, line 44: We can probably remove this, as it makes the gap between the genres and the button line really large (on all pages)

MrTimscampi (approved):

dkanada (approved):

PR #1018, opened 2021-04-14
chore: bump @jellyfin/client-axios to 10.7.4

ferferga:
I fixed the useSingleParameter option [upstream](https://github.com/jellyfin/jellyfin-client-axios/pull/12). Fixed package is up in NPM and GitHub Packages with version 10.7.4 (not yet tagged as latest due to a permission issue in CI)

However, there are some fields in the parameters that are like ``uNKNOWNBASETYPE`` for some reason. I wonder if it's a generator bug because we're fully ES6 now or something in server that's not compliant with OpenAPI spec.

This update also reduces the bundle size by quite a lot :) (master is 4.3 MB)
image

### DIFF
--- a/client/components/Item/Metadata/MetadataEditor.vue
+++ b/client/components/Item/Metadata/MetadataEditor.vue
@@ -359,7 +359,7 @@ export default Vue.extend({
         this.loading = true;
         await this.$api.itemUpdate.updateItem({
           itemId: this.metadata.Id as string,
-          baseItemDto: item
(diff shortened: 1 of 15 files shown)


MrTimscampi:
Marking as blocked due to an upstream bug with the Axios generator in `openapi-generator`.

https://github.com/OpenAPITools/openapi-generator/issues/9163

MrTimscampi:
Closing this, as the client is still not working properly and a new version would be published anyway.

PR #1026, opened 2021-04-19
refactor(css): add some context to a few hacks and remove duplicated rule

dkanada:
Just a few small things a chaned in my local branch that might be useful to merge.

### DIFF
--- a/client/components/Layout/VolumeSlider.vue
+++ b/client/components/Layout/VolumeSlider.vue
@@ -60,13 +60,4 @@ export default Vue.extend({
 .volume-slider {
   width: 10em;
 }
-
-// HACK: https://github.com/vuetifyjs/vuetify/issues/8436.
-// https://vuetifyjs.com/en/api/v-btn/#retain-focus-on-click prop was added
-// but it seems we're using a prop combination that it's incompatible with it.
-
-// SO link: https://stackoverflow.com/questions/57830767/is-it-default-for-vuetify-to-keep-active-state-on-buttons-after-click-how-do-yo/57831256#57831256
-.active-button:focus::before {
-  opacity: 0 !important;
-}
 </style>
--- a/client/assets/global.scss
+++ b/client/assets/global.scss
@@ -49,14 +49,25 @@ body {
 
 /* Vuetify overrides */
 
-// We can't set these through variables, so override them here */
 .v-app-bar.v-app-bar--fixed {
   z-index: 100 !important;
 }
 
-// Cropped text fixes. See https://github.com/vuetifyjs/vuetify/issues/6488 and https://github.com/vuetifyjs/vuetify/issues/4655 */
+// fixes cropped text in some situations
+// there are still issues with line height so this might be deprecated at some point
+// https://github.com/vuetifyjs/vuetify/issues/6488
+// https://github.com/vuetifyjs/vuetify/issues/4655
 .text-truncate {
   line-height: normal !important;
+}
+
+// retain-focus-on-click property is incompatible with NaN so we disable it manually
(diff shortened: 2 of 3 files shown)


ThibaultNocchi (approved):

client/layouts/default.vue
```diff
@@ -100,10 +100,7 @@
     </v-app-bar>
     <v-main>
       <div class="pa-s">
-        <nuxt
-          keep-alive
-          :keep-alive-props="{ max: 10, exclude: ['fullscreen-playback'] }"
-        />
+        <nuxt keep-alive :max="10" :exclude="['fullscreen-playback']" />
```
MrTimscampi, line 103: The Nuxt docs don't mention this being a thing. Do you have a source?

MrTimscampi (changes requested):
LGTM, except for the need for precision on one point.

client/layouts/default.vue
dkanada: Found them referenced in the [Vue](https://vuejs.org/v2/api/#keep-alive) documentation and they looked cleaner to me.

MrTimscampi: Seems like there's some level of abstraction in Nuxt, so it expects the ones we were using previously: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/components/nuxt.js#L27-L31

client/components/Layout/VolumeSlider.vue
```diff
@@ -60,13 +60,4 @@ export default Vue.extend({
 .volume-slider {
   width: 10em;
 }
-
-// HACK: https://github.com/vuetifyjs/vuetify/issues/8436.
-// https://vuetifyjs.com/en/api/v-btn/#retain-focus-on-click prop was added
-// but it seems we're using a prop combination that it's incompatible with it.
-
-// SO link: https://stackoverflow.com/questions/57830767/is-it-default-for-vuetify-to-keep-active-state-on-buttons-after-click-how-do-yo/57831256#57831256
-.active-button:focus::before {
```
ferferga, line 69: Did you check removing this actually works? I'm not sure if this was left here because the styling had to be scoped.

dkanada: It worked for me, is it not working on the demo for you?

client/layouts/default.vue
ferferga: This seems a good tradeoff between clean/functional:

image

dkanada: @MrTimscampi doesn't like them in data and I don't think it makes sense in props :p so I'll probably just revert this part.

MrTimscampi: It's not really that I don't like them in the data, it's that data adds a whole layer of reactivity that isn't needed :)

ferferga: That's why I propose setting them up in props, as they don't have the reactivity part.

client/components/Layout/VolumeSlider.vue
ferferga: I confirm it is

dkanada: It is or is not working?

ferferga: It is working :)

I'll approve as my complaints are now adressed, forgot to do that :)

ferferga (approved):

client/layouts/default.vue
dkanada: Pretty sure this has already been handled in another pull request so I'll close this one.

ferferga (approved):

MrTimscampi (approved):

PR #1027, opened 2021-04-19
refactor(header): remove welcome message and disable looping to avoid some issues

dkanada:
The redundant calls to parent items were also removed which will hopefully improve performance.


MrTimscampi (changes requested):
The code looks good to me, the transition from last to first looks okay as well.

Two issues though:

* There needs to be a placeholder while the swiper loads OR move the request for the items to asyncData in the parent page and pass them down (so everything is loaded by the time the swiper is shown. This would be the preferred option).
* Without the loop, we can't swipe from last to first anymore.

PR #1030, opened 2021-04-21
ci(github): use version and arch matrix to build

MrTimscampi:
**Edit: ARM seems busted, so it was removed**

To provide slightly better coverage of platforms and an easier upgrade path when a new Node LTS shows up, this adds matrices to both the architecture and Node versions.

We now build in CI:

* Node 14 on x64
* Node 14 on Armv7l
* Node 16 on x64
* Node 16 on Armv7l

This gives us an idea of how the client builds on both regular computers (x64) and on recent Raspberry Pi boards (armv7l), as well as warn us if a dependency fails with our next Node upgrade.

A change in requirements for CI passing will be done once this is merged:

* Node 14 on x64 and armv7l will become the required build jobs
* Node 16 jobs will be kept as optional, as it is expected they might fail in some cases

Once we move forward with Node versioning (Expected to happen next around June, when Node 16 becomes the current LTS), the matrix will change as follows:

* Node 14 -> Node 16
* Node 16 -> Node 17

So, essentially, the first entry is expected to track the current LTS, while the other will track the current stable.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,16 @@ on:
 
 jobs:
   ssr:
-    name: Build SSR
+    name: Build SSR (Node ${{ matrix.node }} - ${{ matrix.architecture }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node:
+          - 14
+          - 16
+        architecture:
+          - x64
 
     steps:
       - name: Checkout
@@ -18,7 +26,8 @@ jobs:
       - name: Setup node environment
         uses: actions/setup-node@v2.1.5
         with:
-          node-version: 14
+          node-version: ${{ matrix.node }}
+          architecture: ${{ matrix.architecture }}
 
       - name: Get npm cache directory path
         id: npm-cache-dir-path
@@ -28,10 +37,10 @@ jobs:
         uses: actions/cache@v2.1.5
         id: npm-cache
         with:
-          path: ${{ steps.npm-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          path: ~/.npm
+          key: ${{ runner.os }}-${{ matrix.architecture }}-node${{ matrix.node }}-npm-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
-            ${{ runner.os }}-npm-
+            ${{ runner.os }}-${{ matrix.architecture }}-node${{ matrix.node }}-npm-
 
       - name: Install dependencies
         run: npm ci --no-audit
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -29,7 +38,7 @@ jobs:
         id: npm-cache
         with:
           path: ${{ steps.npm-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          key: ${{ runner.os }}-${{ matrix.architecture }}-node${{ matrix.node }}-npm-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-npm-
```
MarcelCoding, line 43: You should make sure that you use a restore key that matches the node version and architecture.

.github/workflows/build.yml
```diff
@@ -61,7 +79,7 @@ jobs:
...
```
MarcelCoding, line 84: The same as above.

.github/workflows/build.yml
ThibaultNocchi: Isn't it already the case with $matrix.node and $matrix.archi? Idk much ^^

MarcelCoding: The key is the identify for the cache witch will be created. The restore key is the prefix for all keys witch are candidates for restoring the cache at the begin of a run.

MarcelCoding: Most likely you only want to restore a cache with the "correct" node version and architecture

.github/workflows/build.yml
```diff
@@ -8,8 +8,16 @@ on:
...
-    name: Build SSR
+    name: Build SSR (Node ${{ matrix.node }} - ${{ matrix.architecture }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node:
+          - 14
+          - 16
+        architecture:
```
ferferga, line 19: I don't know the purpose of this and maybe I'm understanding it wrong, but this doesn't work the way *I think* you want to. I also comment it because right now only x64 is used, but I recall seeing other arches in previous commits.

GitHub Actions runs in the same environment as Azure virtual machines. Specifically, [standard DS2_v2 series](https://docs.microsoft.com/en-us/azure/virtual-machines/sizes-general#dsv2-series), so all of it is Intel's

Full details about the environment is detailed here at GH docs: https://docs.github.com/es/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

What you want is something like this: https://github.com/marketplace/actions/run-on-architecture

.github/workflows/build.yml
```diff
@@ -60,10 +78,10 @@ jobs:
         uses: actions/cache@v2.1.5
         id: npm-cache
         with:
-          path: ${{ steps.npm-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          path: ~/.npm
```
ferferga, line 81: What was wrong with the old approach? Much safer than hardcoded defaults imo.

ferferga (changes requested):

.github/workflows/build.yml
MrTimscampi: We only test on Linux environments, so the path is always the same. This is what's used in the cache action's documentation when running on Linux hosts.

PR #1040, opened 2021-04-23
Remote devices and Google Cast support

MrTimscampi:
This is a potential followup to #964 

This is mainly a **proof of concept**, since we are encountering some API limitations on the remote devices side.

* Adds the Google Cast Sender SDK as a Nuxt plugin, loaded on Chrome-based browsers
* Adds a Google Cast button to the Remote devices menu, which initiates a Google Cast session (or closes it, if one is active)
* Adds a list of remote Jellyfin sessions to the Remote devices menu
* Adds `ìsRemote` and some remote queue management stuff to the playback manager (Buggy and non-viable, more below)
* Loads the remote queue when connecting to a Jellyfin session

Note: **This doesn't actually cast or remote control anything, at this stage.**

Note 2: **The Google Cast implementation is currently a bit rough and will show up on Chromium-based browsers without actual Cast support. This can lead to errors in some cases, this is normal and expected for now.**

### Google Cast connection
https://user-images.githubusercontent.com/19396809/115844535-37fa6e80-a420-11eb-820b-d6a18dfd45cf.mp4

### Remote session queue (Showing the limitation of the current API)
https://user-images.githubusercontent.com/19396809/115843533-38463a00-a41f-11eb-8439-798cae991794.mp4

Some discussion that needs to happen:

**What would be a good way to get the full list of items in the remote queue? Currently, the API sends a bunch of `QueueItem`, which are useless on their own.**

* One idea would be to introduce a `FullNowPlayingQueue`, similar to the existing `FullNowPlayingItem`, which would contain a bunch of `BaseItemDto`.
* Another one would be to provide a POST version of `/Items`, allowing us to send a (rather large) payload of IDs to get

Obviously the first solution would be nicer for clients, since we save a request. This issue is also common to all clients wanting to implement remote queue, since there is no viable way to convert a queue into items, at the moment.

**The Google Cast icon used to represent remote devices is not ideal. Is there a better option?**

That icon works if we only present Cast devices, but this menu contains SyncPlay groups, Google Cast devices, AirPlay devices and Jellyfin sessions. It would be better to have a relatively generic icon for this.

**Should the icon change depending on the currently connected device?**

An idea that was proposed when first showing this branch to the rest of the Jellyfin team was to potentially switch the button's icon to whatever we are connected to, as a way to quickly show what device is connected.

**What are the various Jellyfin sessions available?**

We show bespoke icons for different session types, based on the information sent by Jellyfin. It would be great to have a list of available DeviceName/DeviceType/Client that we could map to an icon, in order to show the proper icon for everything. Currently implemented are Firefox and Chrome.


client/server.crt
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
ferferga, line 1: What's the purpose of this?

client/server.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
ferferga, line 1: Also this one?

ferferga (commented):
Can you also cherrypick the locale changes to another PR please?

client/server.crt
MrTimscampi: Google Cast requires HTTPS to work, so this was a temporary certificate to perform some testing.

It's not meant to be kept in, but since this PR is essentially blocked until we figure out a solution for the queue, it's been kept in.

MrTimscampi:
Closing this, since we need server changes and it will likely be implemented nicer post-vue 3.

PR #1041, opened 2021-04-23
style(eslint): add no-restrict-globals with confusing-browser-globals

MrTimscampi:
Adds the `no-restrict-globals`, with [confusing-browser-globals](https://www.npmjs.com/package/confusing-browser-globals) to prevent using potentially confusing globals without using `window` first.

The list of provided globals is here: https://github.com/facebook/create-react-app/blob/master/packages/confusing-browser-globals/index.js


ThibaultNocchi (approved):
We didn't even have a single non passing case :o

MrTimscampi:
> We didn't even have a single non passing case :o

Yeah, I was surprised lol

ferferga (approved):

PR #1042, opened 2021-04-24
refactor(player): merge video-player and audio-player as shaka-player

MrTimscampi:
This removes a bunch of duplicate code by merging `video-player` and `audio-player` into a generic `shaka-player` component.


client/components/Players/ShakaPlayer.vue
```diff
@@ -208,30 +225,30 @@ export default Vue.extend({
     onPlay(_event?: Event): void {
       this.unpause();
     },
-    onVideoProgressThrottled: throttle(function (_event?: Event) {
-      // @ts-expect-error - TypeScript confuses the context with lodash's throttle typings
+    onProgressThrottled: throttle(function (_event?: Event) {
+      // @ts-expect-error - TypeScript is confusing the typings with lodash's
       this.onVideoProgress(_event);
     }, 500),
     onVideoProgress(_event?: Event): void {
```
ThibaultNocchi, line 232:
```suggestion
      this.onProgress(_event);
    }, 500),
    onProgress(_event?: Event): void {
```

client/components/Players/ShakaPlayer.vue
```diff
@@ -208,30 +225,30 @@ export default Vue.extend({
...
-      if (this.$refs.videoPlayer) {
-        const currentTime = (this.$refs.videoPlayer as HTMLVideoElement)
+      if (this.$refs.shakaPlayer) {
+        const currentTime = (this.$refs.shakaPlayer as HTMLVideoElement)
```
ThibaultNocchi, line 234:
```suggestion
        const currentTime = (this.$refs.shakaPlayer as HTMLMediaElement)
```

client/components/Players/ShakaPlayer.vue
```diff
@@ -208,30 +225,30 @@ export default Vue.extend({
...
 
         this.setCurrentTime({ time: currentTime });
       }
     },
-    onVideoPause(_event?: Event): void {
-      if (this.$refs.videoPlayer) {
-        const currentTime = (this.$refs.videoPlayer as HTMLVideoElement)
+    onPause(_event?: Event): void {
+      if (this.$refs.shakaPlayer) {
+        const currentTime = (this.$refs.shakaPlayer as HTMLVideoElement)
```
ThibaultNocchi, line 242:
```suggestion
        const currentTime = (this.$refs.shakaPlayer as HTMLMediaElement)
```

client/components/Players/ShakaPlayer.vue
```diff
@@ -208,30 +225,30 @@ export default Vue.extend({
...
         this.setCurrentTime({ time: currentTime });
         this.pause();
       }
     },
-    onVideoStopped(_event?: Event): void {
-      if (this.$refs.videoPlayer) {
-        const currentTime = (this.$refs.videoPlayer as HTMLVideoElement)
+    onStopped(_event?: Event): void {
+      if (this.$refs.shakaPlayer) {
+        const currentTime = (this.$refs.shakaPlayer as HTMLVideoElement)
```
ThibaultNocchi, line 251:
```suggestion
        const currentTime = (this.$refs.shakaPlayer as HTMLMediaElement)
```

client/components/Players/ShakaPlayer.vue
```diff
@@ -195,7 +212,7 @@ export default Vue.extend({
 
           const params = stringify(directOptions);
 
-          this.source = `${this.$axios.defaults.baseURL}/Videos/${mediaSource.Id}/stream.${mediaSource.Container}?${params}`;
+          this.source = `${this.$axios.defaults.baseURL}/${this.getCurrentlyPlayingMediaType}/${mediaSource.Id}/stream.${mediaSource.Container}?${params}`;
```
ferferga, line 215: According to what's above, this is missing an s at the end?

client/components/Players/ShakaPlayer.vue
```diff
@@ -39,19 +38,37 @@ export default Vue.extend({
...
-      return this.getImageInfo(this.getCurrentItem, { preferBackdrop: true });
+    poster(): ImageUrlInfo | string {
+      if (this.getCurrentlyPlayingMediaType === 'Video') {
+        return this.getImageInfo(this.getCurrentItem, { preferBackdrop: true });
+      } else {
+        return '';
+      }
+    },
+    mediaElement(): string {
+      if (this.getCurrentlyPlayingMediaType === 'Audio') {
```
ferferga, line 65: No way we can make this an enum?

ferferga (changes requested):

ferferga:
Also, instead of destroying and recreating the component with ``v-if``, maybe it's a good idea to keep it always alive, passing the audio/video string as a prop.

That way we can reuse the same shaka instance during all the time the client is alive + we can have better Chromecast support with Shaka. We can in a ``beforeUpdate`` hook unlink the Shaka instance from the DOM element and link it to the new one in a watcher that watches that prop.

MrTimscampi:
> we can have better Chromecast support with Shaka

For separation of concern, I would prefer mirroring what jf-web does and have a dedicated player component for Chromecast stuff.

Not only does it give us more control, but some behavior is also different between Chromecast/AirPlay and regular playback.

We also should avoid loading unnecessary DOM elements, since each one takes memory and processing, so I would prefer keeping them behind v-if (It also avoids us accidentally triggering stuff in multiple places, and prevents more reactive stuff from running all the time).

> We can in a beforeUpdate hook unlink the Shaka instance from the DOM element and link it to the new one in a watcher that watches that prop.

I don't think this will work like you expect it to. It's also prone to error. I would much prefer for each player instance to be self-contained and not mess with any other DOM element than its own.

client/components/Players/ShakaPlayer.vue
MrTimscampi: It'd have to be done server-side, since this essentially just gives you the media type from the currently playing item.

ferferga (changes requested):
Code changes look good now, however audio playback is choppy at the beginning now for some reason. Have you tested this properly?

neopc10:
I am also finding that the previous track button does not behave consistently. Sometimes it will go to the previous track, usually it will not.

ferferga:
@neopc10 Can you try now?

ferferga (approved):

neopc10:
> @neopc10 Can you try now?

LGTM

ThibaultNocchi (approved):

MrTimscampi:
@jellyfin-bot rebase

PR #1043, opened 2021-04-24
Lock body scroll when maximizing the player

MrTimscampi:
* Allows the scrollbar to hide and show without making the page jump
* Locks the body scroll when maximizing the video player dialog, hiding the scrollbar


ThibaultNocchi (commented):
LGTM

ThibaultNocchi (approved):

client/assets/global.scss
```diff
@@ -6,6 +6,12 @@
 @import '@fontsource/noto-sans-tc/index.css';
 @import '@fontsource/noto-sans-hk/index.css';
 
+html {
+  width: 100vw;
+  overflow-x: hidden;
+  overflow-y: auto;
```
ferferga, line 12: I still see the page jumping (pay attention to the top right icons). I did a quick test by changing ``overflow-y`` to scroll in Edge devtools and jumping is gone. Not thoroughly tested though.

ferferga (changes requested):

client/assets/global.scss
MrTimscampi: The app bar is the only part of the page that jumps slightly, most likely due to being fixed and sized in a specific way.

I don't think it's really a big deal, since the only thing that jump are the two buttons on the top right. So it's very small and unless you're really looking at that zone of the page (Which you wouldn't normally, since you've just clicked on the maximize button, so your eyesight is around there, or going back to the middle of the page) you won't see it.

It doesn't seem easily fixable either, from a quick look.

With the alternative being either still showing the scrollbar, even in full screen, or having all the page content jump around instead of only a couple of buttons in a zone nobody looks at, I think this is "good enough" (Unless you have an idea for a fix, of course)

ferferga: What happens with the solution I proposed?

MrTimscampi: The same thing happens when you lock the scroll: the bar disappears and the buttons at the top right jump a few pixels.

Although keeping the scrollbar always shown prevents it when going to a page that would hide the scroll bar, but with the solution above, only those buttons jump and we can at least get rid of the scrollbar when it's not necessary.

Though I don't mind, So I pushed a version with the CSS removed.

ferferga (approved):

ferferga:
@jellyfin-bot rebase

PR #1057, opened 2021-04-27
feat(modal): add generic modal component

MrTimscampi:


ferferga:
What is the difference between this and [vuetify-dialog](https://www.npmjs.com/package/vuetify-dialog)?

MrTimscampi:
> What is the difference between this and [vuetify-dialog](https://www.npmjs.com/package/vuetify-dialog)?

This isn't controlled programmatically, but extended through other dialogs and inserted in whatever templates we need.

I'm not too sure about vuetify-dialog, since it has very little users, is based on something that's marked as "in development, can change at any time" and uses a completely different way of building components than everything else in this project...

The idea of this is to make ModalDialog very generic, then extend it in more specific components, to give us full control over the look of our modals, while making it easy to create them and add them to existing code, for anybody. It's just regular Vue, no API to learn or anything. Just insert the component, bind a few things and you're done, essentially.

MrTimscampi:
Closing this, as it's probably a better idea to do this post-Vuetify 3.

PR #1061, opened 2021-04-27
fix(i18n): lang setting display name

s0urcelab:
fix zh-CN & zh-TW display name


ThibaultNocchi (approved):
Thank you for your help, I did the initial translations but as a non speaker it was hard ^^

nyanmisaka (approved):

MrTimscampi (approved):

PR #1062, opened 2021-04-27
fix(series-item): removed original title if it is the same as the name

MarcelCoding:
#890 -> was only for a film
This not also fixes this Issue on a series.

see https://github.com/jellyfin/jellyfin-vue/issues/836#issuecomment-804008091

### DIFF
--- a/client/pages/series/_itemId/index.vue
+++ b/client/pages/series/_itemId/index.vue
@@ -13,7 +13,7 @@
             {{ item.Name }}
           </h1>
           <h2
-            v-if="item.OriginalTitle"
+            v-if="item.OriginalTitle && item.OriginalTitle !== item.Name"
             class="text-subtitle-1"
             :class="{ 'text-center': !$vuetify.breakpoint.mdAndUp }"
           >


ferferga:
Maybe we can move this to the layout we share across all the pages instead of doing this in a per page basis?

@MrTimscampi @ThibaultNocchi @Camc314 @dkanada thoughts?

MrTimscampi:
> Maybe we can move this to the layout we share across all the pages instead of doing this in a per page basis?

Not really, because the layout is only for the columns. This check is per item type, since a bunch of them don't even have original titles.

The way the title is setup varies per item type, so imo that should stay in each template, to avoid falling into a jf-web situation where we have 25 lines to generate the title block and barely anyone understands what goes on in there.

ferferga:
Fair enough 👍🏻

ferferga (approved):

MrTimscampi (approved):

ThibaultNocchi (approved):

ferferga:
@jellyfin-bot rebase

Issue #888, opened 2021-03-16
Fullscreen Music Player


hamburglar2160:
When playing through to the last track of the queue, the height of the bottom bar for the fullscreen player decreases when there is no value to display under "next up. " For consistency, it probably shouldn't do that.

PR #1064, opened 2021-04-28
fix(fullscreen-player-next-up): show empty space if it is the last item

MarcelCoding:
Display an empty space at the location of the `UP NEXT` string to prevent the resizing of the toolbar.

#888


ThibaultNocchi (approved):
Tested and approved

ferferga (approved):

PR #1071, opened 2021-05-01
fix(store (item.ts)): if store.byId contains item.Id & has more keys,...

### LINKED ISSUE #1059: "item detail" page get incomplete data
**Describe The Bug**

 "item detail" page get incomplete data when redirect from "person" page.

**Steps To Reproduce**

1. Go to 'item' page
2. Click on any 'person'
3. Click any 'item' from this person
4. 'item' page get incomplete data

**Expected Behavior**

should get full item data

**Screenshots**

image

image

image

**System (please complete the following information):**

- Platform: Windows 10
- Browser: Chrome 78
- Jellyfin Version: 10.7.2

akkisagiraju:
… don't override its value

If ADD_ITEM is committed with an already set ID and the incoming item object has fewer keys than the
existing object, don't update it.

fix #1059

The issue in #1059 is, if a user clicks on a person/genre on an item (movie/tv show) page, the API request for that person/genre returns only partial data for every individual item inside the items array in the response and the existing state.ById value for those items is getting overridden.

This could be fixed in two ways: 1) Make changes in the server code to always return the entire object for every individual items, or 2) Inside the ADD_ITEM action, check if the state.ById contains the Id of the incoming item. If it does and has more keys than the incoming item object, don't update its value. I chose the second option.


akkisagiraju:
I just realized that this would cause the app to retain old metadata even after editing it since there is no comparison between the existing and the incoming item objects.

So, we either have to fix the APIs to always return full item objects for a person/genre or remove the following line to always fetch the latest and complete item data.

```
if (!store.getters['items/getItem'](itemId)) {
      await $userLibrary.getItem(itemId);
    }
```

MrTimscampi:
We can pretty much always specify which fields we want from the client's side.

Adding something like `fields: Object.values(ItemFields)` (With `ItemFields` being the enum listing the available fields, imported from the Axios API client) to the API calls should essentially get the full item each time. We could add it to itemsPlugin and userLibraryPlugins, like we do for the user ID, then omit it from the copied type.

As for:
```typescript
if (!store.getters['items/getItem'](itemId)) {
  await $userLibrary.getItem(itemId);
}
```
It's there because of the websocket.

Essentially, it works like this:

* When you first get an item, it is added to the store
* We have a websocket that sends us back a list of updated item IDs every time something changes server-side
* If we already have a copy of that item, we fetch an updated version and replace what we have
* If we don't, we ignore it

This *should* ensure that all items that were already fetched once are always the latest version of whatever is server-side (Do note that there is a delay of 30 seconds for the websocket message, as the server tries to batch multiple updates together, which is quite annoying).  
So, when we already have a copy of something, we just use that, since it's supposed to be fully up to date.

akkisagiraju:
> We can pretty much always specify which fields we want from the client's side.

I didn't realize we could do this. Thank you so much for clearing that up!

Specifying the fields of item in the getItems call inside itemsPlugin fixed the issue. Since it is a different kind of fix, I'll close this PR and open a new one.

PR #1087, opened 2021-05-06
refactor: v-hide directive and minor cleanup

ferferga:
* Added a ``v-hide`` directive that applies opacity instead of ``display: none``.
* Scope styles of PlayerManager
* Share more global classes

### DIFF
--- a/client/assets/global.scss
+++ b/client/assets/global.scss
@@ -83,8 +83,22 @@ body {
 
 /* Custom utility classes */
 
+.absolute-cover {
+  height: 100%;
+  width: 100%;
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+}
+
 .pointer {
   cursor: pointer;
+}
+
+.pointer-events-none {
+  pointer-events: none;
 }
 
 .link {
--- a/client/assets/global.scss
+++ b/client/assets/global.scss
@@ -101,6 +101,10 @@ body {
   pointer-events: none;
 }
 
+.all-pointer-events {
+  pointer-events: all;
+}
+
 .link {
   cursor: pointer;
   color: inherit !important;
--- a/client/assets/global.scss
+++ b/client/assets/global.scss
@@ -101,7 +101,7 @@ body {
   pointer-events: none;
 }
 
-.all-pointer-events {
+.pointer-events-all {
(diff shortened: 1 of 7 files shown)


client/layouts/default.vue
```diff
@@ -130,11 +127,21 @@ interface LayoutButton {
 
 export default Vue.extend({
   mixins: [settingsHelper],
+  props: {
+    keepAliveOptions: {
+      type: Object as () => Record<string, unknown>,
+      default: (): Record<string, unknown> => {
+        return {
+          max: 10,
+          exclude: ['fullscreen-playback']
```
ferferga, line 136: We should probably report the fact that this is not working upstream

ThibaultNocchi: What's not working?

ThibaultNocchi (commented):
I think you could use Vuetify's `.d-none` instead of adding a Vue directive, wdyt?

ferferga:
d-none and v-show are the same thing, display: none does omit the element in the layout calculations, while ``opacity 0`` and ``visibility hidden`` does not.

client/layouts/default.vue
ferferga: The page is not being excluded from the keep-alive. According to Nuxt docs I think our approach for excluding it is right, but it doesn't work.

ThibaultNocchi: I guess you could at least bring that to them yeah

client/assets/global.scss
```diff
@@ -83,10 +83,24 @@ body {
...
+  left: 0;
+  right: 0;
+  bottom: 0;
+}
+
 .pointer {
   cursor: pointer;
 }
 
+.pointer-events-none {
```
dkanada, line 100: The `pointer-events-none` and `all-pointer-events` classes should match, and also might benefit from being in the same location.

client/assets/global.scss
```diff
@@ -83,10 +83,28 @@ body {
...
+  pointer-events: none;
+}
+
+.all-pointer-events {
```
dkanada, line 104:
```suggestion
.pointer-events-all {
```

dkanada (approved):

MrTimscampi (approved):

PR #1137, opened 2021-05-23
Update to Node 16

MrTimscampi:
This moves the manifest to NPM's v2 manifest version (used by NPM 7) and moves our CI workflows to use Node 16.

Both NPM 7 and Node 16 have already received a few updates, and it all works locally (plus I need Node 16 for Apple M1 support), so this feels like a good time to update.


ThibaultNocchi:
Don't we gotta change Node versions in /Dockerfile, /Dockerfile.static, /README.md, .devcontainer/Dockerfile, .github/workflows/Dockerfile.ssr?

MrTimscampi:
Not actually needed to work with jf-vue on an M1 Mac, there are ways around the issues I was having, so closing this to wait a bit more, as initially planned.

PR #1138, opened 2021-05-23
fix(relateditems): related items excludes albums from the same artist

n-buck:


ferferga:
Why?

n-buck:
> Why?

In my library, I will be presented with only albums of this interpret of the album (if there are enough albums), which is kind of missing the point for me.
I can get a list of the albums of the interpret in the artist-page already.

Also it is excluded in the current Jellyfin-GUI.

n-buck:
Oh, sorry I just realized that I introduced a bug.

client/components/Item/RelatedItems.vue
```diff
@@ -109,12 +109,19 @@ export default Vue.extend({
...
 
+      const options = {
+        itemId: this.item.Id,
+        userId: this.$auth.user?.Id,
+        limit: this.vertical ? 5 : 12,
+        excludeArtistIds: [] as string[]
+      };
+
+      for (const albumArtist of this.item.AlbumArtists) {
+        options.excludeArtistIds.push(albumArtist.Id);
```
MrTimscampi, line 120: If this is to be done, it should be on the server-side.

Doing it on the client hampers consistency across the various clients. We should strive to return the same results everywhere, not to manipulated them all the time to get what we want.

MrTimscampi (changes requested):
This also completely breaks type inference for the options array, as is.

client/components/Item/RelatedItems.vue
dkanada: I would be fine adding this with a `TODO` or `SERVER` comment. We definitely want to do things server-side when possible, but for smaller items it can be easier to implement within clients until the server has support, especially since we are striving towards a more "stable" server API in the future.

MrTimscampi: This would be very easy to do on the server though, provided we're okay with slightly changing what is returned from it (It won't break clients, just deprecate a field and return different results).

n-buck:
> This also completely breaks type inference for the options array, as is.

This should be ok now.

MrTimscampi:
> > This also completely breaks type inference for the options array, as is.
> 
> This should be ok now.

What I meant is that there is a type (`LibraryApiGetSimilarItemsRequest`) which has all the properties already defined :) Defining an object without assigning that type to it breaks the typings, since we can't see all the other properties and it doesn't get properly checked against the API client :)

client/components/Item/RelatedItems.vue
```diff
@@ -109,12 +110,25 @@ export default Vue.extend({
     async refreshItems(): Promise<void> {
       this.loading = true;
 
+      let excludedArtistsIds;
```
ferferga, line 113: This can be null at the moment of performing the request

```suggestion
      let excludeArtistsIds = [];
```

client/components/Item/RelatedItems.vue
```diff
@@ -55,6 +55,7 @@
 import Vue from 'vue';
 import { mapActions } from 'vuex';
 import { BaseItemDto } from '@jellyfin/client-axios';
+import { LibraryApiGetSimilarItemsRequest } from '@jellyfin/client-axios/api/library-api';
```
ferferga, line 58:
```suggestion
import { LibraryApiGetSimilarItemsRequest } from '@jellyfin/client-axios';
```

client/components/Item/RelatedItems.vue
```diff
@@ -109,12 +110,25 @@ export default Vue.extend({
     async refreshItems(): Promise<void> {
       this.loading = true;
 
+      let excludedArtistsIds;
+
+      if (this.item.AlbumArtists) {
+        excludedArtistsIds = this.item.AlbumArtists.map(
```
ferferga, line 116:
```suggestion
        excludeArtistsIds = this.item.AlbumArtists.map(
```

client/components/Item/RelatedItems.vue
```diff
@@ -109,12 +110,25 @@ export default Vue.extend({
...
+          (albumArtist: { Id: string }) => albumArtist.Id
+        );
+      }
+
+      const requestParameters: LibraryApiGetSimilarItemsRequest = {
+        itemId: this.item.Id,
+        userId: this.$auth.user?.Id,
+        limit: this.vertical ? 5 : 12,
+        excludeArtistIds: excludedArtistsIds
```
ferferga, line 125:
```suggestion
        excludeArtistIds
```

ferferga (changes requested):
This is non blocking, but I also think the logic of this can be much more simpler without the additional variables. Something along the lines of:

```
let excludedArtistsParams = {};

      if (this.item.AlbumArtists) {
        excludedArtistsParams = { excludeArtistsIds: this.item.AlbumArtists.map(
          (albumArtist: { Id: string }) => albumArtist.Id
        ) };
      }

      if (this.item.Id) {
        const response = await this.$api.library.getSimilarItems(
          {
        itemId: this.item.Id,
        userId: this.$auth.user?.Id,
        limit: this.vertical ? 5 : 12,
        ...excludedArtistsParams
      }
        );
```

client/components/Item/RelatedItems.vue
```diff
@@ -109,12 +110,25 @@ export default Vue.extend({
...
 
+      let excludeArtistsIds = [];
+
+      if (this.item.AlbumArtists) {
+        excludeArtistsIds = this.item.AlbumArtists.map(
+          (albumArtist: { Id: string }) => albumArtist.Id
+        );
+      }
+
+      const requestParameters: LibraryApiGetSimilarItemsRequest = {
```
MrTimscampi, line 121: This could be directly inside `getSimilarItems` like it was before.

Currently, it's also outside of the condition on `this.item.Id`, which kind of negates the purpose of that condition and could lead to issues.

MrTimscampi:
Sorry, I seemed to have messed up the pull request when pushing to it (I hate having to deal with detached heads....)

The new PR is #1355

PR #1159, opened 2021-05-29
fix(shaka-player): fix HLS playback

MrTimscampi:
We were importing Shaka before setting mux.js as a global. As such, Shaka couldn't find it when
initializing and didn't enable support for transmuxing HLS on non-Apple browsers.


ferferga (approved):

dkanada (approved):

ThibaultNocchi (approved):
I think it's a good place to let a comment in the code ^^

PR #1163, opened 2021-05-30
feat(video-osd): add playback information

MrTimscampi:
Adds some playback information for debugging purposes.

Screenshot from 2021-05-31 13-05-46

### DIFF
--- a/client/components/Players/PlaybackInfoCard.vue
+++ b/client/components/Players/PlaybackInfoCard.vue
@@ -344,6 +344,7 @@ export default Vue.extend({
     mediaTotalBitrate(): string | null | undefined {
       if (
         this.sessionInfo?.TranscodingInfo?.Bitrate &&
+        this.currentMediaSource?.Bitrate &&
         this.sessionInfo.TranscodingInfo.Bitrate !==
           this.currentMediaSource.Bitrate
       ) {
@@ -352,9 +353,11 @@ export default Vue.extend({
         )} ➞ ${this.getDisplayBitrate(
           this.sessionInfo?.TranscodingInfo?.Bitrate
         )}`;
+      } else if (this.currentMediaSource?.Bitrate) {
+        return this.getDisplayBitrate(this.currentMediaSource.Bitrate);
       }
 
-      return this.getDisplayBitrate(this.currentMediaSource.Bitrate);
+      return null;
     }
   },
   beforeMount() {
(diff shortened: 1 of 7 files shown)


neopc10:
image

Also, an X button would be helpful here

MrTimscampi:
> Also, an X button would be helpful here

That's why it's still in draft ;) It was fairly late yesterday evening, so I put it in draft :smile:

dkanada:
1. Transcoding -> Transcode
2. Black -> Translucent Purple

I think dialogs on the playback page might benefit from a small amount of opacity. At minimum though, I'd rather avoid black and switch to the default purple for this modal.

MrTimscampi:
Added the "Close" button, removed the black color, changed a few strings and fixed an issue where transcoded video would be shown as not being transcoded.

@dkanada I'd rather not use transparency for this, as I find the use of transparency annoying in jf-web for what should be easily readable debugging information. I swapped out the color though.

Artiume:
the screen shot shows direct and hls. i thought we didn't use hls for direct play?

MrTimscampi:
> the screen shot shows direct and hls. i thought we didn't use hls for direct play?

It's something with Vue and refreshing the dev player ;) It looks as the session to see what it needs to use, but it flips out a bit due to the Vue dev server only refreshing parts of the app, so disregard the erroneous display in the screen.

As for the HLS bit, here's how it works: if the source is a `.m3u8` file, it's HLS. If the source is an `.mpd` file, it's DASH (Future-proofing, here), if it's something else, it's "Video".

Here, it's a transcoded file, so HLS.

client/components/Players/PlaybackInfoCard.vue
```diff
@@ -0,0 +1,417 @@
...
+    ...mapState('playbackManager', ['currentMediaSource']),
+    ...mapGetters('playbackManager', [
+      'getCurrentVideoTrack',
+      'getCurrentAudioTrack',
+      'getCurrentSubtitleTrack'
+    ]),
+    isTranscoding(): boolean {
+      return !!(
+        this.sessionInfo?.PlayState?.PlayMethod === 'Transcode' ||
+        this.sessionInfo?.TranscodingInfo
```
ThibaultNocchi, line 260: Wouldn't the or and equality check already return a boolean?

MrTimscampi: Apparently not, TypeScript was complaining about TranscodingInfo returning an object, hence the `!!` here :)

client/components/Players/ShakaPlayer.vue
```diff
@@ -101,7 +102,9 @@ export default Vue.extend({
       shaka.polyfill.installAll();
 
       if (shaka.Player.isBrowserSupported()) {
-        this.player = new shaka.Player(this.$refs.shakaPlayer);
+        // We use a global for ease of debugging and to fetch data from the playback information popup
+        window.player = new shaka.Player(this.$refs.shakaPlayer);
```
ferferga, line 106: We probably want to have this settled up as undefined in beforeDestroy, like we do with muxjs

ThibaultNocchi (approved):

PR #1187, opened 2021-06-03
Refactors and minor improvements

ferferga:
* Use v-cloak at the template level to avoid scenarios where mustache tags could appear because the template compilation is not going fast enough
* Move vuetify and simple-icons to devdependencies
* Remove nuxt-vuex-localstorage and use vuex-persistedstate only
* Use Cookies + localStorage (depending on the data that must be present at the server level, like auth, leaving the rest of the stuff in localStorage in the client-side. See [this description](https://github.com/jellyfin/jellyfin-vue/pull/1187/files#diff-beeff111fb09d2c3014b2de578b9fc9c0e3271aad1ec2f1f792cc260c2813bf9R11-R25)) only when running in SSR mode and use localStorage only when running in SPA mode (this way we avoid cookie pollution in the request headers)

*This also means that we can work in scenarios where cookies are disabled completely, as long as SSR is not being used*

* Kill the whole page layout as soon as user is undefined, so there's no popping anywhere anymore.
* Refactor the login schema to work better with the Cookie/localStorage mix we're doing now by using Vuex all the time.
* Fix lint warnings.
* Make Vue updatable by ourselves, instead of relying in nuxt for upstream updates. Supersedes #1179


client/schemes/jellyfinScheme.ts
```diff
@@ -20,33 +21,46 @@ export default class JellyfinScheme {
...
   }
 
   _getRememberMe(): boolean {
-    return this.$auth.$storage.getUniversal('rememberMe');
+    return destr(this.$auth.$storage.getState('rememberMe'));
   }
 
-  mounted(): Promise<never> {
-    const token = this.$auth.syncToken(this.name);
+  async mount(): Promise<void> {
```
MrTimscampi, line 50: Having both `mount` and `mounted` is confusing.

What you're doing here is stuff that would fit in other, standard, Nuxt Auth methods.

client/schemes/jellyfinScheme.ts
```diff
@@ -59,20 +73,21 @@ export default class JellyfinScheme {
     rememberMe: boolean;
   }): Promise<AxiosResponse<AuthenticationResult> | void> {
     // Ditch any leftover local tokens before attempting to log in
-    await this.$auth.reset();
+    await this.$auth.logout();
```
MrTimscampi, line 76: `reset()` is how it should be done: https://github.com/nuxt-community/auth-module/blob/master/lib/schemes/local.js#L38

client/schemes/jellyfinScheme.ts
```diff
@@ -82,38 +97,35 @@ export default class JellyfinScheme {
...
-  setUserToken(tokenValue: string): Promise<void> {
+  async setUserToken(tokenValue: string): Promise<void> {
     const token = `MediaBrowser Client="${this.$auth.ctx.app.store.state.deviceProfile.clientName}", Device="${this.$auth.ctx.app.store.state.deviceProfile.deviceName}", DeviceId="${this.$auth.ctx.app.store.state.deviceProfile.deviceId}", Version="${this.$auth.ctx.app.store.state.deviceProfile.clientVersion}", Token="${tokenValue}"`;
 
     this.$auth.setToken(this.name, token);
     this._setToken(token);
 
-    return this.fetchUser();
+    if (tokenValue) {
+      await this.fetchUser();
```
MrTimscampi, line 127: This absolutely doesn't fit in something called `setUserToken`.

client/schemes/jellyfinScheme.ts
```diff
@@ -153,16 +165,14 @@ export default class JellyfinScheme {
     }
 
     // Reset everything
-    return this.$auth.reset();
+    this.$auth.reset();
   }
 
-  reset(): Promise<void> {
```
MrTimscampi, line 159: Don't change the standard stuff used by Nuxt Auth, please.

If it returns a Promise in the local scheme ([which it does](https://github.com/nuxt-community/auth-module/blob/master/lib/schemes/local.js#L114) it should return one here.

That makes it easier to reference back to the original and update it when/if necessary.

package.json
```diff
@@ -130,11 +127,15 @@
...
     "lint-staged": "^11.0.0",
     "prettier": "^2.3.0",
+    "simple-icons": "^5.0.0",
     "stylelint": "^13.13.1",
     "stylelint-config-prettier": "^8.0.2",
     "stylelint-config-standard": "^22.0.0",
     "ts-jest": "^26.5.6",
     "ts-loader": "^9.0.0",
+    "vue-template-compiler": "^2.6.13",
+    "vue-server-renderer": "^2.6.13",
```
MrTimscampi, line 137: We should keep depending on Nuxt for this, to align our used version with theirs.

If anything, we should remove the `vue` package, which was only added to placate SonarCloud, and let Nuxt handle all Vue versioning, as is intended.

MrTimscampi (changes requested):
I don't agree with a bunch of the choices here, even if it works in practice.

We should stay as close as possible to upstream for the scheme, and using localStorage instead of cookies for authentication is not a good thing to do.

client/plugins/store/persistedStatePlugin.ts
```diff
@@ -0,0 +1,68 @@
...
+   * process.server or process.browser lets us distinguish between both environments when running SSR,
+   * but the check client-side just lets us know that we're in the browser environment, without further
+   * information on whether browser there's SSR behind the scenes.
+   *
+   * A hardcoded variable provides us that information, which helps us to manage the Vuex persisted storage properly
+   * by saving all the store's information in localStorage when the client is fully SPA and combining cookies and localStorage
+   * when using SSR, as SSR rendering needs the server to have access to the authentication data so the redirection
+   * middleware acts accordingly.
+   *
+   * Cookies pollutes a lo the headers of the clients and are problematic privacy-wise, so it's a good idea to avoid using them
```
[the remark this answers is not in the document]
ferferga: The thing is that they're not needed at all when client is built without SSR, so we're sending a bunch of data in each request done (yes, the cookie header gets long when you add multipe servers), even being unnecessary in a client-side scenario, so we're wasting resources unnecessarily when the client is built without SSR.

With this approach we keep the same thing we've been doing now for SSR (with cookies for auth stuff) without affecting SPA mode (being localStorage-only now), as the cookies were implemented on the SSR switch because server-side needs to know the auth details to run the redirection middleware

There are also some stuff I don't agree with the article:

* Having cookies available to subdomains can introduce inconsistencies between what we have in the store and the user that logged in. For instance, changing subdomains will still keep the user logged in but displaypreferences, homesections (once they're customizable) will be different because the mismatch between both. Having everything in Vuex ensures a consistent state
* Expiration should be handled server-side and not client-side in my opinion. If we want to have expirations, that should be controlled server-side, and not have our own limits, this can be confusing for users.
* We don't need to manage anything from localStorage as that's already handled by Nuxt Auth using Vuex. Nuxt Auth already invalidates everything as soon as it detects the session was revoked server side.


Of course all this stuff doesn't apply to clients built in SSR which will require cookies, that won't change. All I'm talking about is SPA mode of course.

client/plugins/store/persistedStatePlugin.ts
```diff
@@ -0,0 +1,68 @@
...
+   * as long as they're not necessary, hence all of this.
+   */
+  const isSSR = destr(process?.env?.NUXT_SSR);
```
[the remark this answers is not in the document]
ferferga: @MrTimscampi It's not the same thing.

Nuxt plugins runs both in server and client on SSR scenarios and in client only when in SPA. ``process.server`` lets us add specific code for each of these environment, but when the plugin runs client-side it doesn't know anything about the SSR part. 

By injecting the environment variable at build time, this check allows us to basically distinguish if the client has been built with SSR or SPA to change between cookies + localStorage or localStorage only when in SPA mode.

client/schemes/jellyfinScheme.ts
```diff
@@ -20,33 +21,46 @@ export default class JellyfinScheme {
     this.$auth = auth;
 
     this.options = Object.assign({}, {}, options);
+    this.mount();
```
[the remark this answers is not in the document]
ferferga: Sorry, some testing leftovers!

client/schemes/jellyfinScheme.ts
```diff
@@ -20,33 +21,46 @@ export default class JellyfinScheme {
...
+
   _clearToken(): void {
     // Clear Authorization token for all axios requests
     this.$auth.ctx.app.$axios.setHeader('X-Emby-Authorization', false);
   }
 
   _setRememberMe(value: boolean): void {
     // Sets the remember me key which is used to relogin someone
-    this.$auth.$storage.setUniversal('rememberMe', value);
+    this.$auth.$storage.setState('rememberMe', value);
```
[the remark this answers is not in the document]
ferferga: As explained above, imo we should have everything handled by Vuex only to keep consistency and handle the persistence using ``vuex-persistedstate``, which allows us to have complete control of how data is stored and consistency if we ever need to access or subscribe to mutations in the accessToken or servers, for instance.

Also, saving to Vuex before was useless as we didn't set the ``auth`` namespace as one of the persisted ones.

client/schemes/jellyfinScheme.ts
```diff
@@ -20,33 +21,46 @@ export default class JellyfinScheme {
...
   }
 
   _getRememberMe(): boolean {
-    return this.$auth.$storage.getUniversal('rememberMe');
+    return destr(this.$auth.$storage.getState('rememberMe'));
```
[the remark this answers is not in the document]
ferferga: Explained reasoning above

client/schemes/jellyfinScheme.ts
ferferga: As said above, this is testing bits, will remove.

client/schemes/jellyfinScheme.ts
ferferga: Calling logout does the same thing + sends a call to server to revoke the old session, which is something reset doesn't do and I think it's nice to have.

package.json
ferferga: Seems reasonable. I didn't know why the vue package was there in the first place and that's why I took the updating approach, but I agree with you 👍.

client/plugins/store/persistedStatePlugin.ts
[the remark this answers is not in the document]
MrTimscampi: > Having cookies available to subdomains can introduce inconsistencies between what we have in the store and the user that logged in.

You are misunderstanding how the cross-subdomain thing works, here. Per the [spec](https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.3):

```
5.1.3.  Domain Matching

   A string domain-matches a given domain string if at least one of the
   following conditions hold:

   o  The domain string and the string are identical.  (Note that both
      the domain string and the string will have been canonicalized to
      lower case at this point.)

   o  All of the following conditions hold:

      *  The domain string is a suffix of the string.

      *  The last character of the string that is not included in the
         domain string is a %x2E (".") character.

      *  The string is a host name (i.e., not an IP address).
```

So a cookie set for `vue.example.com` will only be readable by `vue.example.com`, while `.example.com` will be readable by any subdomain of `example.com`, and `example.com` is readable only by `example.com`.

> Expiration should be handled server-side and not client-side in my opinion.

I really disagree with this.

What happens if I login at a friend's house and forget to disconnect? They'll have access to the session pretty much forever.

The point of expiration is that the client can "timeout" if the cookie isn't used for a while. So if I don't use my session for a few days, it will actively ask me to re-login, which your localStorage way won't do, since it only expires if the session on the server does. There's a reason why, even with localStorage available for years, everybody still uses cookies to store sessions (Except Emby and jf-web).

> We don't need to manage anything from localStorage as that's already handled by Nuxt Auth using Vuex.

And we don't need to manage cookies either, except parsing them on the server (which you need to do anyway). I fail to see how this changes anything?

---

Point is:

* We should use cookies for auth stuff
* And use localStorage for non-auth stuff (Saved servers, for example)

client/plugins/store/persistedStatePlugin.ts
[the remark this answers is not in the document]
MrTimscampi: But why should the persisted stores change if we're in SPA or SSR? That makes no sense and only adds complexity for no good reason.

client/schemes/jellyfinScheme.ts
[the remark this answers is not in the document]
MrTimscampi: I disagree, we shouldn't circumvent how Nuxt Auth works just because we persist some other stuff.

Then, we may as well completely remove Nuxt Auth and do everything ourselves, if we follow your logic...

Nuxt Auth still uses the store, but it manages it itself, which is what it's expected to do. Again, this adds complexity to the implementation for no reason, and makes it harder for someone coming in to understand what the hell is happening and why, while letting Nuxt Auth do its thing and staying as close as possible to the upstream schemes makes it pretty easy to simply go there and look at docs.

client/schemes/jellyfinScheme.ts
MrTimscampi: The point of `login` is not to log you out of the server, though. It's only meant to, well, log you in.

It'll never be called when there is a user logged in anyway.

PR #1200, opened 2021-06-07
refactor(pages/series): use tab-based layout for series pages

MrTimscampi:
Brings the new tabbed interface from the People and Artist pages to Series pages.

image

* Seasons now show their watched status and the year they aired, providing more info at a glance
* The first unwatched season is now shown by default, reducing the amount of clicks needed to get to the content the user wants to see
* "You may also like" is now in the same position as on Artist and People pages, for consistency

image

The Information tab provides a few more info about the series, such as links, runtime, premiere and finale date, rating and the network associated with the series.

**Cast & crew tab is still work in progress**

### DIFF
--- a/client/assets/global.scss
+++ b/client/assets/global.scss
@@ -148,3 +148,9 @@ body {
 .text-capitalize-first-letter::first-letter {
   text-transform: uppercase;
 }
+
+/* Global Jellyfin styles */
+
+.item-overview {
+  text-align: justify;
+}
--- a/client/components/Item/Card/Card.vue
+++ b/client/components/Item/Card/Card.vue
@@ -164,7 +164,7 @@ export default Vue.extend({
      * @returns {string} Either the item name or the series name
      */
     cardTitle(): string {
-      if (this.item.Type !== 'Episode') {
+      if (this.item.Type !== 'Episode' || this.episode) {
         return this.item.Name || '';
       } else {
         return this.item.SeriesName || '';
@@ -176,6 +176,10 @@ export default Vue.extend({
      *                   or the album artist
      */
     cardSubtitle(): string {
+      if (this.episode) {
+        return '';
+      }
+
       switch (this.item.Type) {
         case 'Episode':
           return `${this.$t('seasonEpisodeAbbrev', {
--- a/client/components/Item/ItemGrid.vue
+++ b/client/components/Item/ItemGrid.vue
@@ -24,6 +24,7 @@
             v-for="card of item.chunk"
             :key="card.Id"
             :item="card"
+            :episode="episode"
             margin
             text
             overlay
@@ -62,12 +63,13 @@ export default Vue.extend({
       }
     },
     loading: {
-      type: Boolean,
-      required: false
+      type: Boolean
+    },
+    episode: {
+      type: Boolean
     },
     large: {
-      type: Boolean,
-      required: false
+      type: Boolean
     }
   },
   computed: {
--- a/client/components/Item/MediaInfo.vue
+++ b/client/components/Item/MediaInfo.vue
@@ -1,7 +1,11 @@
 <template>
   <div class="text--secondary">
-    <span v-if="item.ProductionYear && year">{{ item.ProductionYear }}</span>
-    <span v-if="item.OfficialRating && rating">{{ item.OfficialRating }}</span>
+    <span v-if="productionYear && year">{{ productionYear }}</span>
+    <span v-if="item.OfficialRating && rating">
+      <v-chip class="text-overline" small label>
+        {{ item.OfficialRating }}
+      </v-chip>
+    </span>
     <span v-if="item.CommunityRating && rating">
       <v-icon class="rating-icon" size="16">mdi-star</v-icon>
       {{ item.CommunityRating }}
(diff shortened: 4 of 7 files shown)


neopc10:
Can we hide the tags into the information tab or, better yet, kill them with fire?
The synopsis is super useful and something I really want to have available at first glance, not tucked away in another tab. I'd prefer  to see that back where the tags currently are (and on master).

>>Seasons now show their watched status and the year they aired, providing more info at a glance
This would be a great spot to have a clickable checkmark to set a season as watched or unwatched. I find that card very useful on jellyfin-web. How would a user edit the season metadata? Maybe hovering over that checkmark exposes an overflow menu which has a "mark as watched" entry...

The external links are difficult to see in several scenarios with the blue hyperlink. Presenting them like the tags or genres may be viable.

I think the "airs on"/"studio" currently in the information tab might be better served in place of the "ends at:" on the very top of the page.

Perseusdehond:
> This would be a great spot to have a clickable checkmark to set a season as watched or unwatched. I find that card very useful on jellyfin-web. How would a user edit the season metadata? Maybe hovering over that checkmark exposes an overflow menu which has a "mark as watched" entry...
> 
I think that maybe a button could just be put above the episode cards so you can change things for every season. Or you could make every seasons metadata changeable within the metadata menu of the series that is already present. So you would have extra options/tabs to switch to a different season (although I don't know if this is the best user experience)

> I think it'd be a pity not to use season posters. Can they load in place of the series poster when the user navigates to different seasons?
> 
I would maybe just prefer to have an (new) image tab that shows you background images and season posters for consistency in the look.

MrTimscampi:
> Can we hide the tags into the information tab or, better yet, kill them with fire?

They only show up if you have tags on an item.

Long-term, the plan is to make them searcheable/browseable, like genres. Them being always there is due to a change in some of the providers, so is out of this client's control.

> The synopsis is super useful and something I really want to have available at first glance, not tucked away in another tab. I'd prefer to see that back where the tags currently are (and on master).

The main issue, and the reason why we're adopting this way of doing this, is that we don't control the length of the synopsis.

Ideally, we would have a "Short synopsis" field on items, but currently it's not the case. So they tend to vary a LOT depending on which providers you have. They can go from one paragraph to novel-size, which breaks the entire layout.

Once the database rewrite is done, I'd like to introduce a shorter field in the server, which we could use without worrying about its length, and use the synopsis field for really long, in-detail, descriptions of content.

However, for now, this is the best we can do.

> This would be a great spot to have a clickable checkmark to set a season as watched or unwatched. I find that card very useful on jellyfin-web. How would a user edit the season metadata? Maybe hovering over that checkmark exposes an overflow menu which has a "mark as watched" entry...

We can't really do tabs with a menu in them, since tabs are already clickable. It kind of runs counter to all the other patterns in the client.

There are a few ways to provide the same functionality, but I need to think a bit more about it.  
What we may do is improve the metadata editor for the series, in order to allow you to edit some of its children. For example, a series would allow you to edit the seasons to some extent.

The watch mark for seasons is currently set based on the children (the episodes). Multi-select can, in the future, make it possible to make a season as watched by allowing you to select all episodes in the season, then mark them as played. But that's for the future :)

> I think it'd be a pity not to use season posters. Can they load in place of the series poster when the user navigates to different seasons?

That's not really a thing I want to do, because for it to make sense, you'd have to select the first season manually, which means at least one click to get to your episode.

I'm generally in favor of getting rid of a lot of artworks, and season posters is one of the artwork types which don't make much sense to me.

We might reconsider in the future and make it a toggle-able option, but for not the plan is to only have this tabbed view and ditch the season artwork entirely :smiley: 

> The external links are difficult to see in several scenarios with the blue hyperlink. Presenting them like the tags or genres may be viable.

Agreed there. I'm exploring other solutions for how to surface links in a better, which would then be spread to other pages.

The current alternative option is this:
image

But I need to play a bit more with it to make clear that you can click on the icons. Then it'd be some work to build a few helper methods to take a link and return the icon and such, figure out a decent fallback for domains without icons, etc.

> I think the "airs on"/"studio" currently in the information tab might be better served in place of the "ends at:" on the very top of the page.

That's a decent suggestion for series, indeed. Ends at doesn't work anyway for shows, so we may as well replace it there.

> I would maybe just prefer to have an (new) image tab that shows you background images and season posters for consistency in the look.

In my Grand Vision™ (:smile:) of having everything inter-linked, I'd like to be able to tag shows on images and such, so I do eventually see an image tab popping up there :wink:  

My general idea is for you to be able to find anything related to an item, be it actors, crews (including "deep cuts" like the costume department, riggers, stunt coordinators, etc), the soundtrack(s), production photos, etc.

Of course, we're *very* far from that currently, but that's the dream, for me, at least :yum:

Perseusdehond:
> The current alternative option is this:
> image
> 
I like the icons more if possible ofcourse 

> In my Grand Vision™ (😄) of having everything inter-linked, I'd like to be able to tag shows on images and such, so I do eventually see an image tab popping up there 😉
> 
> My general idea is for you to be able to find anything related to an item, be it actors, crews (including "deep cuts" like the costume department, riggers, stunt coordinators, etc), the soundtrack(s), production photos, etc.
> 
That would be really cool, but for now I meant just a simple (extra) tab that shows the artwork already present on the server (between Cast& Crew and Information e.g.). In that gallery sort of view with all the images you have available of that series.

ThibaultNocchi:
> I'm generally in favor of getting rid of a lot of artworks, and season posters is one of the artwork types which don't make much sense to me.

I agree, when you go down the series rabbit hole, season posters are severally lacking. I prefer having one good quality artwork visible than a multitude of shitty ones :p

lxqueen:
Will the season sidebar tabs still support custom season names? Because I'm not sure if that's enough room if so. Otherwise I love the layout.

ThibaultNocchi:
Haven't checked the code, but the column can be auto sized in Vue.

PR #1205, opened 2021-06-08
feat(playback): add track selection to details pages

MrTimscampi:
Adds pre-playback stream selection.

TrackSelector needs a refactor for this to be a bit cleaner, as I currently have to work around it to fit what Jellyfin expects to get (This PR already has some changes, but it's only workarounds... The whole "clearing" part doesn't work for this, passing it an entire item and a type doesn't work (filtering should be done above), etc.

Supersedes #640


ThibaultNocchi (approved):

MrTimscampi:
@jellyfin-bot rebase

PR #1352, opened 2021-07-18
fix(playback): fix endpoint spamming during playback

### LINKED ISSUE #1351: Progress Endpoint is spammed
**Describe The Bug**
Progress Endpoint is spammed nearly very second

**Steps To Reproduce**
1. Play a media
2. Have a look in the Network tab in the Dev Tools

**Expected Behavior**
Only execute the Progress endpoint (e.g.) every 10 seconds. (like the build in frontend)

**Logs**
image

MrTimscampi:
The reporting and playback data were spamming the network with requests every seconds, even when the playback data wasn't shown.

This adjusts the timers to 10 seconds for both, and prevents the playback data update from running when the dialog is not shown.

fix #1351


client/components/Players/UpNext.vue
```diff
@@ -149,3 +149,4 @@ export default Vue.extend({
   z-index: 9999;
 }
 </style>
+const now = new Date().getTime();
```
ThibaultNocchi, line 152: I don't see the usage of this in the diff, am I missing something?

MrTimscampi: Some VSCode plugin adding shit in my files, I assume... (Either Vetur or Volar, I would guess)

ThibaultNocchi (approved):

PR #1445, opened 2021-08-18
feat(seasontabs): add episode numbering

### LINKED ISSUE #1442: Show Episode number in Seasons view
### Discussed in https://github.com/jellyfin/jellyfin-vue/discussions/1441

<div type='discussions-op-text'>

<sup>Originally posted by **Pfuenzle** August 16, 2021</sup>
Currently, when looking at all the Episodes of a season, it's only showing the episode name and a description, but not an Episode number. This gets more and more confusing, the more episodes a season has, as you have to count the episodes to get the one you want to watch. 
Is an episode counter already on the To-do list?
I'm not experienced with Vue, but I think adding an counter should not be that extensive</div>

ThibaultNocchi:
fix #1442


neopc10:
Can we get rid of the episode numbering for the "specials" season? Movies for TV shows tend to be featured there and tend to be a bit of an eyesore on jf-web when a number is appended to the movie title.

MrTimscampi:
This will also conflict with #1200 since that uses cards.

PR #1459, opened 2021-08-22
chore: rename client dir to src

MrTimscampi:
The original idea was to have wrappers in this directory.

Since this isn't going to be the case anymore, having a `client` directory doesn't make much (if any) sense.

As such, this renames the `client` directory to `src` and updates references where appropriate.


MrTimscampi:
Note: needs an update to the pages.dev configuration.

ThibaultNocchi (approved):

PR #1476, opened 2021-08-28
feat(home): remove library tiles and add on deck

MrTimscampi:
* Removes the library tiles, since they show up in the sidebar anyway
* Replaces "Next up" and "Continue watching/listening" with "On deck", which shows every content the user has started (It merges both of the old sections)
* Renames "Latest {library}" to "Recently added in {library}", since these sections are not media types but library names, so this should be easier to translate in other languages and make sense in more use cases.

image


neopc10:
I was never a fan of the library tiles so 86ing them is a welcome change, but I think "Recently added to..." should link to their respective libraries. I find myself clicking them out of habit expecting to navigate but to no avail.

MrTimscampi:
> I think "Recently added to..." should link to their respective libraries. I find myself clicking them out of habit expecting to navigate but to no avail.

You're right, it was originally planned that way, but somehow we never ended up adding it here.

PR #1478, opened 2021-08-28
Re-design some areas of the client

MrTimscampi:
**WIP**

Coming in with a fresh look and a rested mind after a few months of less activity, it's fairly obvious that some design changes need to be made.

This PR aims to do just that: rework parts I feel can be improved, make the design make sense and tighten everything up.

## Development and production icons

Not really a design change per-se, but the client will now use the official Jellyfin logo colors when running in production.

Only development builds will show the classic Vue-green icons.

## General design

Along with the color updates made recently, the general layout needs a bit of improvement.

We're cleaning up some code around the navigation drawer, as well as changing the position of a few elements.

image
image


Extarys:
Were there discussions about having an option to have a dedicated drawer when in a Music library? It could display a list of playlists, and the obvious artists, albums, favorite artists (like followed artists on Spotify for example). This could also be displayed when the main drawer is collapsed, or we could add a tab above the current drawer to display such info.

I for one would like to use Jellyfin as my go to place for music but the navigation, although quite functional, lacks usability for avid music users.

I'm not sure if this is the place to have such discussion though so forgive me if this is out of topic.

PR #1492, opened 2021-09-06
feat(ui): new transitions and CSS fixes for library app bar

ferferga:
Adds an sliding effect while switching pages. Also, removed unused CSS, as Vuetify itself positions the element correctly and it previously had some visual quirks.

### DIFF
--- a/src/pages/library/_viewId.vue
+++ b/src/pages/library/_viewId.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <v-app-bar dense class="second-toolbar">
+    <v-app-bar dense flat>
       <span class="text-h6 hidden-sm-and-down">
         {{ collectionInfo.Name }}
       </span>
@@ -37,7 +37,7 @@
       <play-button :item="collectionInfo" shuffle />
       <play-button :item="collectionInfo" />
     </v-app-bar>
-    <v-container class="after-second-toolbar">
+    <v-container>
       <skeleton-item-grid v-if="loading" :view-type="viewType" />
       <item-grid :loading="loading" :items="items">
         <h1 v-if="!hasFilters && isDefaultView" class="text-h5">
--- a/src/pages/library/_viewId.vue
+++ b/src/pages/library/_viewId.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <v-app-bar fixed flat dense class="second-toolbar">
+    <v-app-bar dense class="second-toolbar">
       <span class="text-h6 hidden-sm-and-down">
         {{ collectionInfo.Name }}
       </span>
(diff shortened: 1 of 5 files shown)


dkanada:
I don't think this is a good idea unless we can slide in the opposite direction when navigating back in the browser history.

ferferga:
So going back goes right, correct? Entering content will always appear going from left to right though.

src/pages/library/_viewId.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <v-app-bar fixed flat dense class="second-toolbar">
```
MrTimscampi, line 3: This breaks the current way these toolbar work, as they no longer stay at the top when scrolling, which is imo a lot less user-friendly (You have to scroll back up, sometimes through hundreds of items, to change a filter or something...)

src/pages/library/_viewId.vue
```diff
@@ -339,27 +339,6 @@ export default Vue.extend({
 </script>
 
 <style lang="scss" scoped>
-@import '~vuetify/src/styles/styles.sass';
```
MrTimscampi, line 342: #1478 already cleans this up a lot by using the Vuetify values to position it properly.

MrTimscampi (commented):
If the intent is to ditch the toolbar staying up at the top of the library, then by all means go ahead, of course.

The page transitions also likely need a low motion toggle. See https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

ferferga:
@jellyfin-bot rebase

ferferga:
@ThibaultNocchi Using Portal Vue to append the content to main app's app bar fixes the issue

ferferga:
@jellyfin-bot rebase

ThibaultNocchi:
@jellyfin-bot rebase

src/layouts/default.vue
```diff
@@ -56,6 +56,7 @@
       class="pt-s pl-2 pr-2 app-bar-safe-zone"
       flat
       app
+      :extension-height="isSecondToolbarActive ? 48 : 0"
```
ThibaultNocchi, line 59: On the homepage the toolbar is way too big, I *think* this may be due to that?

ferferga:
This PR now requires #1593 merged first. The sizing issue is caused mostly by all the trickery needed for this to work with keep alive.

ferferga:
After discussing this a little in Matrx, I think it makes sense to not make the filtering elements sticky and force the user to scroll up to see them. Filtering elements while the view is scrolled down makes all the content jump out of place.

I don't recall seeing an UI design where the filtering cues are sticky (Amazon, Google, Spotify comes to my head).

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

ThibaultNocchi (approved):

ferferga:
Merging this, a subsequent PR for an "scroll to top button" will be made

Issue #1662

PR #1493, opened 2021-09-06
fix(home): cannot read properties of null (reading '_normalized')

ferferga:
Adds missing conditionals to the functional components used in the home swiper.


ferferga:
@dkanada For me it was just waiting for home to load. 

I vaguely remember if I checked which of my media caused it (clearly something that's missing a parent, which I don't know if it's possible, but who knows).

PR #1571, opened 2021-10-27
chore(deps): Update to Node 16

ferferga:
Node 16 is now LTS, so it's fully safe to upgrade to it.

Enforces it as the minimum version required for installing the frontend, alongside, npm 8.1.0, the npm version that comes pre-bundled with Node 16.13.0 (the first LTS version)

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     "vue-jest": "^3.0.7"
   },
   "engines": {
-    "node": ">=16.13.0",
+    "node": ">=16.13.0 <17.0.0",
     "npm": ">=8.1.0",
     "yarn": "Yarn is not supported. Please use NPM."
   },


package.json
```diff
@@ -142,8 +142,8 @@
     "vue-jest": "^3.0.7"
   },
   "engines": {
-    "node": ">=14.0.0 <15.0.0",
-    "npm": "<2.0.0",
+    "node": ">=16.13.0",
```
ThibaultNocchi, line 145: Maybe add a `<17` so that we stay on LTS?

ferferga: @ThibaultNocchi Good catch! 👏

ThibaultNocchi (approved):

Issue #1365, opened 2021-07-19
"Soft" back arrow can be clicked when hidden and also gets you out of JF Vue


ThibaultNocchi:
The back arrow on the top bar (the Vue one, not the browser one) is hidden on the homepage but can be clicked.

Also, if you click on it it's a simple backward navigation in the history, returning you to wherever you were, be it Vue or not.

MrTimscampi:
That would be because the `hide` directive added by @ferferga uses `el.style.opacity = '0'`, which doesn't disable any of the elements or their events.

PR #1572, opened 2021-10-28
fix(ui): "Soft" back arrow can be clicked when hidden

Th1nkK1D:
From the issue #1365 : The back arrow on the top bar is hidden on the homepage but can be clicked.

So I change the hide directive logic from using CSS opacity to the display property.


ferferga:
@jellyfin-bot rebase

ferferga (approved):
Thank you very much!

Sorry for the time this simple change took - I'm overwhelmed by uni this semester and it's almost impossible to dedicate time to Vue :(.

PR #1588, opened 2021-11-04
fix(mediastreamselector): fix type detection on empty lists

ThibaultNocchi:


ferferga:
What exactly does this fixes? Repro steps?

ferferga:
@jellyfin-bot rebase

ThibaultNocchi:
Best example is when you don't have any subtitle. The current way is to get the type from the first provided media stream. But when there's non, it crashes (silently in prod).

ferferga (approved):

PR #1590, opened 2021-11-05
feat(video): iPhone native fullscreen

ThibaultNocchi:
Signed-off-by: Thibault Nocchi <1619359+ThibaultNocchi@users.noreply.github.com>

### DIFF
--- a/src/components/Players/HlsPlayer.vue
+++ b/src/components/Players/HlsPlayer.vue
@@ -411,6 +411,13 @@ export default Vue.extend({
       } else if (document.pictureInPictureEnabled) {
         this.videoElement.requestPictureInPicture();
       }
+    },
+    toggleNativeFullscreen(): void {
+      // @ts-expect-error - `webkitEnterFullscreen` does not exist in relevant type
+      if (this.videoElement?.webkitEnterFullscreen) {
+        // @ts-expect-error - `webkitEnterFullscreen` does not exist in relevant type
+        this.videoElement.webkitEnterFullscreen();
+      }
     }
   }
 });
--- a/src/components/Players/PlayerManager.vue
+++ b/src/components/Players/PlayerManager.vue
@@ -598,11 +598,17 @@ export default Vue.extend({
         return;
       }
 
-      this.$fullscreen.toggle((this.$refs.playerContainer as Vue).$el, {
-        callback: (fullscreen: boolean) => {
-          this.fullScreenVideo = fullscreen;
-        }
-      });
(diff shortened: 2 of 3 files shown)


ferferga:
@jellyfin-bot rebase

ferferga:
What's the point of vue fullscreen if we need this kind of overrides anyway?

ThibaultNocchi:
vue-fullscreen sets a DOM element as the main element of the page and puts the browser in fullscreen. iPhone doesn't support that.

What I'm doing here is put the `<video>` element in fullscreen on iPhones instead. This allows the bottom bar on the screen to disappear

ferferga:
Now that you're mentioning it I remembered how vue fullscreen worked.

Do we know though why it does do that? Why vue fullscreen doesn't simply request the video element to be fullscreen without doing any div trickery? Quick guess is that it does that work to overcome old browser's limitations. However, do we know why it doesn't check if current browser doesn't support fullscreen to avoid doing the div manipulation?

My point still remains, these changes can be upstreamed anyway (just avoid the div stuff there for iOS and you're done). I don't want to have two paths for reaching the same goal - Either we implement all the fullscreen behaviour on our own, or we fully rely on an external library, not a mix because that's confusing and makes things difficult to debug later.

I'm not opposed to full library or full our own (I maybe prefer our own a little bit more because this doesn't mean the implementation is fully "agnostic", vue fullscreen for instance needs a switch to vue3-fullscreen, which is an additional step we need to remember), I'm just opposed to the mix of both.

ThibaultNocchi:
We don't want the video HTML element to go fullscreen, as we need to just be displayed in the background of our OSD. That's why we use vue-fullscreen: it's like pressing F11 in any desktop browser, but programmatically.

On iOS iPhone, 1. we can't call that 2. but on PWA it's already fullscreen. But, on the button-less iPhones, there's always the bar on the bottom to go to the homescreen.

next

The only solution is to make the video element go fullscreen, but at the price of the OSD. This is how it was done for a while in the iOS client (didn't check lately). So on iOS you get two OSD, the JF one to manage audio and such, and then the iOS one to have a true fullscreen.

So, at least that's something that shouldn't go to vue-fullscreen as it's not its use case. We can further discuss if this fullscreen for iPhone feature is needed, but at least it's not a feature for vue-fullscreen as they are two different use cases. :)

ferferga (approved):

ferferga:
@jellyfin-bot rebase

PR #1591, opened 2021-11-05
feat(playback): keep volume between playback sessions

### LINKED ISSUE #1545: Music playback volume resetting when starting a new playback instance
### Description of the bug

The volume you set in the player keeps resetting to 100 when starting a new playback anywhere.
This does not happen when you go to a different track in the current queue.

### Steps to reproduce

I am using a docker build of the latest master on my server but this also happens on the online demo.
Here's a small gif showing the issue

### Expected behavior

Preserve set volume

### Logs

_No response_

### Screenshots

_No response_

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.7.6

### Additional context

_No response_

ThibaultNocchi:
fix #1545

### DIFF
--- a/src/store/playbackManager.ts
+++ b/src/store/playbackManager.ts
@@ -319,7 +319,10 @@ export const mutations: MutationTree<PlaybackManagerState> = {
     state.status = PlaybackStatus.Paused;
   },
   STOP_PLAYBACK(state: PlaybackManagerState) {
+    const volume = state.currentVolume;
+
     Object.assign(state, defaultState());
+    state.currentVolume = volume;
   },
   RESET_LAST_ITEM_INDEX(state: PlaybackManagerState) {
     state.lastItemIndex = null;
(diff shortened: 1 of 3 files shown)


src/store/playbackManager.ts
```diff
@@ -319,7 +319,10 @@ export const mutations: MutationTree<PlaybackManagerState> = {
     state.status = PlaybackStatus.Paused;
   },
   STOP_PLAYBACK(state: PlaybackManagerState) {
+    const volume = state.currentVolume;
+
     Object.assign(state, defaultState());
+    state.currentVolume = volume;
```
ferferga, line 325: I'm sure you have tested this :P, but just in case:

Are you sure volume isn't a reference to the state object? So as it's assigned by Object.assign, ``volume`` becomes the default value instead of the old one (which is the one we want to persist). So maybe we need to use Object.assign as well for volume.

ThibaultNocchi: `volume` is assigned to the `currentVolume` value. It's only when passing around objects that you get a reference. And yes it was tested :p

ferferga (approved):

PR #1592, opened 2021-11-10
refactor: remove fullscreen dep to do it with native JS method

ThibaultNocchi:
I've tested it on FF and Chrome, can someone test on Safari Desktop and Android please?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
         "nuxt": "^2.15.8",
         "nuxt-vuex-localstorage": "^1.3.0",
         "qs": "^6.10.1",
+        "screenfull": "^6.0.0",
         "shaka-player": "^3.2.1",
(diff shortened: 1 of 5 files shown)


ferferga:
@ThibaultNocchi Safari fails with ``TypeError: document.documentElement.requestFullcreen is undefined``

- S21 Ultra with Chrome: Works fine
- S21 Ultra with Firefox: Works fine

All tested in BrowserStack. Wanted to test with Samsung Internet Explorer (so we have some info of a common but not mainstream) browser but Browser'sStack local connect didn't work properly with it.

ferferga:
As a reminder, we decided to stick with screenfull, which handles all of this for us and it's simple enough without all the extra boilerplate that an additional Vue component adds to this context.

ferferga (approved):

PR #1593, opened 2021-11-11
feat: remove keep-alive

ferferga:
Keep alive introduces lots of complexities to the state management of the app and it consumes a lot of memory. It's true that removing it introduces some loading time, but what we need is proper item caching instead and we're still not production ready, so this is a minor annoyance.

It never worked properly when switching pages when a promise is running (can be seen easily by going to a library page from the home page while it's loading).

### DIFF
--- a/src/pages/fullscreen/playback/index.vue
+++ b/src/pages/fullscreen/playback/index.vue
@@ -95,15 +95,13 @@ export default Vue.extend({
       this.setBackdrop({ hash: this.backdropHash });
     });
   },
-  activated() {
+  mounted() {
     this.setAppBarOpacity({ opaqueAppBar: false });
     this.setBackdropOpacity({ newOpacity: 0.5 });
     this.setMinimized({ minimized: false });
-  },
-  mounted() {
     this.swiper = (this.$refs.playbackSwiper as Vue).$swiper as Swiper;
   },
-  deactivated() {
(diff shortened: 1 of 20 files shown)


src/pages/fullscreen/playback/index.vue
```diff
@@ -95,15 +95,13 @@ export default Vue.extend({
...
     });
   },
-  activated() {
+  mounted() {
     this.setAppBarOpacity({ opaqueAppBar: false });
     this.setBackdropOpacity({ newOpacity: 0.5 });
     this.setMinimized({ minimized: false });
-  },
-  mounted() {
     this.swiper = (this.$refs.playbackSwiper as Vue).$swiper as Swiper;
```
ThibaultNocchi, line 102: I'd put the old mounted command before the old activated method: in the lifecycle mounted comes before activated

ThibaultNocchi (commented):
Trying it around, we effectively lose some feature. Not caching, but stuff such as not recreating library lists or the home header. For instance, every time you go back to the home menu, you get the welcome message. So idk if deleting the keep-alive right now is good?

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

ThibaultNocchi (approved):

PR #1594, opened 2021-11-11
chore: SPA first and other improvements

ferferga:
This use cases for this client doesn't really warrant to be SSR first and foremost. This PR makes SPA our default choice (leaving the option of SSR for people who want it) and further improvements:

- Updated README with more clear and concise instructions and styling aligned to the one used in other Jellyfin projects
- Added emojis to CI workflows
- Added missing expose label to the dockerfile that holds the SPA version of our client.


ferferga:
@ThibaultNocchi Thoughts on removing ``dev`` commands altogether and having start instead?

ThibaultNocchi:
There's a difference between these commands right?

ferferga:
@ThibaultNocchi So it works like current Web, where you simply run ``npm start`` instead of ``npm run dev:static`` like we do know.

So we only would have a ``npm start`` (for spa development) and ``npm run start:ssr`` (for SSR development).

ThibaultNocchi:
Oh you were meaning just changing the npm alias to start? Cause nuxt has a start command and it just starts a web server with the dist content :p I got no problem with having `npm start` running `nuxt dev` :)

package.json
```diff
@@ -13,22 +13,19 @@
...
     "lint": "npm run lint:js && npm run lint:style",
     "lint:js": "eslint --ext .ts,.js,.json,.vue .",
     "lint:style": "stylelint **/*.{vue,css} --ignore-path .gitignore",
     "prepare": "husky install",
-    "prod": "npm run build src && npm run start src",
-    "prod:static": "npm run build:static src && npm run start:static src",
-    "start": "NUXT_SSR=1 nuxt start src",
-    "start:static": "nuxt start src",
+    "prod": "npm run build src && nuxt start src",
+    "prod:ssr": "npm run build:ssr src && NUXT_SSR=1 nuxt start src",
```
ThibaultNocchi, line 28:
```suggestion
    "prod": "npm run build && nuxt start src",
    "prod:ssr": "npm run build:ssr && NUXT_SSR=1 nuxt start src",
```

I think it should be `npm run build` instead of `npm run build src` right?

README.md
```diff
@@ -36,72 +36,87 @@
...
 
-## Build Process
+   ```bash
+   npm start
+   ```
 
-```bash
-# install dependencies
-$ npm install
+   The client will be served at http://127.0.0.1:3000 by default.
```
ThibaultNocchi, line 68:
```suggestion
   The client will be available at http://127.0.0.1:3000 by default.
```

I'd use this instead of served, cause really it's served on 0.0.0.0 ^^

README.md
```diff
@@ -36,72 +36,87 @@
...
 
-# build for production in static mode
-# you will need a web server to host the client
-$ npm run build:static
+### Running a production build
 
-# build for production in static mode and run server
-$ npm run prod:static
+Instead of a development version, you can run a server with a production-ready build of the client
+directly with Nuxt, so you can verify in advance how the client will work in your production environment:
```
ThibaultNocchi, line 86: I'd add something to say it's not a server suitable for real production :)

package.json
ferferga: No, because source code is under the "src" directory

ThibaultNocchi: But it's already in the built in npm command :)

README.md
```diff
@@ -83,12 +83,15 @@ This is an experimental web client for Jellyfin based on Vue.js. We welcome all
 ### Running a production build
 
 Instead of a development version, you can run a server with a production-ready build of the client
-directly with Nuxt, so you can verify in advance how the client will work in your production environment:
+directly with Nuxt, so you can verify in advance how the client will work in your production environment with a reverse proxy:
```
ThibaultNocchi, line 86: Not even a RP, you can just use a classic HTTP server or what else. I'd just add a different sentence like "Please note this isn't suitable for a real production hosting." or the same sentence they put in the Nuxt doc if you wish :)

ThibaultNocchi (approved):

ferferga:
LGTM now as well!

PR #1611, opened 2021-12-04
Generalize HomeHeader and rework the loading process

ferferga:
This is a first pass at generalizing the HomeHeader component so it's more reusable across multiple places, as you can see in this video (this index page was not committed, is just for demo purposes):

https://user-images.githubusercontent.com/10274099/144713716-710e6a76-3bfb-490b-8b2d-692485506b1c.mp4

Styling in the video is nonexistent, but this should allow for more complex designs like [this](https://github.com/jellyfin/jellyfin-vue/discussions/498) and [this](https://github.com/jellyfin/jellyfin-vue/discussions/478) in the future.

**Why I call it first pass?**: Because the ideal architecture in my opinion would be to have a "GenericCarousel" component that have all the slides inside a slot, so we can have subcomponents like "ImagesCarousel", "VideoCarousel" etc all sharing similar stylings and so on. 

* HomeHeader as is was renamed to "ItemsCarousel"
* "ItemsCarousel" takes the items as a prop, instead of doing the fetch itself
* Items for the index page are fetched in the index's page context and using *asyncData* (so no popping at all, thus, **the welcome message is no longer needed**)
*The fetch and asyncData hook should be merged in [Nuxt 3](https://v3.nuxtjs.org/docs/usage/data-fetching/), so we have all the full context in the hook but also the possibility of block navigation until all the data is fetched using Vue 3's suspense component)*

* Generalized the ItemTitles components


ThibaultNocchi:
@jellyfin-bot rebase

src/components/Layout/Carousel/Item/CarouselItemLogo.vue
```diff
@@ -0,0 +1,82 @@
...
+              { Id: this.item.SeriesId },
+              'Series'
+            );
+          }
+
+          break;
+      }
+    } else {
+      this.link = this.getItemDetailsLink(this.item);
+    }
```
ThibaultNocchi, line 77: The last else should be the default case of the if(this.parent) :)

ferferga: Good catch!

However, there's another problem with your logic: what happens if an album doesn't have a music artist, for instance?

So I added an if to the end to check if the string is still empty to populate it properly :)

src/components/Layout/Carousel/Item/CarouselItemLogo.vue
```diff
@@ -0,0 +1,88 @@
...
+          }
+
+          break;
+      }
+
+      // Instead of using 'default', we need this additional extra check
+      // in case an Album doesn't have artists, for example.
+      if (this.link === '') {
+        this.link = this.getItemDetailsLink(this.item);
+      }
```
ThibaultNocchi, line 80: I'd merge those switch and if into a single if
```suggestion
          if (this.item.Type==='MusicAlbum' && this.item.AlbumArtists?.length) {
            this.link = this.getItemDetailsLink(
              this.item.AlbumArtists[0],
              'MusicArtist'
            );
          }
          else if (this.item.Type==='Episode' && this.item.SeriesId) {
            this.link = this.getItemDetailsLink(
              { Id: this.item.SeriesId },
              'Series'
            );
          }

      // Instead of using 'default', we need this additional extra check
      // in case an Album doesn't have artists, for example.
      else{
        this.link = this.getItemDetailsLink(this.item);
      }
```

src/components/Layout/Carousel/Item/CarouselItemTitle.vue
```diff
@@ -0,0 +1,90 @@
...
+          }
+
+          break;
+      }
+
+      // Instead of using 'default', we need this additional extra check
+      // in case an Album doesn't have artists, for example.
+      if (this.link === '') {
+        this.link = this.getItemDetailsLink(this.item);
+      }
```
ThibaultNocchi, line 84: Same here

ThibaultNocchi (commented):
The title is appearing with the logo and titles alone seem to be reduced in height :)

ferferga:
This PR is ready to be merged (and I want it merged first) but it requires #1593 so the activated/deactivated hooks are not needed. Otherwise, we need an additional variable in the ``ItemsCarousel`` component that keeps track of the current page, as when not using keep alive all of this is event-bound.

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

ThibaultNocchi:
@jellyfin-bot rebase

src/components/Layout/Carousel/Carousel.vue
```diff
@@ -0,0 +1,128 @@
...
+    <carousel-progress-bar
+      v-if="progressBar && topProgressBar && slides > 0"
+      :pages="slides"
+      :current-index="currentIndex"
+      :duration="slideDuration"
+      :paused="isPaused"
+      class="px-2 px-sm-4 progress-bar"
+      hoverable
+      @on-animation-end="onAnimationEnd"
+      @on-progress-clicked="onProgressClicked"
```
ThibaultNocchi, line 12: For now it's not a bother, but I'm sure we'll be able to factorize the top and bottom progress bars

src/components/Layout/Carousel/Item/ItemsCarouselTitle.vue
```diff
@@ -0,0 +1,125 @@
...
+  data() {
+    return {
+      itemLink: '',
+      titleString: '',
+      logoLink: '',
+      subtitle: '',
+      logo: {} as ImageUrlInfo
+    };
+  },
+  watch: {
```
ThibaultNocchi, line 61: I guess this watcher could be done as a computed, wdyt?

ThibaultNocchi: Yeah I see what you're saying. It's all good to me then :)

ThibaultNocchi (approved):

ThibaultNocchi:
@jellyfin-bot rebase

PR #1627, opened 2021-12-08
Feat/commit link

ThibaultNocchi:
This adds the possibility to add the `NUXT_ENV_COMMIT` var at build or dev time to specify a commit SHA. If it exists this will add a link to this commit on GH at the bottom of the nav bar.

print_screen_2021-12-09-10-34-26

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
         run: npm ci --no-audit
 
       - name: Build 🛠️
-        run: 'npm run build:ssr'
+        run: 'NUXT_ENV_COMMIT=$GITHUB_SHA npm run build:ssr'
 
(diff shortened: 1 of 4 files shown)


ferferga (commented):
Dont you need to also define a key for this variable in nuxt.config?

Also, please post your screenshot in the original post, so later on if we need to bisect we know exactly what this PR does and how does it look like :)

ThibaultNocchi:
Nope, per [this doc](https://nuxtjs.org/docs/configuration-glossary/configuration-env/#automatic-injection-of-environment-variables) env vars beginning with `NUXT_ENV_` are injected into the `process.env` var. I also tested it locally so that should be good.

ferferga (approved):

PR #1632, opened 2021-12-13
Deprecate SSR

ferferga:
PR for implementing #1620


ferferga:
All the files under the ``plugins`` directory were renamed without changes (but the ``components.ts`` one) so don't be afraid of the huge diff! :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install dependencies 📦
+        run: npm ci --no-audit
+
+      - name: Build client 🛠️
+        run: npm run build
+
+      - name: Configure QEMU ⚙️
+        uses: docker/setup-qemu-action@v1.2.0
+
+      - name: Configure Docker Buildx ⚙️
```
dkanada, line 34: Buildx or BuildX? Honestly not sure.

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+            ghcr.io/jellyfin/jellyfin-vue:latest
+
+      - name: Build and push release candidate images 🛠️⬆️
+        if: github.event.action == 'prereleased'
+        uses: docker/build-push-action@v2.6.1
+        with:
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm,linux/arm/v6
+          file: '.github/workflows/Dockerfile.nginx'
```
dkanada, line 73: Since we only have one Dockerfile now maybe we can drop the suffix at the end?

README.md
```diff
@@ -96,25 +96,6 @@ npm run prod
...
-
-This client fully supports Server-Side Rendering. All the commands listed above have their SSR siblings:
-
-```bash
-npm run build:ssr
-npm run start:ssr
-npm run prod:ssr
-```
-
-#### Running standalone
```
dkanada, line 109: Is this provided somewhere else now?

src/plugins/nuxt/playbackProfiles/helpers/webmVideoFormats.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @param {Context} context - Nuxt context
+ * @param {HTMLVideoElement} videoTestElement - A HTML video element for testing codecs
+ * @returns {string[]} An array of supported codecs
+ */
+export function getSupportedWebMVideoCodecs(
+  context: Context,
+  videoTestElement: HTMLVideoElement
+): string[] {
```
dkanada, line 12: Not really a fan of moving all the plugins into subdirectories.

README.md
ferferga: Standalone has nothing to do with CSR: https://nuxtjs.org/docs/configuration-glossary/configuration-build/#standalone

For serving the client, there's ``npm run prod``

src/plugins/nuxt/playbackProfiles/helpers/webmVideoFormats.ts
ferferga: Why?

.github/workflows/release.yml
ThibaultNocchi: Buildx it seems https://docs.docker.com/buildx/working-with-buildx/

.github/workflows/release.yml
ferferga: @dkanada Renamed to ``Dockerfile.ci``, so it's clear the distinction between the "normal" Dockerfile (where the client is also built in the client context) and this one, where the client is first built in CI and, after that, copied to a Docker image.

dkanada: I wonder why those need to be different in the first place but this is a good change so it sounds fine to me.

README.md
dkanada: But is that included in the README after this new pull request or is it left for the users to figure out now?

src/plugins/nuxt/playbackProfiles/helpers/webmVideoFormats.ts
dkanada: After comparing it with master I can see that this is a net benefit - the nuxt folder just feels like a catch-all even though there's not really a better way to categorize them. It also looks like `directives` wasn't removed from its old location on your branch by the way.

README.md
ferferga: Just above this section, called "Running a  production build".

src/plugins/nuxt/playbackProfiles/helpers/webmVideoFormats.ts
ferferga: Good catch! Will fix it!

ThibaultNocchi (commented):
1. I'd just rename Dockerfile.ci in the workflows to Dockerfile, but that's a nitpick (renaming it should make us able to remove the line in the workflows pointing to the file cause it should default to `Dockerfile` I guess) 
2. I think you duplicated the whole `/src/plugins/nuxt/playbackProfiles` to `/src/plugins/playbackProfiles` to `/src/plugins/nuxt/playbackProfiles` (or vice versa)

.github/workflows/release.yml
ferferga: @dkanada The main one builds the client entirely in the Docker context while the CI one takes the building output from the previous step (which performs the build in the host of the CI environment), saving resources and time.

dkanada: The time savings might be useful, but I honestly couldn't say without specific values for comparison. Trying to improve runtime by sidestepping the official image sounds entirely counterproductive though - the whole point of CI is to test our build, but in this case we're only testing the native build, not the Docker image that gets shipped to users.

ThibaultNocchi (approved):

PR #1636, opened 2021-12-14
feat(menu): show context menu when clicking on parent element

ferferga:
Right clicking components that have the ``item-menu`` component as a child will make the menu appear where the pointer is located

### DIFF
--- a/src/components/Item/ItemMenu.vue
+++ b/src/components/Item/ItemMenu.vue
@@ -12,6 +12,7 @@
         top
       >
         <template #activator="{ on, attrs }">
+          <!-- See comments for this in the onRightClick method -->
           <v-btn
             icon
             :outlined="outlined"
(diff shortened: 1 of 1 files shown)


ferferga:
@jellyfin-bot rebase

ThibaultNocchi:
@jellyfin-bot rebase

src/components/Item/ItemMenu.vue
```diff
@@ -1,39 +1,45 @@
...
+      >
+        <template #activator="{ on, attrs }">
+          <v-btn
+            icon
+            :outlined="outlined"
+            :dark="dark"
+            v-bind="attrs"
+            v-on="on"
+            @click.stop.prevent="onActivatorClick"
+            @contextmenu="onRightClick"
```
ThibaultNocchi, line 22: Couldn't we add the `.stop.prevent` here instead of the method?

ferferga: Originally, I planned this to work as follow:

- We get the ref of the parent element
- We bind the element using ``$on`` [docs](https://vuejs.org/v2/api/#vm-on)

That way we could've used that syntax. However, that syntax only works when the event is a custom one (emitted using ``$emit()``) and we're using an standard JS one. This means we need to bind them the JS way, no the Vue way.

That's why the ``e.preventDefault()`` and ``e.stopPropagation()`` are in the event handler.

However, here we're binding the same callback to a ``contextmenu`` event in a child component of the ``ItemMenu`` one (so we can use the SFC syntax, instead of the ref one) in order to reuse the code

We are forced to use the stopPropagation and preventDefault in the function so the paren'ts right click are handled correctly. The decision here relies on if we make another function that gets called for the parent right click or reuse the same code. Something like this:

````
onParentRightClick(e) {
   e.preventDefaukt();
   e.stopPropagation();
   this.onRightClick();
}
onRightClick() {
  // The rest of the current logic. This would be the one that it's referenced in the SFC.
}
````

ferferga: Imo the current approach in this PR is better, the problem is this quirk in Vue's API (which hopefully gets fixed in Vue 3, I haven't checked honestly).

ferferga: Maybe I should add a comment for this?

ThibaultNocchi: Yes please do :)

ThibaultNocchi (approved):

Issue #1610, opened 2021-12-04
Building error on arm64 machine


martadinata666:
### Description of the bug

Error installing node deps `opting-bin`
```
dedyms@rpi4 jellyfin-vue % docker build -t jellyfin-vue:local -f Dockerfile.ssr .
Sending build context to Docker daemon  3.913MB
Step 1/15 : FROM node:16-alpine AS build
 ---> 8074a83d0ba9
Step 2/15 : WORKDIR /app
 ---> Using cache
 ---> 633ca21c678c
Step 3/15 : RUN apk add --no-cache --virtual .build-deps git python3 make automake autoconf g++ libpng-dev libtool nasm file
 ---> Running in 19dd676b7a96
fetch https://dl-cdn.alpinelinux.org/alpine/v3.14/main/aarch64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.14/community/aarch64/APKINDEX.tar.gz
(1/48) Installing ca-certificates (20191127-r5)
(2/48) Installing brotli-libs (1.0.9-r5)
(3/48) Installing nghttp2-libs (1.43.0-r0)
(4/48) Installing libcurl (7.79.1-r0)
(5/48) Installing expat (2.4.1-r0)
(6/48) Installing pcre2 (10.36-r0)
(7/48) Installing git (2.32.0-r0)
(8/48) Installing libbz2 (1.0.8-r1)
(9/48) Installing libffi (3.3-r2)
(10/48) Installing gdbm (1.19-r0)
(11/48) Installing xz-libs (5.2.5-r0)
(12/48) Installing mpdecimal (2.5.1-r1)
(13/48) Installing ncurses-terminfo-base (6.2_p20210612-r0)
(14/48) Installing ncurses-libs (6.2_p20210612-r0)
(15/48) Installing readline (8.1.0-r0)
(16/48) Installing sqlite-libs (3.35.5-r0)
(17/48) Installing python3 (3.9.5-r2)
(18/48) Installing make (4.3-r0)
(19/48) Installing perl (5.32.1-r0)
(20/48) Installing perl-error (0.17029-r1)
(21/48) Installing perl-git (2.32.0-r0)
(22/48) Installing git-perl (2.32.0-r0)
(23/48) Installing automake (1.16.3-r0)
(24/48) Installing m4 (1.4.18-r2)
(25/48) Installing autoconf (2.71-r0)
[...]

PR #1645, opened 2021-12-23
ci(docker): Build Docker in CI

ferferga:
Performs all the builds of the client using the main Dockerfile we have in the root of this project.

This will allow us to check for both client builds and Docker builds (as, if client fails, Docker will fail too), instead of using a separate file for CI only, which makes stuff unreproducible and might re-introduce issues like #1610

Removed also the files under the .docker folder, as the Docker image works perfectly without them and it means less maintenance burden. Requesting review to @Shadowghost to verify this is the right move and they're not needed anymore.

Only downside to this PR is that it adds some time to build for all the architectures, but I think it's worth it and makes our pipeline way more robust (specially for #1627 changes).


Shadowghost (approved):
Changes look fine to me.

The modified nginx config was used because we wanted gzip compression where applicable.
If that's not a priority the removal is fine.

.github/workflows/pull_request.yml
```diff
@@ -57,20 +57,31 @@ jobs:
...
-        run: NUXT_ENV_COMMIT=$GITHUB_SHA npm run build
+      - name: Build Docker images 🛠️
+        uses: docker/build-push-action@v2.6.1
+        with:
+          context: .
+          platforms: ${{ matrix.platform }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          tags: |
+            jellyfin/jellyfin-vue:pr-build
```
ThibaultNocchi, line 87: What does that do if nothing is pushed? Is it for the cache or just because the action requires something? If the second answer, maybe add a comment to say so :)

Dockerfile
```diff
@@ -1,9 +1,15 @@
 ## This dockerfile builds the client entirely in a Docker context
 
 FROM node:16-alpine AS build
+# Set labels
+LABEL maintainer="Jellyfin Packaging Team - packaging@jellyfin.org"
+LABEL org.opencontainers.image.source="https://github.com/jellyfin/jellyfin-vue"
```
ThibaultNocchi, line 6: Idk much the utility of those, but shouldn't they go the last stage?

ferferga: I've been always seeing those at the top of the Dockerfile, so I assume it's common practice to do so. Do you have any source with common practices for this? [Nothing is specified in the best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#label)

.github/workflows/pull_request.yml
ferferga: Just builds the image to see if build succeeds. I removed the client build step altogether because we can assume that, if Docker images build, client will also be built :).

ferferga:
@jellyfin-bot rebase

.github/workflows/pull_request.yml
ThibaultNocchi: Np about that, I was just wondering what that line did in that case :)

Dockerfile
ThibaultNocchi: I got no problem with, I'm just wondering if it should be present in the last stage as that's the one pushed

ferferga: That doesn't matter, Docker stacks the layers one on top of another, so the last one contains all that are above it.

ThibaultNocchi (approved):

PR #1646, opened 2021-12-23
fix(arm): revert #1136 to fix arm builds

ferferga:
NodeJS GC fails on ARM architectures when running ``fork-ts-checker``. Reverting the change until a better solution is found.

Also fixed an small typo in readme.


ThibaultNocchi (approved):

ferferga:
@jellyfin-bot rebase

PR #1668, opened 2022-01-03
test with only git as dependency

ferferga:


ferferga:
Oops testing some stuff in my fork and opened in the wrong repo!

PR #1672, opened 2022-01-06
Add Taui target

crobibero:
Getting started: https://tauri.studio/en/docs/getting-started/intro


ferferga (approved):
The Windows checks are irrelevant to Tauri but the frontend itself, so it's irrelevant to fix it here and make this PR bigger.

Now fingers crossed so Weblate doesn't decide to commit in the meantime :)

ThibaultNocchi:
Can @ferferga or you update the README for the new path?

ferferga:
@ThibaultNocchi What exactly needs fixing in the readme? The default workspace it's still the frontend and nothing changes in our workflow, besides having the ability of bundling the app if we want to.

I want to create a separate BUILDING.md file with all the instructions (like the new env vars, etc), but I'd rather do that in another PR as it's not Tauri exclusive.

DesertCookie:
> Getting started: https://tauri.studio/en/docs/getting-started/intro

That link is, sadly, dead.

PR #1677, opened 2022-01-08
[Media type redesign]: Fully-featured app bar and cosmetic changes

ferferga:
This PR is the first one to tackle the media type redesign mentioned in our 2022 roadmap, **focusing in the app bar component of the navigation experience**. This is a generic overview of how the home screen appears now:
Sin título

# App bar
All the icons that represent the state of the app are here now:

#### Only shown in specific situations
* **No connection** : Indicates that the user has lost the internet connection *in general (does not take into account yet disconnections from the server, as the component was refactored as is from the sidebar)*
* **Task Manager**: Shows tasks that are running in the client
* **Cast**: Appears when there are casting targets available. *(Not implemented yet, as in master)*

#### Always shown
* **Magnifying glass**: Search your media *(Not implemented yet, see "Further details")*
* **User**: Access settings, metadata editor and logout.

## Other details
The locale and dark mode switch buttons are still in this branch (leaving a cluttered view). However, they will be moved to settings once another pass focuses on getting the settings page done.

# Task manager
This is one of the big improvements of this branch. Task manager shows the user the long running tasks that are happening. As of right now those are:

* Library refreshes
* Display preferences syncing to server

Task logs will stay for 5 seconds for the user to check until the button and the tasks disappears
The new task manager store allows to hook any task that might rn in the future (offline media syncing, server's scheduled tasks, Live TV recordings...).

See it in action:

https://user-images.githubusercontent.com/10274099/148623730-e878d908-eae8-4984-aae2-0c725264d2be.mp4

# Navigation drawer
On pages where the app bar is not opaque, the navigation drawer is transparent. It's also denser, with rounded corners in selected items, so it takes less space, but still gives a sensation of amplitude when it's transparent.

# Further details

* Created task manager store to handle running tasks
* Refactored the main components into subcomponents to clean up the main layout view
* Created an app-bar-button-layout component that allows easy creation of more buttons for the app bar. All the new buttons use it.
* This design will be unusable in mobile until the bottom navigation bar is done in another pass of this redesign process
* Search isn't available until the bottom navigation bar is done in another pass of this redesign process

# General walkthrough video

https://user-images.githubusercontent.com/10274099/148624152-4d307a40-a9c0-41c5-8bdf-78218bc22969.mp4


ThibaultNocchi:
So I like that, seems nice. I've got a few points:

* I think it feels kinda cluttered with everything on the right and nothing on the left. I'd let the search field as it on the left just to see. Going on this idea, maybe they could be collapsed when there is > X buttons for each display size. You could pin search and cast and make X other collapsable. Like small screens you'd see search, cast and three dots, on medium search, cast, white/dark mode and three dots, and so on. 
* Why is search not working if we keep the current format? Can't the logic be kept? Is it because you intend to do something with the bottom bar?
* Do you have any idea how you will plan to put them in the mobile design?

Thanks :)

ThibaultNocchi:
Ok those are interesting explanations.

So I kinda agree with the button pills and headers, just gotta see how it'll end up in action. But on the paper that sounds good. And yeah, if the left part of the bar feels "filled", then I don't mind the search button on the right. And yeah expanding it along the bar is a "must have" I'd say, more so on mobile.

About the bottom menu navigation on phone, I'm not convinced. It's twice the "design logic" for something that's not so outdated on mobile. I think that if we happen to do this, we could first do the rest of the redesign to master and then talk about navigation.

Wdyt?

Issue #1671, opened 2022-01-06
The collections do not contain the items


doc-code-hub:
### Description of the bug

![image](https://user-images.githubusercontent.com/61033694/148436260-af5f1bf9-a662-4283-b432-1036a0d33aac.png)

Where are the titles in the collections?

From the original website, I set up the films page to group the films in collections.
This means that in my movie list, all of those in a collection are not visible (outside of the collection itself).

But when I open a collection, I don't see any of the items that belong to it. Instead I only see suggestions of similar collections ...

The only way I have to see those contents is to search for them from the search bar or reproduce (at random) the collection, and view the single elements from the queue, during playback.

I believe that as in the original site, in the page of a collection you should be able to view the elements that belong to it (among the various types: series, films, etc).

### Steps to reproduce

- Create a collection
- Put some films in that collection
- Set this option (In Admin Dashboard > Library > Display tab, from the original site)
![image](https://user-images.githubusercontent.com/61033694/148435084-e4cb671e-daed-421b-9503-80d54008e4ae.png)
- From the Vue App, go to that collection

You shouldn't see any of the titles you added on that page

### Expected behavior

The collections page should contain the elements that belong to it, divided by type.

For example as in the photo taken from the original site
![image](https://user-images.githubusercontent.com/61033694/148436095-f249fa41-6830-4a95-b204-1ecab113842d.png)

### Platform

Windows

### Browser

Chrome

### Jellyfin server version

10.7.7

### Additional context

```text
I deployed with the provided Dockerfile on the same server running Jellyfin 10.7.7
```

ferferga:
Hello! This is a duplicate of #760
This is not implemented yet as this is still an experimental client and its a continuous work in progress.

Closing this in favor of the aforementioned PR

Issue #760, opened 2021-02-16
[Feature] Make a collections page


krutburken:
**Describe The Bug**

Collections are displayed only as one item, and I'm unable to choose which item in the collection to play (the only options available are play and shuffle). The indicator displays the correct number of items in the collection though.

Perseusdehond:
As of right now, there hasn't been anyone who made a proper item page for this. That is why they don't show up for now. My guess is that this will be implemented at some point (soon?). 
Alternatively instead of only showing just the movies, I have also made a design myself https://github.com/jellyfin/jellyfin-vue/discussions/478 for the collections item page for if anyone wants to spend the time to add something like that :)

krutburken:
Ah I see, I just looked through the issues and saw that no-one had reported on it. Should maybe be changed to a feature request then?

For me it would be fine if it just replicated the current "folder" behavior as seen in jellyfin-web, but I guess it depends on the use case.

Perseusdehond:
> Ah I see, I just looked through the issues and saw that no-one had reported on it. Should maybe be changed to a feature request then?
> 
Yeah, I think that is better.
> For me it would be fine if it just replicated the current "folder" behavior as seen in jellyfin-web, but I guess it depends on the use case. Jellyfin-web in general looks pretty outdated and bare-bones though so I'm really looking forward to this project reaching feature parity
>
I agree with that. I make those design ideas as an inspiration of what Vue could look like because I really like this project. 
I would be fine with just showing the movies at first so the collections would work at least. But my design is more of a not-super-bland-way to show these and imo a cool way to evolve the client

ferferga:
Implemented in #1680

Some design concepts are still present in discussions that are also relevant and can inspire for further work in this area

PR #1680, opened 2022-01-09
feat(collections page): collection detail with tabs divided by conten…

doc-code-hub:
As discussed in some issues (#1671   , #760) an exhaustive page is required for the collections.

Given my inability to view some of my content (as I described here #1671 ) I decided to propose a solution myself inspired by the series page. 

I believe that the subdivision into tabs (as for the seasons of the series) is a good solution to organize the contents present within a collection, with a subdivision of the contents by type.

screencapture-172-28-135-247-3000-collections-76e710870f48691ebda1ba44efd4b6b8-2022-01-09-02_22_29

## A few words about what I did

I created a new page called "collections/_itemId" starting with the "series/_itemId" page (and the SeasonTabs component) as a starting point.

I created a store called collections where I provided an action (getCollectionChildren) to call the "items.getItems" API by setting {userId, parentId} as filters to get all the items belonging to the collection (and added the related functions of Success and Fault, as well as the delete function for the reset).

In the store I set the "children" property which consists of a list of {type, items} objects to facilitate the use of the tabs on the page.

For better UX I have also reordered this list (divided by item type) by number of items.

------------------------------------------------------------------------------------------------------------

I think this is a **good starting point** to make the **collections available**, and soon expandable eventually as suggested in #760 with ease.

**The page works very well** in both **Desktop** and **Mobile**, however I have never used Vue.js before, I usually work with Angular or React.

I do not therefore exclude that it may be necessary to refactor for improvements or for something unnecessary coming from the series page that can be removed, and I am therefore open to any suggestion for some improvement before performing the merge.


doc-code-hub:
I guess this sonar failure is due to the **duplication** of lines in the new index.vue compared to that of the series.

How to optimize these portions of code shared by the two pages?

src/store/collections.ts
```diff
@@ -0,0 +1,116 @@
+import Vue from 'vue';
```
ferferga, line 1: Why do you think we need an specific store for this and not the items store? 

The item state handling still needs a huge refactor, our store design needs a refactor, which is still on hold: https://github.com/jellyfin/jellyfin-vue/issues/801

src/pages/collections/_itemId/index.vue
```diff
@@ -0,0 +1,298 @@
...
+    return {
+      itemId: '' as string,
+      parentItem: {} as BaseItemDto,
+      currentTab: 0,
+      loading: true,
+      viewType: '',
+      backdropImageSource: '',
+      currentVideoTrack: undefined as number | undefined,
+      currentAudioTrack: undefined as number | undefined,
+      currentSubtitleTrack: undefined as number | undefined
```
ferferga, line 209: Unneeded as well

src/pages/collections/_itemId/index.vue
```diff
@@ -0,0 +1,298 @@
...
+  },
+  destroyed() {
+    this.setAppBarOpacity({ opaqueAppBar: true });
+    this.clearBackdrop();
+  },
+  methods: {
+    ...mapActions('page', ['setPageTitle', 'setAppBarOpacity']),
+    ...mapActions('backdrop', ['setBackdrop', 'clearBackdrop']),
+    ...mapActions('collections', {
+      getCollectionChildren: 'getCollectionChildren'
```
ferferga, line 294: This is only needed if we want to change the way that function is invoked inside this vue instance. As the name you used is the same, you can simply map it as the other stores.

ferferga (changes requested):
The item details page in general require a huge refactor, moreover with the plans stated in [our roadmap](https://github.com/jellyfin/jellyfin-vue/discussions/1679)

I'm happy to merge this following "the old schema", however there's still a lot of useless stuff in this page that doesn't belog to collections at all.

It's also simpler imo, if you just add the collection's item to the current item page, instead of creating a whole new page. As I said we're going to refactor that at some point, but in the meantime we have less duplication.

doc-code-hub:
Thanks @ferferga , you are a great help!

I just approached the project and tried to get as much as possible by reading the code and discussions.

I imagined that there might still be code that is not useful and the fact that you show me what can be improved helps me to get more and more in the right perspective with vue and with the project.

I take everything you wrote, and I get to work reorganizing my new code to work on the item page without a dedicated store.

Thanks for your time'.

I'll be back with a new commit

doc-code-hub:
Following the above, I have:

- Moved the strictly necessary logic from the collection store to the items store
- Created the CollectionTabs component that retrieves the children of the collection from the "items" store, and displays them in separate tabs by type of content
- Removed the new "collections / _itemId" page in favor of the implementation of CollectionTabs in the Items page
- Removed the collection store

This is the new result:
screencapture-localhost-3000-item-76e710870f48691ebda1ba44efd4b6b8-2022-01-09-19_02_22

src/components/Item/CollectionTabs.vue
```diff
@@ -0,0 +1,69 @@
...
+        },
+        1904: {
+          visibleSlides: 5
+        }
+      }
+    };
+  },
+  async fetch() {
+    this.loading = true;
+    await this.fetchBoxsetChildren({ itemId: this.item.Id }).finally(
```
ferferga, line 53: Just FYI: Leaving this here means that, if the parent item changes (the collection item), nothing will change for this instance, as this logic is only called when the Vue instance for this component is being created (i.e when the item details page is loaded)

I don't know if the parent collection component can change at some point (by removing an item from a collection, but iirc that's not implemented yet), but I guess it's better to do things correctly from the start

You have two options:

* Move this logic to a watcher: [Vue watchers allow you to call a function when some variable changes](https://vuejs.org/v2/guide/computed.html). By default, they're executed once [the lifecycle is in mounted stage](https://es.vuejs.org/v2/guide/instance.html#Diagrama-del-Ciclo-de-vida), but you can override it and run the handler as well as soon as the component instance is created. See here: https://github.com/jellyfin/jellyfin-vue/blob/4f18f67fa9bbab78441d84fb74200207fa5db6f9/src/components/System/Snackbar.vue#L20-L28

* Keep this in the fetch hook but create a watcher for ``item`` that calls ``this.$fetch``. You can even remove ``loading`` from ``data`` and use fetchState and forget about handling the loading state yourself. [More info at nuxt docs](https://nuxtjs.org/docs/components-glossary/fetch/)

src/components/Item/CollectionTabs.vue
```diff
@@ -0,0 +1,69 @@
...
+
+<script lang="ts">
+import { BaseItemDto } from '@jellyfin/client-axios';
+import Vue from 'vue';
+import { mapGetters, mapActions } from 'vuex';
+
+export default Vue.extend({
+  props: {
+    item: {
+      type: Object,
```
ferferga, line 27: It's ugly but we will have better typing support with Vue 3 and Nuxt 3.

```suggestion
      type: Object as () => BaseItemDto,
```

src/store/items.ts
```diff
@@ -8,14 +8,21 @@ import merge from 'lodash/merge';
...
   byId: Record<string, BaseItemDto>;
   allIds: readonly string[];
+  boxSetChildrenById: Record<string, ItemsByType[]>;
 }
 
 export const defaultState = (): ItemsState => ({
   byId: {},
-  allIds: []
+  allIds: [],
+  boxSetChildrenById: {}
```
ferferga, line 25: Why we're treating the collection items differently? They're normal items, just with a parent.

ferferga (changes requested):
Looks much better than before!
Second round of code inspection until I actually test this out locally 🔎.

src/store/items.ts
doc-code-hub: I have added the fixes you pointed out to me above, 
however I have a problem with this one.


I was trying to modify the items store to add only child items, calling
```ts
response.Items? .forEach ((item) => {
      commit ('ADD_ITEM', {item});
    });
    commit ('ADD_ITEMS', {
      items: response.Items
    });
```
from the Success function of my action, to avoid having the 'boxSetChildrenById' property.


I modified my getter to return the appropriate data by extracting it from the state.byId but I found that the **parentId** of the elements I get by **filtering by parentId in the api** call, does _**not match**_ that of the collection (because in fact I believe it matches the reference library instead , as is correct).

**For these reasons**, I am **not** being able to recover the elements that are children of the collection from my getter (from state.byId), which is why I had opted to save them in this way.

This being the case (to avoid keeping duplicates in the store) it might be better to keep a similar property, but to save only the references (id) for collections to the state.byId?

What do you think? Suggestions?

ferferga: Fair enough and thanks for the finding. Let's leave it like this, code evolves and always can be revised later :)

But for now, let's not overengineer stuff.

doc-code-hub: Perfect! I leave it like this and throw local changes
Thanks for taking the time!

I am happy to help the project and I intend to continue doing it as much as I can when time permits!

I like the project and it is the first github project in which I collaborate :)

Ps. I noticed a test failed, but only for a failed call to some github service

doc-code-hub: In relation to this (3 comments back)
> I modified my getter to return the appropriate data by extracting it from the state.byId but I found that the **parentId** of the elements I get by **filtering by parentId in the api** call, does _**not match**_ that of the collection (because in fact I believe it matches the reference library instead , as is correct).
>

it would not be convenient to keep in mind this need to get the children elements of the collections regarding #801?

They seem to require proper handling as indicated in the diagram regarding _ShowSeasons_, don't you think?

ferferga: I don't understand what you mean, care to make an example?

Don't pay much attention to that diagram either, it's not final and we've not set ourselves so much in the need of some stores (like for the home sections).

ferferga (approved):
I tested this locally and LGTM, happy to merge as soon as I get a reply to my last comment [here](https://github.com/jellyfin/jellyfin-vue/pull/1680#discussion_r781649134)

src/store/items.ts
doc-code-hub: I mean that given the impossibility of obtaining the children of a particular element (except for the original folder) from the generic list of elements in the store, I believe the need arises to have an appropriate management of the collections (since having a "boxSetChildrenByType" property does not it is properly the competence of the item store, as you pointed out to me)

But it's probably best to bring the topic to the relative issue we were talking about

ferferga: @doc-code-hub Oh yes, so you mean leaving it exactly how it's right now in this PR, correct? For me the current implementation is fine for the time being, until we come up with something better with the whole refactor and everything it's redone.

ferferga:
@jellyfin-bot rebase

ferferga:
Thank you very much!

doc-code-hub:
Is a real pleasure! :)

PR #1682, opened 2022-01-09
fix(episode info): add season and episode index in 'Item' page with 3…

doc-code-hub:
## Micro fix

I add the season name and episode number in the episode detail.

image

I check the type of the item and if it is an episode, I show the information: {SeasonName, IndexNumber} 
just before the year (in the line below title)

src/components/Item/MediaInfo.vue
```vue
    <span v-if="item.Type === 'Episode'">
      {{ item.SeasonName }} &#8226; {{ item.IndexNumber }}        
    </span>
```

------------------------------

Just for the record, I misread this old closed issue (#1442) and made this fix, but I still find it useful. :)

### DIFF
--- a/src/components/Item/MediaInfo.vue
+++ b/src/components/Item/MediaInfo.vue
@@ -1,5 +1,13 @@
 <template>
   <div class="text--secondary">
+    <span v-if="item.Type === 'Episode'">
+      {{
+        $t('seasonEpisodeAbbrev', {
+          seasonNumber: item.ParentIndexNumber,
+          episodeNumber: item.IndexNumber
+        })
+      }}
+    </span>
     <span v-if="item.ProductionYear && year">{{ item.ProductionYear }}</span>
     <span v-if="item.OfficialRating && rating">{{ item.OfficialRating }}</span>
     <span v-if="item.CommunityRating && rating">


src/components/Item/MediaInfo.vue
```diff
@@ -1,5 +1,8 @@
 <template>
   <div class="text--secondary">
+    <span v-if="item.Type === 'Episode'">
+      {{ item.SeasonName }} &#8226; {{ item.IndexNumber }}
```
ferferga, line 4: I'd rather use the ``seasonEpisode`` or ``seasonEpisodeAbbrev`` translation keys for consistency.

ferferga (changes requested):

src/components/Item/MediaInfo.vue
doc-code-hub: Of course, you are right, I had not noticed the presence of this key.

I will use "seasonEpisodeAbbrev" to avoid unnecessary word wraps in mobile!

I'll come back soon :)

ferferga: Ammend and force repush your commit please! 👍🏻

doc-code-hub:
Updated as requested! 2878f8570a4a78db0ed175d78534ff64a9596735

image

_Code Changes_
```vue
<span v-if="item.Type === 'Episode'">
      {{
        $t('seasonEpisodeAbbrev', {
          seasonNumber: item.ParentIndexNumber,
          episodeNumber: item.IndexNumber
        })
      }}
</span>
```

src/components/Item/MediaInfo.vue
doc-code-hub: I'm not entirely sure what you mean by " force repush your commit ".

I made a new commit, before reading your new comment and I wonder if you were asking me to force overwrite the previous commit (which in case, I didn't know was possible).

What did you mean? A simple new commit or an overwrite somehow I don't know about?

ferferga: @doc-code-hub Yes, exactly that.

You can overwrite your previous commit using (with the files previously staged): ``git commit --amend --no-edit``
If also want to change the commit message, just replace ``--no-edit`` with ``-m "new commit message"``

If you want to do this with other commits (not your previous one), you can run ``git log``, scroll down to the previous commit to the one where you want to make changes, copy the hash and do ``git rebase -i *hash*``

A text document with all the possible rebase options will appear for you. You can drop, reorder or squash commits this way (squash is exactly what I wanted, to keep a more clear history).

No worries, I'll do it for you.

ferferga (approved):

src/components/Item/MediaInfo.vue
doc-code-hub: I see!

Helpful, I keep learning something from you from time to time.

I'll keep it in mind for the collections page I'm working on and use it instead of making a new commit for \#1680

ferferga: Making new commits per se it's not bad, the content itself of the commits is what you should be aware of :)

In this case, having two commits for a three line change is wasteful. But in your other PR, you can make commits for each stage you go by. Use a rule of thumb everything that's worth keeping for historical reasons. For instance:

1. Commit cleaning up unused vars from your page
2. Commit removing the component store
3. Commit migrating everything to the item page store.
4. Whatever else you think it's worth snapshotting.

This way we still keep your original implementation in the commit history, which might be useful in the future if we want to know why we ended up having a feature in X way instead of Y way.

So no worries to do commits, just made them when they're needed, and in this case it clearly wasn't :).

Issue #1683, opened 2022-01-10
[Docker - nginx] 404 when refreshing detail pages (/*/*)


doc-code-hub:
### Description of the bug

I ran the latest unstable image on docker, on 80:80 but I have a problem.

Every time I refresh the page, this from 404, except for the paths / and / * /

Excluding Login and Host page, everything that is /*/* returns 404 (summing up, the detail pages).

Making a list:

It works
- Home
- Login / Host
- Settings
- Metadata

It does not work
- library/*
- item/*
- series/*

I don't have some types of elements and I don't know if there are other specific paths, however I think it is also reflected in any other detail pages.

### Steps to reproduce

docker pull jellyfin/jellyfin-vue:unstable
docker run -p 80:80 jellyfin/jellyfin-vue:unstable
Go to localhost (or wathever), login
Go to some library or item or series page
Refresh (F5) 

![image](https://user-images.githubusercontent.com/61033694/148853501-807016f1-4cc7-4cb3-94a2-78fb1a70f671.png)

### Expected behavior

The page reloads correctly

### Screenshots

![image](https://user-images.githubusercontent.com/61033694/148853819-d279ca3f-d65e-4ce9-80ec-389413d4315c.png)

### Platform

Windows

### Browser

Chrome

### Jellyfin server version

10.7.7

ferferga:
This is something I thought about when casually exploring the Vue router docs for some information about the meta tags, so I think I might be helpful :).

I think the cause it's that [Vue router uses hashing mode by default](https://router.vuejs.org/guide/essentials/history-mode.html#example-server-configurations). Can you check by changing the [mode](https://router.vuejs.org/api/#mode) to ``history``, as we no longer need abstract for fallback since we don't use SSR anymore?

All browsers (even really ancient versions, which we no longer plan tu support) [supports the HTML5 History API](https://caniuse.com/mdn-api_history) so this shouldn't be a problem.

ferferga:
This is something I thought about when casually exploring the Vue router docs for some information about the meta tags, so I think I might be helpful :).

I think the cause it's that [Vue router doesn't use hash mode, as there's no # appended to our urls (probably a Nuxt thing, as it's their default supposedly)](https://router.vuejs.org/guide/essentials/history-mode.html). Can you check by changing the [mode](https://router.vuejs.org/api/#mode) to ``hash``, as we no longer need abstract for fallback since we don't use SSR anymore?

doc-code-hub:
[...]

PR #1684, opened 2022-01-11
fix(router): 404 on reloads

ferferga:
Turns out this was easier than I expected. Should fix #1683 

I tested a few routes and all of them worked but items, so I fixed their path resolution. I don't know if somewhere else changes are needed. @doc-code-hub Care to pull this branch and test extensively if all your tests pass now?

PR #1685, opened 2022-01-11
fix(nginx): fix 404 when refreshing page with custom nginx.conf

doc-code-hub:
Refering to #1683 i've made this simple edit by adding a custom nginx.conf and adding it into the Dockerfile.

I started from the default nginx.conf ([link](https://gist.githubusercontent.com/nishantmodak/d08aae033775cb1a0f8a/raw/ebea0ae66e0a0188009accae2acba88cc646ddcd/nginx.conf.default)) and I modified some parts by adding in particular:
```nginx
 location / {
           # First attempt to serve request as file, then
           # as directory, then fall back to redirecting to index.html
           try_files $uri $uri/ $uri.html /index.html;
        }
```

This requires no action by users to deploy and solve the problem!

### DIFF
--- a/.docker/nginx.conf
+++ b/.docker/nginx.conf
@@ -1,14 +1,11 @@
-    server {
-        # You would want to make a separate file with its own server block for each virtual domain
-        # on your server and then include them.
-        listen 80;
+server {
+    listen 80;
 
-        root /usr/share/nginx/html;
+    root /usr/share/nginx/html;
 
-        location / {
-           # First attempt to serve request as file, then
-           # as directory, then fall back to redirecting to index.html
-           # This is needed for history mode in Vue router
-           try_files $uri $uri/ /index.html;
-        }
+    location / {
+        # First attempt to serve request as file, then as directory, then fall back to redirecting to index.html
+        # This is needed for history mode in Vue router
+        try_files $uri $uri/ /index.html;
     }
+}
--- a/.docker/nginx.conf
+++ b/.docker/nginx.conf
@@ -1,15 +1,14 @@
     server {
         # You would want to make a separate file with its own server block for each virtual domain
         # on your server and then include them.
-        listen       80;
-        #tells Nginx the hostname and the TCP port where it should listen for HTTP connections.
-        # listen 80; is equivalent to listen *:80;
+        listen 80;
 
         root /usr/share/nginx/html;
 
         location / {
            # First attempt to serve request as file, then
            # as directory, then fall back to redirecting to index.html
+           # This is needed for history mode in Vue router
            try_files $uri $uri/ /index.html;
         }
     }
--- a/.docker/nginx.conf
+++ b/.docker/nginx.conf
@@ -1,37 +1,15 @@
-
-worker_processes  1;
-
-events {
-    worker_connections  1024;
-    # worker_processes and worker_connections allows you to calculate maxclients value: 
-    # max_clients = worker_processes * worker_connections
-}
-
-
-http {
-    include       mime.types;
-
-    default_type  application/octet-stream;
-
-    sendfile        on;
-
-    keepalive_timeout  65;
-
     server {
         # You would want to make a separate file with its own server block for each virtual domain
         # on your server and then include them.
         listen       80;
         #tells Nginx the hostname and the TCP port where it should listen for HTTP connections.
         # listen 80; is equivalent to listen *:80;
-        
-        server_name  localhost;
 
         root /usr/share/nginx/html;
 
         location / {
            # First attempt to serve request as file, then
            # as directory, then fall back to redirecting to index.html
-           try_files $uri $uri/ $uri.html /index.html;
+           try_files $uri $uri/ /index.html;
         }
     }
-}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,11 @@
 ## This dockerfile builds the client entirely in a Docker context
 
+
 FROM node:16-alpine AS build
+
 # Set environment variables
 ARG ROUTER_MODE_HISTORY=false
+ENV ROUTER_MODE_HISTORY=$ROUTER_MODE_HISTORY
 
 ARG IS_STABLE=0
 
@@ -17,6 +20,9 @@ WORKDIR /app
 # Copy files to workdir
 COPY . .
 
+# Set router mode variable
+# RUN if [[ $ROUTER_MODE_HISTORY == 'true' ]] ; then ROUTER_MODE_HISTORY="true" ; fi
+
 # Install dependencies
 RUN npm ci --no-audit
 
@@ -25,33 +31,27 @@ RUN if [[ $IS_STABLE == "0" ]] ; then NUXT_ENV_COMMIT=$(git rev-parse HEAD) ; fi
 
 
 # Build client
-RUN if [[ $ROUTER_MODE_HISTORY == 'true' ]] ; then \ 
-echo "building with router.mode = history" ; \
-ROUTER_MODE_HISTORY="true" | npm run build ; \
-else \
-echo "building with router.mode = hash" ; \
-npm run build ; \
-fi
-
+RUN npm run build
 
 
 # Deploy built distribution to nginx
 FROM nginx:alpine
 
+# Set environment variables
 ARG ROUTER_MODE_HISTORY=false
 
 COPY --from=build /app/src/dist/ /usr/share/nginx/html/
 
-COPY --from=build /app/nginx.conf /etc/nginx/custom.conf 
+COPY --from=build /app/.docker/nginx.conf /etc/nginx/conf.d/custom.conf 
 
 # If specified, set's custom nginx.conf to support spa (for 'history' routing mode)
 RUN  if [[ $ROUTER_MODE_HISTORY == 'true' ]] ; then \
   echo "configuring nginx for router.mode = history" ; \
-  mv /etc/nginx/custom.conf /etc/nginx/nginx.conf  ; \
+  mv /etc/nginx/conf.d/custom.conf /etc/nginx/conf.d/default.conf  ; \
 else \
+  rm /etc/nginx/conf.d/custom.conf ; \
   echo "standard nginx configuration for router.mode = hash" ; \
 fi
-
 
 
 EXPOSE 80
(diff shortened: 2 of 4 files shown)


ferferga:
This fixes Docker but leaves users that deploy their client using their own reverse proxy with the need to tweak their own config (which means reading a documentation, etc, and not everybody is willing to do so, when we're aiming to make this as simple to use as possible).

This doesn't also cover the cases for when the client is hosted under a Jellyfin server (like the current web client does), which is another situation [we might need to cover](https://github.com/jellyfin/jellyfin-vue/discussions/1679)

PS: We also used our custom nginx config before, but we removed it because [it was yet another thing to maintain for ourselves](https://github.com/jellyfin/jellyfin-vue/pull/1645)

Please refer to #1684 and, if that fixes your issue, please close this pull request.

ThibaultNocchi:
Hey, so we're going this way:

We'll move to hash by default. But, "history" mode (that's how current routing is named) will be toggleable with an env var. It'll also be used in our Dockerfiles. So we'll first merge ferf PR to set the new default, and we'll continue with you on this PR. :)

ThibaultNocchi:
So, I see a few points:

* You should copy the Nginx conf to `/etc/nginx/conf.d/nginx.conf` so you can have an easier config
* That easier config can drop everything except the `server` block. In that one you can simplify it to:

```
    server {
        # You would want to make a separate file with its own server block for each virtual domain
        # on your server and then include them.
        listen       80;
        #tells Nginx the hostname and the TCP port where it should listen for HTTP connections.
        # listen 80; is equivalent to listen *:80;

        root /usr/share/nginx/html;

        location / {
           # First attempt to serve request as file, then
           # as directory, then fall back to redirecting to index.html
           try_files $uri $uri/ /index.html;
        }
    }
```

So we dropped the server_name as that *should* work without and edited the try_files to match VueRouter config file https://router.vuejs.org/guide/essentials/history-mode.html#nginx

Tell me if you need help figuring out how to enable "history" mode with env vars. :)

doc-code-hub:
Fantastic, I like the possibility of being able to choose what you propose!

Later I will try to implement the change (and the nginx.conf update as you suggest)! (around 13:00 Italian time)

If I have trouble figuring it out, I'll definitely come back here to ask for some help :)

Thank you!

ferferga:
@doc-code-hub Btw in case you aren't yet signed up in Matrix, present yourself in our [Vue development room](https://matrix.to/#/#jellyfin-vue:matrix.org) (we also have it bridged to Discord) so we can discuss this kind of stuff in a more informal way!

doc-code-hub:
Very gladly :)

Later I will definitely do it, thanks!

Dockerfile
```diff
@@ -20,15 +26,35 @@ RUN npm ci --no-audit
...
 FROM nginx:alpine
+
+# Set environment variables
+ARG HISTORY_ROUTER_MODE=false
+
 COPY --from=build /app/src/dist/ /usr/share/nginx/html/
+
+COPY --from=build /app/.docker/nginx.conf /etc/nginx/conf.d/custom.conf 
+
+# If specified, set's custom nginx.conf to support spa (for 'history' routing mode)
```
ferferga, line 44: I would honestly remove all this block and just copy our custom config no matter the history mode. The default config gives the same end result to the user (the only difference is the try block) and this just clutters our Dockerfile for the little benefit it provides 

*(now at least, maybe in the future it's useful for E2E testing)*

Thoughts @ThibaultNocchi?

doc-code-hub: I was thinking that a custom nginx.conf whoud be a problem for some cases [(your comment)](https://github.com/jellyfin/jellyfin-vue/pull/1685#issuecomment-1009509281)

No?

ferferga: Oh yes, but I meant users that dont use Docker (not everyone does).  
If an user builds the client himself or pulls it as a deb (like current Web) or from a release (we don't have debs or releases, but once we reach an stable state we will) and serves the client through his webserver of choice, the user must configure it with the ``try_files`` directive.


Docker in that regard its our walled garden and we can provide our best config where everything is beautiful. But we can't expect everybody will have our ideal configuration, hence the need for the default to be hash. 

In this repo we are "selling" users two "bundles":

* The client (our "best-seller", as its what the repo is about) 

* Client + webserver (Docker) 

This repo is all about the client and, as is, it should work as universally as possible with the default config. History mode doesn't allow that as it adds the user's webserver into the mix, and adds variables out of our control. An user with a misconfigured webserver will have a degraded experience, so we should stick to "the universal best". 

However, we're offering too (through Docker) the mix client + webserver: with it, we remove the webserver variable from the user's hands and bring it to us. This means that for users choosing our bundle, we can give them what we think are the best defaults (we all agree is history mode). 

To sum up: the file does not hurt and provides "universally" safe defaults. I also expect nobody in Docker will use the hash mode, because it makes no sense for them (most people just fire docker build or pull the image, and those will use history mode by default). 
I only expect people building outside Docker to care about changing that option, since those are the ones who are likely not going to use "our walled garden" and are more likely to be scatterbrained about their webserver config. 

That's why the current config is ideal: outside docker we use hash, inside it we use history. The environments where they both make more sense.

doc-code-hub: Then in this case it is better to change the default ARG to true, 
to pass the --build-arg only in the unlikely event that you want to disable. Do you agree?

bc now default is 'hash' for docker too

Dockerfile
```diff
@@ -20,15 +26,22 @@ RUN npm ci --no-audit
...
+
+COPY --from=build /app/.docker/nginx.conf /etc/nginx/conf.d/default.conf 
+
 EXPOSE 80
 
 # Set labels
 LABEL maintainer="Jellyfin Packaging Team - packaging@jellyfin.org"
 LABEL org.opencontainers.image.source="https://github.com/jellyfin/jellyfin-vue"
 LABEL org.opencontainers.image.description "Commit: ${NUXT_ENV_COMMIT}"
+LABEL org.opencontainers.image.description "History Router Mode: ${HISTORY_ROUTER_MODE}"
```
ferferga, line 47: Doing duplicates remove the previous one: https://docs.docker.com/config/labels-custom-metadata/#label-keys-and-values

```suggestion
LABEL org.opencontainers.image.description="Commit: ${NUXT_ENV_COMMIT} History router mode: ${HISTORY_ROUTER_MODE}"
```

.docker/nginx.conf
```diff
@@ -0,0 +1,15 @@
+    server {
+        # You would want to make a separate file with its own server block for each virtual domain
+        # on your server and then include them.
+        listen       80;
```
ferferga, line 4:
```suggestion
        listen 80;
```

.docker/nginx.conf
```diff
@@ -0,0 +1,15 @@
...
+        #tells Nginx the hostname and the TCP port where it should listen for HTTP connections.
+        # listen 80; is equivalent to listen *:80;
+
+        root /usr/share/nginx/html;
+
+        location / {
+           # First attempt to serve request as file, then
+           # as directory, then fall back to redirecting to index.html
```
ferferga, line 12:
```suggestion
           # First attempt to serve request as file, then
           # as directory, then fall back to redirecting to index.html
           # This is needed for history mode in Vue router
```

.docker/nginx.conf
```diff
@@ -0,0 +1,15 @@
+    server {
+        # You would want to make a separate file with its own server block for each virtual domain
+        # on your server and then include them.
+        listen       80;
+        #tells Nginx the hostname and the TCP port where it should listen for HTTP connections.
+        # listen 80; is equivalent to listen *:80;
```
ferferga, line 6: Not relevant for us

```suggestion
```

.docker/nginx.conf
```diff
@@ -0,0 +1,15 @@
+    server {
+        # You would want to make a separate file with its own server block for each virtual domain
+        # on your server and then include them.
```
ferferga, line 3: Not relevant for us

```suggestion
```

Dockerfile
```diff
@@ -1,8 +1,14 @@
...
+
 FROM node:16-alpine AS build
+
 # Set environment variables
+ARG HISTORY_ROUTER_MODE=true
+ENV HISTORY_ROUTER_MODE=$HISTORY_ROUTER_MODE
+
 ARG IS_STABLE=0
+
 ENV NUXT_ENV_COMMIT=""
```
ferferga, line 12: Keep them together and remove the excess of spaces:

```suggestion
FROM node:16-alpine AS build

# Set build arguments
ARG HISTORY_ROUTER_MODE=true
ARG IS_STABLE=0

# Set environment variables
ENV HISTORY_ROUTER_MODE=$HISTORY_ROUTER_MODE
ENV NUXT_ENV_COMMIT=""
```

Dockerfile
```diff
@@ -20,15 +26,22 @@ RUN npm ci --no-audit
 # Set commit hash
 RUN if [[ $IS_STABLE == "0" ]] ; then NUXT_ENV_COMMIT=$(git rev-parse HEAD) ; fi
 
+
```
ferferga, line 29:
```suggestion
```

Dockerfile
```diff
@@ -20,15 +26,22 @@ RUN npm ci --no-audit
...
 # Build client
 RUN npm run build
 
+
```
ferferga, line 33:
```suggestion
```

Dockerfile
```diff
@@ -20,15 +26,22 @@ RUN npm ci --no-audit
...
 # Deploy built distribution to nginx
 FROM nginx:alpine
+
 COPY --from=build /app/src/dist/ /usr/share/nginx/html/
+
```
ferferga, line 38:
```suggestion
```

ferferga (changes requested):
Some "linting" stuff for our Dockerfile and then LGTM

Dockerfile
ThibaultNocchi: I agree that we can copy the config in any case, it's basic SPA routing anyway. :)

ThibaultNocchi:
I'd like to add something: for the env var, drop the true or false value and keep it as bash scripts usually do: either you put fill the env var to enable a feature `HISTORY_ROUTER_MODE=1` or you don't fill it and so the default is used. That'd also simplify the JS file to just detect the presence of the env var.

ferferga (approved):

ferferga (approved):

doc-code-hub:
@ThibaultNocchi `That'd also simplify the JS file to just detect the presence of the env var.`

Just for the record:
I tried to limit myself to checking the presence of the variable ( 
```ts
provess.env.HISTORY_ROUTER_MODE ? 'history' : 'hash'
```
), but I still had to keep the check 
```ts
provess.env.HISTORY_ROUTER_MODE === '1'
```
 because the environment variable arrives as a string, for this reason, both '1' and '0' were true ( using history mode).

Just to let you know

ThibaultNocchi (approved):

PR #1686, opened 2022-01-11
feat(login): focus forms on login pages automatically

ferferga:
It's annoying the need to go and click the field manually. ith this, we're mimicking all other login forms that do this (i.e Google, Amazon, etc...)

### DIFF
--- a/src/components/Forms/AddServerForm.vue
+++ b/src/components/Forms/AddServerForm.vue
@@ -10,6 +10,7 @@
           <v-text-field
             v-model="serverUrl"
             outlined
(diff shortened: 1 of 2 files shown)


ThibaultNocchi (commented):
It can be easier than that: HTML and Vuetify support the `autofocus` prop on inputs https://developer.mozilla.org/fr/docs/Web/HTML/Global_attributes/autofocus https://vuetifyjs.com/en/api/v-text-field/#props

ThibaultNocchi (approved):

PR #1694, opened 2022-01-15
Regex url fix

### LINKED ISSUE #1688: Add server without specifying a protocol
### Not quite a bug, more a mini-feature request:

As allowed by the original website, I think it would be nice and convenient, to allow the insertion of an IP address (without protocol) with any specified port.

Taking it a step further, it could still improve over the original site (which allows any text in the field, and tries the call) by updating the current validation regex, with one that actually covers only the interesting cases (and maybe adding a default http protocol beahind the scenes if needed).

I don't have enough RegEx skills to propose a valid and consistent change.

Does anyone have an interest in dealing with it?

### Steps to reproduce

I.E.  192.168.1.5 does not allow adding

### Expected behavior

I.E.  192.168.1.5 as input should work fine! 

### Screenshots 

_From the original website_

image

MoraOgunro:
Add Server form now accepts URLs without protocol.

Modified the regex of the url validator to accept server URLs that do not have a port or an HTTP protocol.
The server form will now add a default port (8086) and protocol (http://) if none are present.

fix #1688

### DIFF
--- a/src/components/Forms/AddServerForm.vue
+++ b/src/components/Forms/AddServerForm.vue
@@ -48,7 +48,7 @@ import { mapActions, mapState } from 'vuex';
 export default Vue.extend({
   data() {
     return {
-      serverUrl: '',
+      serverUrl: '192.168.4.122',
       loading: false,
       rules: {
         serverUrl: {
@@ -75,6 +75,14 @@ export default Vue.extend({
     ...mapActions('servers', ['connectServer']),
     async connectToServer(): Promise<void> {
       this.loading = true;
+
+      if (!this.serverUrl.match(/^https?:\/\//)) {
+        this.serverUrl = 'http://' + this.serverUrl;
+      }
+
+      if (!this.serverUrl.match(/:[0-9]+$/)) {
+        this.serverUrl = this.serverUrl + ':8096';
+      }
 
       try {
         await this.connectServer(this.serverUrl);
--- a/src/components/Forms/AddServerForm.vue
+++ b/src/components/Forms/AddServerForm.vue
@@ -48,7 +48,7 @@ import { mapActions, mapState } from 'vuex';
 export default Vue.extend({
   data() {
     return {
-      serverUrl: '192.168.4.122',
+      serverUrl: '',
       loading: false,
       rules: {
         serverUrl: {
--- a/src/plugins/nuxt/veeValidate.ts
+++ b/src/plugins/nuxt/veeValidate.ts
@@ -6,7 +6,7 @@ import { required } from 'vee-validate/dist/rules';
 extend('required', required);
 
 extend('mustBeUrl', (value: string): boolean => {
-  return /^https?:\/\/.+/.test(value);
+  return /^(https?:\/\/)?[0-9.]+(:[0-9]+)?/.test(value);
 });
 
 extend('bothPasswordsSame', {


doc-code-hub:
In my opinion, the default port should be 80 (regardless of the jellyfin server default port).

Do not you think? :)

T-bond:
> In my opinion, the default port should be 80 (regardless of the jellyfin server default port).
> 
> Do not you think? :)
> 
> For example I prefer to keep my jellyfin server deploy on 80 to use it comfortably from the local network by specifying only an IP and map it to another port possibly to expose it (but not too much).
> 
> This is why in my case it would be an inconvenience (and also different from what happens when you don't normally specify a port in http).
> 
> I also believe that it would not be clear for users

I have the same concerns as @doc-code-hub stated. I am also not using the default port.
A better solution would be maybe, to first try the user provied URL, and if that not matches try with the default Jellyfin port.
For the http protocol try the default 80, for https 443.

Also maybe I am wrong about this, but won't the new regex: `^(https?:\/\/)?[0-9.]+(:[0-9]+)?` only allow IP addresses and not DNS records? Wouldn't it cause problems?

src/components/Forms/AddServerForm.vue
```diff
@@ -76,6 +76,14 @@ export default Vue.extend({
...
       this.loading = true;
 
+      if (!this.serverUrl.match(/^https?:\/\//)) {
+        this.serverUrl = 'http://' + this.serverUrl;
+      }
+
+      if (!this.serverUrl.match(/:[0-9]+$/)) {
+        this.serverUrl = this.serverUrl + ':8096';
+      }
+
```
doc-code-hub, line 86: Mmm. 
_**When this PR will be completed**_, I would like to use this fallback logic in that of the [Default Servers](https://github.com/jellyfin/jellyfin-vue/pull/1698) (moving it to a utils function) to apply it also to the addresses specified in the environment variable.

But still think we can always skip the port fallback  :)

ThibaultNocchi: I don't think you should interpret URLs from the env var too much, _for me_ they should be as standard as possible

doc-code-hub: If you think better I can take it off.

I did this to provide an adequate warn, if for example you provide a list of URLs one of which has a typo.

Only to facilitate the finding of the specific occurrence, indicating the exact problematic address.

It seemed useful to me, didn't you think @ThibaultNocchi ?

ThibaultNocchi:
So I'm gonna say what the others said already:

- The validation regex will only let IP go through
- I'm not sure if implicitly adding protocols and ports is good. What I'd see, but that's up to discussion, is that the client tries the different classic protocols and ports when typing something lacking one of them.
  - `HOST`: it tries `https://HOST:443`, `http://HOST:80`, `http://HOST:8096` in that order
  - `http://HOST`: it tries  `http://HOST:80`, `http://HOST:8096` in that order
  - `HOST:PORT`: it tries `https://HOST:PORT`, `http://HOST:PORT` in that order

But that's my own take, and may not be optimal at all haha. In the best world I'd prefer it the current way, meaning specifying https or http and adding the port if it's not 443 or 80. But I understand that's it may not be the best way.

T-bond:
> So I'm gonna say what the others said already:
> 
>     * The validation regex will only let IP go through
> 
>     * I'm not sure if implicitly adding protocols and ports is good. What I'd see, but that's up to discussion, is that the client tries the different classic protocols and ports when typing something lacking one of them.
>       
>       * `HOST`: it tries `https://HOST:443`, `http://HOST:80`, `http://HOST:8096` in that order
>       * `http://HOST`: it tries  `http://HOST:80`, `http://HOST:8096` in that order
>       * `HOST:PORT`: it tries `https://HOST:PORT`, `http://HOST:PORT` in that order
> 
> 
> But that's my own take, and may not be optimal at all haha. In the best world I'd prefer it the current way, meaning specifying https or http and adding the port if it's not 443 or 80. But I understand that's it may not be the best way.

Maybe on `HOST` also try `https://HOST:8096` before `http://HOST:8096`?

ThibaultNocchi:
Yes that's a nice idea, though I'd use the SSL default port of JF, ie. 8920. I'll edit my message :)

src/components/Forms/AddServerForm.vue
```diff
@@ -76,6 +76,14 @@ export default Vue.extend({
     async connectToServer(): Promise<void> {
       this.loading = true;
 
+      if (!this.serverUrl.match(/^https?:\/\//)) {
+        this.serverUrl = 'http://' + this.serverUrl;
+      }
+
+      if (!this.serverUrl.match(/:[0-9]+$/)) {
+        this.serverUrl = this.serverUrl + ':8096';
+      }
```
ferferga, line 85: Any reason to add the rules here instead of doing it in vee-validate?

ferferga (changes requested):

MoraOgunro:
Thanks for the insight guys. Admittedly, I didn't know much about IPs and Domains. I'm going to create an array of possible addresses [according to this list](https://github.com/jellyfin/jellyfin-vue/pull/1694#issuecomment-1013863062) and try to connect with each of them. 

I'm going to leave DNS records alone

src/components/Forms/AddServerForm.vue
MoraOgunro: No. I will move them

PR #1698, opened 2022-01-15
[Default Servers] Specify a server list at build time

### LINKED ISSUE #1603: [Enhancement] Ability to predefine a jellyfin server in the docker-compose.yaml
### Description of the bug

It would be great if a predefined jellyfin server could be added via an ENV variable in the docker-compose file. 

### Steps to reproduce

--

### Expected behavior

---

### Logs

_No response_

### Screenshots

_No response_

### Platform

Linux

### Browser

Chrome

### Jellyfin server version

10.8.0

### Additional context

_No response_

doc-code-hub:
fix #1603

Added support for the "DEFAULT_SERVERS" **environment variable** as a **comma separated url string**.   

## Possible uses

_/src/.env_
```.env
DEFAULT_SERVERS=http://192.168.1.1:8080, https://example.com
```

Docker Build (overrides /src/.env)
```bash
docker build --tag jellyfin-vue --build-arg DEFAULT_SERVERS=http://192.168.1.1 .
```

docker-compose.yml  (overrides /src/.env)
```docker-compose.yml
version: '3'
services:
  app:
    container_name: jellyfin-vue
    restart: always
    build:
      context: .
      args:
        - DEFAULT_SERVERS=http://192.168.1.1, https://example.com
    ports:
      - '80:80'

```
---

The publicInfo of the specified servers are **fetched simultaneously** (**_ONLY if missing from local storage_**). 

### Default
- If only one server is specified, it will be set as the default, bypassing the add / choose server step.
- If two or more servers are specified, the first time, you will be redirected to the server selection page.

###  All default servers, as the community thinks best, will not have a chance to be deleted.
image

 
 ### The login page now displays the server name
 
image


doc-code-hub:
I'm fixing for the test that fails

doc-code-hub:
in any case, and I speak with the mainteiners (who merge), be careful because I moved the url validation ReGex in utils / server.ts (to be able to use it to validate the url, without duplication)

So I mark to be careful in this [PR](https://github.com/jellyfin/jellyfin-vue/pull/1694) just in relation to this :)

As it has been updated in that PR (but I had moved it from where it was replaced)

src/nuxt.config.ts
```diff
@@ -3,6 +3,14 @@ import type { NuxtConfig } from '@nuxt/types';
 import { siJellyfin as jellyfinIcon } from 'simple-icons/icons';
 
 const config: NuxtConfig = {
+  /*
+   ** Nuxt Environment Variables
+   ** See https://nuxtjs.org/docs/configuration-glossary/configuration-env
+   */
+  env: {
+    // Comma separated list of predefined servers (I.e. 'http://192.168.1.1' or 'http://192.168.1.1, https://your.server.domain')
+    server_url_list: process.env.DEFAULT_SERVERS || ''
```
ThibaultNocchi, line 12: I guess the OR could be omitted, that'd only make the `server_url_list` undefined I guess?

src/utils/servers.ts
```diff
@@ -0,0 +1,42 @@
...
+      serverUrlList.push(url);
+    } else {
+      console.warn(
+        `[Default Servers] This url is not valid. It will be excluded from the server list to avoid errors. (${url})`
+      );
+    }
+  }
+
+  const splitByComma = value.split(',');
+  const trimmedSplit = splitByComma.map((url) => url.trim());
```
ThibaultNocchi, line 34:
```suggestion
  const trimmedSplit = value.split(',').map((url) => url.trim());
```

I wouldn't mind it being merged

src/utils/servers.ts
```diff
@@ -0,0 +1,42 @@
...
+  function checkAndAddUrl(url: string): void {
+    if (urlRegEx.test(url)) {
+      // Valid Url
+      serverUrlList.push(url);
+    } else {
+      console.warn(
+        `[Default Servers] This url is not valid. It will be excluded from the server list to avoid errors. (${url})`
+      );
+    }
+  }
```
ThibaultNocchi, line 31: You can just put that function directly in the foreach below, at least that's what I'd do. @ferferga ?

src/plugins/nuxt/appInitPlugin.ts
```diff
@@ -1,10 +1,81 @@
...
+    const defaultServer = fullInfoServers[0];
+
+    if (missingServers.includes(defaultServer.address)) {
+      context.store.commit('servers/SET_SERVER_USED', {
+        ...defaultServer,
+        isDefault: true
+      });
+    }
+
+    context.$axios.setBaseURL(defaultServer.address);
```
ThibaultNocchi, line 78: I wouldn't do it personally, if you set up 2+ defaults servers you'd want to pick your preferred I guess. And that makes the edit you did to the middleware not working. (Well the >1 redirect doesn't work for me, and I guess that must be because of that) :)

src/nuxt.config.ts
doc-code-hub: Actually env is of type Record <string, string> if you go to see the NuxtOptionsEnv interface :)

Only for this

src/utils/servers.ts
doc-code-hub: Fair enough :)

src/utils/servers.ts
doc-code-hub: You're right, initially I was using it in two places, that's why I pulled it out.

Now I only use it once, I can actually move it back directly into the forEach

src/plugins/nuxt/appInitPlugin.ts
doc-code-hub: Uh, you're right, I just fixed ...

I'll be back in a few minutes with a commit with your bug fixes :)

src/utils/servers.ts
ThibaultNocchi: It makes sense then :)

src/nuxt.config.ts
ThibaultNocchi: No don't worry, I haven't done my homework, you're right ^^ Then maybe default to `undefined` instead of empty string?

doc-code-hub: But will have the same problem bc is not a string no? :)

I pick '' cause it's a string and returns true in if (!value) condition (which is the one i use in the parse function to fallback to [])

ThibaultNocchi: Ok I'm stupid. Sorry ^^

ThibaultNocchi (approved):

ThibaultNocchi:
LGTM, I'll let @ferferga quickly review it but that works well at least :)

src/pages/server/login.vue
```diff
@@ -36,10 +36,17 @@
...
-        </h1>
-        <h1 v-else class="text-h4 mb-6 text-center">{{ $t('login.login') }}</h1>
+        <v-col align="center" justify="center" dense class="mt-6 mb-3">
+          <h1 v-if="!isEmpty(currentUser)" class="text-h4 mb-3 text-center">
+            {{ $t('login.loginAs', { name: currentUser.Name }) }}
+          </h1>
+          <h1 v-else class="text-h4 mb-3 text-center">
+            {{ $t('login.login') }}
+          </h1>
+          <label class="server-used">{{
```
ThibaultNocchi, line 46: We tend to prefer using Vuetify classes, as above. So you'd use the `text-center` class at least. For the opacity you can pick [here](https://vuetifyjs.com/en/styles/text-and-typography/#opacity) (I'd think `text--secondary` or `text--disabled` would suit you) and for the sizing something from [here](https://vuetifyjs.com/en/styles/text-and-typography/#typography) (so I'd guess something like `text-subtitle-1` for instance)

src/pages/server/login.vue
```diff
@@ -36,10 +36,17 @@
         md="6"
         lg="5"
       >
-        <h1 v-if="!isEmpty(currentUser)" class="text-h4 mb-6 text-center">
-          {{ $t('login.loginAs', { name: currentUser.Name }) }}
-        </h1>
-        <h1 v-else class="text-h4 mb-6 text-center">{{ $t('login.login') }}</h1>
+        <v-col align="center" justify="center" dense class="mt-6 mb-3">
```
ThibaultNocchi, line 39: Dropping the v-col and using a block instead of label (say a h1) should work :)

src/pages/server/login.vue
```diff
@@ -36,10 +36,17 @@
...
+          <h1 v-if="!isEmpty(currentUser)" class="text-h4 mb-3 text-center">
```
ThibaultNocchi, line 40: I _think_ the 2 mb-3 are a little too much, I'd reduce them but that's up to you :)

ThibaultNocchi (commented):
Can you apply the same subtitle to the login page of public users? (around L10)

src/pages/server/login.vue
```diff
@@ -36,10 +36,15 @@
         md="6"
         lg="5"
       >
-        <h1 v-if="!isEmpty(currentUser)" class="text-h4 mb-6 text-center">
-          {{ $t('login.loginAs', { name: currentUser.Name }) }}
+        <h1 v-if="!isEmpty(currentUser)" class="text-h4 mb-3 text-center">
+          {{ $t('logn.loginAs', { name: currentUser.Name }) }}
```
ThibaultNocchi, line 40:
```suggestion
          {{ $t('login.loginAs', { name: currentUser.Name }) }}
```

Typo :)

src/pages/server/login.vue
```diff
@@ -123,3 +129,10 @@ export default Vue.extend({
   }
 });
 </script>
+
+<style scoped>
+.server-used {
+  text-align: center;
+  opacity: 0.6;
+}
+</style>
```
ThibaultNocchi, line 138: It can now be removed

src/middleware/__tests__/serverMiddleware.spec.ts
```diff
@@ -48,21 +50,21 @@ describe('serverMiddleware', () => {
     expect(mockRedirect.mock.calls[0][0]).toBe(EXPECTED_REDIRECT);
   });
 
-  it('redirects to "/server/add" when the serverlist length is zero', (): void => {
-    serverMiddleware(INPUT_WITH_BASEURL);
+  it('redirects to "/server/add" when baseUrl is set the serverlist length is zero', (): void => {
```
ferferga, line 53:
```suggestion
  it('redirects to "/server/add" when baseUrl is set and the serverlist length is zero', (): void => {
```

src/middleware/serverMiddleware.ts
```diff
@@ -13,5 +13,10 @@ export default function (context: Context): void {
     !context.store.state.servers.serverList.length
   ) {
     return context.redirect('/server/add');
+  } else if (
+    context.store.state.servers.serverList.length > 1 &&
+    !context.store.state.servers?.serverUsed?.address
```
ferferga, line 18: I think this condition doesn't belong here? If no server is selected you should be redirected to the selection page I'd guess?

src/pages/server/login.vue
ferferga: We usually also have Vuetify classes for this, so check them out first: https://vuetifyjs.com/en/styles/text-and-typography/

src/plugins/nuxt/appInitPlugin.ts
```diff
@@ -1,10 +1,84 @@
...
+          console.warn('[Default Servers] Unreachable server: ', serverUrl);
+
+          resolve(null);
+        });
+    });
+
+    promises.push(promise);
+  }
+
+  await Promise.all(promises);
```
ferferga, line 55: Disabling the rules that enforce our style doesn't seem like the best idea :)

Check how you can accomplish the Promise.all thing properly here: https://github.com/jellyfin/jellyfin-vue/blob/master/src/store/index.ts#L67-L86

src/utils/servers.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  // Url Validity Check
+  trimmedSplit.forEach((url: string) => {
+    if (urlRegEx.test(url)) {
+      // Valid Url
+      serverUrlList.push(url);
+    } else {
+      console.warn(
+        `[Default Servers] This url is not valid. It will be excluded from the server list to avoid errors. (${url})`
+      );
```
ferferga, line 27: Here you can add the eslint ignore, as this is a warning and this console statement is indeed useful :)

ferferga (changes requested):

src/middleware/serverMiddleware.ts
doc-code-hub: @ferferga 
Ok, you are right, the login page, redirect to the /select page when the call fails.

BUT However it does it after trying a call with an empty address (in case of 2+ servers in the list) then calling the same URL as Vue (obviously without results).

To avoid this waste of time in the login page (since I know that in this context an empty call is tried), since the Middleware server was already implemented in the login, I thought to add this case and force the redirect to / select.

This translates into less time spent waiting, as it saves a call that definitely fails (when you have 2+ servers listed and none selected).

image

In the photo, we see that the login took us to /server/select (without the middleWare), but after failing a call (which surely would have failed, with no URL set)

-------------

These are the reasons that led me to take care of this case in the Middleware server.

doc-code-hub:
I am also rebasing this branch in advance for convenience...

src/middleware/serverMiddleware.ts
ferferga: @doc-code-hub Sorry I completely forgot this PR was still unmerged 😅.

I forgot about the redirect in the login page, so what you say makes sense. I hope you also tested the cases where no env var was provided and there are no servers, right? (I assume so as tests should also pick those)

I dwon't put more impediments to the middleware stuff as it's a huge mess right now and I already have it refactored in a local branch after the Pinia migration.

doc-code-hub: Hello @ferferga!
Don't worry, the commits don't run away and neither does my fork :)

Yes, I had personally tested the 3 possible cases:
- No Env Var provided
- Env Var with 1 Server
- Env Var with 2+ Server

I also modified the tests to consider all the cases listed (if my memory serves me right).

Later I go back to the branch and I make sure, as it's been a while now :)

I'll let you know, _but it should be a yes._

---

**Instead** with 

> "I dwon't put more impediments to the middleware stuff as it's a huge mess right now and I already have it refactored in a local branch after the Pinia migration"

 are you asking me to remove this check that I added on the middleware or were you talking in general?

ferferga: @doc-code-hub In general 😄

ferferga (approved):

ferferga:
@jellyfin-bot rebase

Issue #1697, opened 2022-01-15
[Player Enhancement] Keyboard and Click controls for better UX (especially for videos)


doc-code-hub:
### Proposal of Enhancement  _(with [PR on my fork](https://github.com/doc-code-hub/jellyfin-vue/pull/3/files) to preview changes)_

I don't know if I'm the only one
but I find it really **_vital for a good user experience_** to have a quick control on the player (especially for videos).

I think the following features are very very important:
- SPACE for Play / Pause
- Click to Play / Pause (only on the video and not on the rest of the UI)
and less important but still very comfortable
- press 'f' for the fullscreen toggle

We only think about when the doorbell rings and we have to quickly pause.

It can be frustrating to have to grab the mouse and correctly aim for the play / pause icon when time is running out.

Press the video or press space, are immediate solutions, applicable even (and above all) when you are in a hurry.

---------------------------

I noticed that a handleKeyPress was already provided, but the listener was instantiated only when the mutation TOGGLE_MINIMIZE = false is performed (and therefore not at the first opening of the player)

Since I did not want to open a PR before discussing it, I have prepared a [PR on my fork](https://github.com/doc-code-hub/jellyfin-vue/pull/3/files), just to show you the changes I propose (**which I have tested without encountering any issues**) in a clear and immediate way.

The changes I made are (regarding the Click and 'f' press) referring only to the MediaType = 'Video'.

I also changed the listener from 'keydown' to 'keyup' to **avoid** the **toggle loop** when holding a key

What do you think? 
Does anyone else find this lack a not insignificant problem?

If everything seems fine to you and there are no objections, I open the PR here

PR #1700, opened 2022-01-16
Keyboard and Click player controls for better UX

doc-code-hub:
Related to #1697:

I **changed** the **keyboard handler** in the player and improved the way listeners are handled.

Added support for the following shortcuts:

_For the full-player_
```
[f] = toggle fullscreen
[Spacebar, k] = play / pause (Spacebar only if there is no focus element, to support Tab)
[Arrowleft, j] =  prev
[Arrowright, l] = next
[m] = mute / unmute
```

_For the mini-player_
```
[f] = toggle back to full-player
```

_Mouse_
Also I added play / pause support by clicking (or touching) on the video player (ignoring the rest of the interface)

### DIFF
--- a/frontend/components/Players/PlayerManager.vue
+++ b/frontend/components/Players/PlayerManager.vue
@@ -252,13 +252,7 @@ export default Vue.extend({
       'getCurrentlyPlayingMediaType'
     ]),
     ...mapState('clientSettings', ['darkMode']),
-    ...mapState('playbackManager', [
-      'status',
-      'isMinimized',
-      'currentTime',
-      'isMuted',
-      'currentVolume'
-    ]),
+    ...mapState('playbackManager', ['status', 'isMinimized', 'currentTime']),
     isPlaying(): boolean {
       return this.status !== PlaybackStatus.Stopped;
     },
@@ -308,7 +302,6 @@ export default Vue.extend({
       'setNextTrack',
       'setPreviousTrack',
       'setLastItemIndex',
-      'setVolume',
       'toggleMute',
       'playPause',
       'pause',
--- a/frontend/components/Players/PlayerManager.vue
+++ b/frontend/components/Players/PlayerManager.vue
@@ -241,8 +241,7 @@ export default Vue.extend({
       keepOpen: false,
       playbackData: false,
       isUpNextVisible: false,
-      stretchVideo: true,
-      previousVolume: 0
+      stretchVideo: true
     };
   },
   computed: {
@@ -257,6 +256,7 @@ export default Vue.extend({
       'status',
       'isMinimized',
       'currentTime',
+      'isMuted',
       'currentVolume'
     ]),
     isPlaying(): boolean {
--- a/frontend/components/Players/PlayerManager.vue
+++ b/frontend/components/Players/PlayerManager.vue
@@ -391,6 +391,12 @@ export default Vue.extend({
           case 'ArrowLeft':
             this.skipBackward();
             break;
+          case 'f':
+            if (this.getCurrentlyPlayingMediaType === 'Video') {
+              this.toggleFullScreen();
+            }
+
+            break;
         }
       }
     },
--- a/frontend/components/Players/PlayerManager.vue
+++ b/frontend/components/Players/PlayerManager.vue
@@ -368,7 +368,7 @@ export default Vue.extend({
       this.setNextTrack();
     },
     handleKeyPress(e: KeyboardEvent): void {
-      if (!this.isMinimized && this.isPlaying) {
+      if (!this.isMinimized) {
         const focusEl = document.activeElement;
 
         let spaceEnabled = false;
@@ -409,7 +409,7 @@ export default Vue.extend({
             this.toggleMute();
             break;
         }
-      } else if (this.isMinimized && this.isPlaying) {
+      } else {
         switch (e.key) {
           case 'f':
             if (this.getCurrentlyPlayingMediaType === 'Video') {
--- a/frontend/components/Players/PlayerManager.vue
+++ b/frontend/components/Players/PlayerManager.vue
@@ -309,6 +309,7 @@ export default Vue.extend({
       'setPreviousTrack',
       'setLastItemIndex',
       'setVolume',
+      'toggleMute',
       'playPause',
       'pause',
       'unpause',
@@ -337,14 +338,6 @@ export default Vue.extend({
           this.fullScreenOverlayTimer = null;
         }
       }, this.getOsdTimeoutDuration());
-    },
-    toggleMute(): void {
-      if (this.currentVolume !== 0) {
-        this.previousVolume = this.currentVolume;
-        this.setVolume({ volume: 0 });
-      } else {
-        this.setVolume({ volume: this.previousVolume });
-      }
     },
     handleMouseMove(): void {
       if (
--- a/frontend/components/Players/PlayerManager.vue
+++ b/frontend/components/Players/PlayerManager.vue
@@ -228,7 +228,6 @@ import { mapActions, mapGetters, mapState } from 'vuex';
 import screenfull from 'screenfull';
 import imageHelper from '~/mixins/imageHelper';
 import timeUtils from '~/mixins/timeUtils';
-import { AppState } from '~/store';
 import { PlaybackStatus } from '~/store/playbackManager';
 
 export default Vue.extend({
@@ -242,7 +241,8 @@ export default Vue.extend({
       keepOpen: false,
       playbackData: false,
       isUpNextVisible: false,
-      stretchVideo: true
+      stretchVideo: true,
+      previousVolume: 0
     };
   },
   computed: {
@@ -253,7 +253,12 @@ export default Vue.extend({
       'getCurrentlyPlayingMediaType'
     ]),
     ...mapState('clientSettings', ['darkMode']),
-    ...mapState('playbackManager', ['status', 'isMinimized', 'currentTime']),
+    ...mapState('playbackManager', [
+      'status',
+      'isMinimized',
+      'currentTime',
+      'currentVolume'
+    ]),
     isPlaying(): boolean {
       return this.status !== PlaybackStatus.Stopped;
     },
@@ -283,17 +288,6 @@ export default Vue.extend({
     window.addEventListener('click', this.handleVideoClick);
 
     this.addMediaHandlers();
-
-    this.unsubscribe = this.$store.subscribe((mutation, state: AppState) => {
-      switch (mutation.type) {
-        case 'playbackManager/TOGGLE_MINIMIZE':
-          if (state.playbackManager.isMinimized === true) {
-            window.removeEventListener('keyup', this.handleKeyPress);
-          }
-
-          break;
-      }
-    });
   },
   beforeDestroy() {
     if (this.fullScreenOverlayTimer) {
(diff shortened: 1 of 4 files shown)


src/components/Players/PlayerManager.vue
```diff
@@ -341,6 +338,14 @@ export default Vue.extend({
...
       }, this.getOsdTimeoutDuration());
     },
+    toggleMute(): void {
+      if (this.currentVolume !== 0) {
+        this.previousVolume = this.currentVolume;
+        this.setVolume({ volume: 0 });
+      } else {
+        this.setVolume({ volume: this.previousVolume });
+      }
+    },
```
ThibaultNocchi, line 348: As we're duplicating logic from VolumeSlider, what I'd do is just set this as an action in the store. That way we can also store previousVolume there and factorize the logic in this component and VolumeSlider.

ThibaultNocchi:
Also, if you write something like `fix #XXX` in a PR, it'll close the associated issue when the PR is merged :)

doc-code-hub:
Forgive me but I don't understand exactly where [you are suggesting](https://github.com/jellyfin/jellyfin-vue/pull/1700#issuecomment-1013896015) me to use it and on what occasion.
At what stage of `npx cz` should I use it? :)

src/store/playbackManager.ts
```diff
@@ -73,6 +74,7 @@ export const defaultState = (): PlaybackManagerState => ({
   currentItemChapters: null,
   currentTime: null,
   lastProgressUpdate: 0,
+  previousValume: 100,
```
T-bond, line 77: Typo? ;)

ThibaultNocchi:
@doc-code-hub Yup, just write "fix #1697" in the PR message on GH and it should auto close when the PR is merged

src/store/playbackManager.ts
doc-code-hub: Yes :) on it

ferferga:
@doc-code-hub "Related" isn't recognised, you must use "fix" or "fixes" specifically. However, I already liked the issue manually, so no worries for this PR, but take that into account for the others :).

src/components/Players/PlayerManager.vue
```diff
@@ -378,18 +376,65 @@ export default Vue.extend({
...
+          case 'f':
+            if (this.getCurrentlyPlayingMediaType === 'Video') {
+              this.toggleMinimized();
+            }
+
+            break;
+        }
+      }
+    },
+    handleVideoClick(e: any) {
```
ferferga, line 430:
```suggestion
    handleVideoClick(e: HTMLMediaElement) {
```

src/store/playbackManager.ts
ferferga, line 77: To simpligy all of this, I would just save a boolean ``isMuted`` in the store. In the same store action you can check if the new volume === 0 and commit the mutation that toggles the ``isMuted`` variable.

In the components, just map the state of that var and bind it to the muted attribute of the video and audio element.
Thoughts @ThibaultNocchi ?

src/components/Players/PlayerManager.vue
```diff
@@ -242,7 +241,8 @@ export default Vue.extend({
       keepOpen: false,
       playbackData: false,
       isUpNextVisible: false,
-      stretchVideo: true
+      stretchVideo: true,
+      previousVolume: 0
```
ferferga, line 245: This can be removed I think as you have now the store?

src/components/Players/PlayerManager.vue
```diff
@@ -279,28 +284,19 @@ export default Vue.extend({
   },
   mounted() {
     document.addEventListener('mousemove', this.handleMouseMove);
+    window.addEventListener('keyup', this.handleKeyPress);
+    window.addEventListener('click', this.handleVideoClick);
```
ferferga, line 288: Why not ``@ keyup`` or ``@ click`` in the template?

ferferga (changes requested):

src/components/Players/PlayerManager.vue
doc-code-hub: Click bc overlay on the video (and I did not find the right elements after several attempts).

For the keyup I had tried to use that of the Dialog however I encountered some strange behaviors.

After the attempts I made this turned out to be the mix that provided the best experience.

I can try to go through it a second time and double check afterwards.

src/store/playbackManager.ts
ThibaultNocchi: Yup I agree

doc-code-hub:
Sorry guys (@ferferga @ThibaultNocchi ), yesterday wasn't clearly my day.
I left some typos lying around and useless stuff lying around... My fault

I only realized now that you are both referring to the PR message while I meant the commit message :)

Today I do a nice round of corrections on both of my PRs.

Thank you for the support :)

src/components/Players/PlayerManager.vue
```diff
@@ -312,6 +308,8 @@ export default Vue.extend({
       'setNextTrack',
       'setPreviousTrack',
       'setLastItemIndex',
+      'setVolume',
```
ThibaultNocchi, line 311: you can remove this mapping too :)

src/components/Players/PlayerManager.vue
```diff
@@ -253,7 +252,13 @@ export default Vue.extend({
       'getCurrentlyPlayingMediaType'
     ]),
     ...mapState('clientSettings', ['darkMode']),
-    ...mapState('playbackManager', ['status', 'isMinimized', 'currentTime']),
+    ...mapState('playbackManager', [
+      'status',
+      'isMinimized',
+      'currentTime',
+      'isMuted',
+      'currentVolume'
```
ThibaultNocchi, line 260: you can remove both those mappings it seems :)

ThibaultNocchi:
@jellyfin-bot rebase

TKFRvisionOfficial:
any news on this stuff?

ferferga:
@doc-code-hub Can you rebase so we can resume the reviewing with the new folder structure and get this in?

Git should fix the merge conflicts automatically.

doc-code-hub:
@ferferga, done.

Successfully rebased without conflict 👍

frontend/components/Players/PlayerManager.vue
```diff
@@ -378,20 +369,67 @@ export default Vue.extend({
...
+
             break;
         }
       }
     },
+    handleVideoClick(e: MouseEvent) {
+      const target = e.target as HTMLElement;
+
+      if (
+        target?.classList?.contains('player-overlay') &&
```
ferferga, line 427:
```suggestion
      const target = e.target;

      if (
        target &&
        target.classList.contains('player-overlay') &&
```

This way you have proper checks at runtime

frontend/components/Players/PlayerManager.vue
```diff
@@ -279,28 +278,19 @@ export default Vue.extend({
...
-        case 'playbackManager/TOGGLE_MINIMIZE':
-          if (state.playbackManager.isMinimized === true) {
-            window.removeEventListener('keydown', this.handleKeyPress);
-          } else if (state.playbackManager.isMinimized === false) {
-            window.addEventListener('keydown', this.handleKeyPress);
-          }
-
-          break;
-      }
-    });
```
ferferga, line 296: How you handle this now?

This means that clicking in the button directly (without a keypress) doesn't perform the action, which may be conflicting for music for instance

frontend/components/Players/PlayerManager.vue
ferferga: Btw, no luck at binding this to the ``playPause`` method [here](https://github.com/jellyfin/jellyfin-vue/blob/824c7008071e2e0335569883ae57cc9775ffb730/frontend/components/Players/PlayerManager.vue#L31)?

ferferga (changes requested):

frontend/components/Players/PlayerManager.vue
doc-code-hub: I'm not on the project right now, but this is the mutation subscription, right?

Shouldn't it be done in every case where mutation occurs? (therefore also pressing with a "physical" button instead of keyboard)

ferferga: The unsubscription doesn't happen on every mutation, just when the component is unmounted. 

With the subscription you make sure that any mutation to the store that changes the mutation store is handled. Without it, you can whatever you want that nothing will happen.

frontend/components/Players/PlayerManager.vue
doc-code-hub: Hi @ferferga, sorry if I answer you late but I was having trouble installing dependencies from the structure change ... (a TLS problem with npm, which I solved).

Ah anyway, I tried (and had tried at the time) to directly insert the playPause on the click of the 'v-overlay' as you suggest.
The problem is that it is fired all the time (even when you press anything else on the screen, with nasty side effects).

Using the click handler and checking the 'player-overlay' class, I am able to obtain only the significant clicks (i.e. those in the purple area of the photo, avoiding any possible side effect (like double action @ once)), 
_which in my opinion results in the best experience._

image

frontend/components/Players/PlayerManager.vue
doc-code-hub: Sorry, I didn't realize it was code I removed.
----
If I removed it, it is because the listener is used for the player mini as well.

If you notice, the listener code, performs the actions based on whether the player is mini or normal, so the listener can always remain active, instead of being inserted and removed (this mainly because "f" for the player mini, puts it back to normal player, as the description of the PR indicates).

ferferga: @doc-code-hub But how other components dispatching the mutation works now then? You must put a replacement in place then.

frontend/components/Players/PlayerManager.vue
ferferga: Okay, add the type checks to that section and all fine to me.

doc-code-hub: Oh sure, forgotten.

doc-code-hub: @ferferga, 
I made the change, however I have to keep "as HTMLEvent" because MouseEvent declares target as an EventTarget which does not hold the properties of an HTMLElement, however it is an HTMLElement that I receive (and that i need).

The alternative, but less elegant, would be to type a: any target, but it didn't seem the best...

Look good to you?

frontend/components/Players/PlayerManager.vue
doc-code-hub: I see no problems on the horizon, considering that a check on isMinimized takes place in the handler (which is a property mapped directly from the "playbackManager" store).

So if another component were to change the isMinimized state, this would be automatically updated in the component as well, making sure that the handler always has the current and reliable value of "isMinimized".

It shouldn't matter who throws an "isMinimized" mutation since it is a property mapped to the component with mapState.

Do you agree @ferferga?

ferferga: Fine, missed the mapped state, sorry!

ferferga (approved):
LGTM.

Can you squash https://github.com/jellyfin/jellyfin-vue/pull/1700/commits/3ed513b3e39e15c39dbfbe01b6544bde8299a03e, https://github.com/jellyfin/jellyfin-vue/pull/1700/commits/c394c92f0b7aaa3b54e8aa2d3f40a9ca52fef393 and https://github.com/jellyfin/jellyfin-vue/pull/1700/commits/5b7552b7155d316da5ac5046142d27d3c9e54a11 together please?

Too many commits for just a minor typing thing 😅.

doc-code-hub:
Okay @ferferga, I squashed those 3 commits (and rebased to master in the meanwhile) 👍

PR #1707, opened 2022-01-23
add stale issue check workflow

h1dden-da3m0n:
### Description

This PR aims to add a GitHub Action workflow to monitor, label and close stale issues.
It uses the Jellyfin-Bot token to comment and close stale Issues.

> NOTE: It explicitly does not touch PRs

### Changes

+ add a basic GitHub action workflow to monitor, label, and close stale issues within this repository

### Issues

+ successor to the discontinued Stale-Bot config

### DIFF
--- a/.github/workflows/repo-stale.yaml
+++ b/.github/workflows/repo-stale.yaml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/stale@v4.1.0
         with:
-          repo-token: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


ThibaultNocchi:
I think for now the consensus with @ferferga was that we removed it not too long ago. We did that as we are not overflowing with issues so we can triage them as they come.

h1dden-da3m0n:
As we can always re-open a PR similar to this and the current consensus is to not use it here I will close the PR for now :+1:

PR #1735, opened 2022-03-06
fix episode counter updating

### LINKED ISSUE #1717: Remaining episodes counter doesn't update
### Description of the bug

When in an overview of a library which is marked as TV-Shows there's the option to mark entire seasons as watched.
However doing so doesn't update the view but rather only calls the route.

Manually refreshing the page is required to display the changes made.

### Steps to reproduce

In a TV-Shows marked library overview:
- find a show with at least 1 unseen episode
- click on the tick button in the overview
- nothing happens until you refresh the page manually or navigate away and back to it again

### Expected behavior

Pill with the remaining episodes is updated as soon as the call to the server is successful.

### Logs

_No response_

### Screenshots

_No response_

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.7.7

### Additional context

_No response_

HazilMohamed:
**Changes**
- Add emitter to update episode counter on clicking  show as marked as watched

**Issues**
- Fixes #1717 

_First time working on vue_

### DIFF
--- a/frontend/components/Buttons/MarkPlayedButton.vue
+++ b/frontend/components/Buttons/MarkPlayedButton.vue
@@ -60,6 +60,8 @@ export default Vue.extend({
             itemId: this.item.Id || ''
           });
         }
+
+        this.$emit('change', this.isPlayed);
       } catch (error) {
         this.pushSnackbarMessage({
           message: this.$t('unableToTogglePlayed'),
--- a/frontend/components/Item/Card/Card.vue
+++ b/frontend/components/Item/Card/Card.vue
@@ -24,9 +24,13 @@
             color="white"
             size="24"
           />
-          <watched-indicator v-if="item.UserData && item.UserData.Played" />
+          <watched-indicator v-if="watchedIndicator" />
           <v-chip
-            v-if="item.UserData && item.UserData.UnplayedItemCount"
+            v-if="
+              !watchedIndicator &&
+              item.UserData &&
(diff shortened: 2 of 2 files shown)


ferferga:
@HazilMohamed As Thibault mentions, this is a design we've been trying to avoid since the beginning, as it's prone to introduce a lot of complexities and make the maintenance more difficult (since tracking events is definitely more difficult as they go from the bottom to the top, instead of the other way around). Tracking reactivity changes, which Vue devtools extension, is far easier, as you know that they propagate from the top to the bottom and is shared by all the components (something that you can't really assure well with events).

As Tiw said, I'm working in migrating to Pinia and, along the way, items that are fectched through the API would be directly reactive (without any additional work) to whatever is informed through the socket. This is implemented in my local branch but still untested, so I can't promise anything, but so far looks promising.

I think we can keep your PR around and deprecate it as soon as mine with the reactivity directly done in the store is published (that way I don't forget to check that what you're fixing it's fixed in my work :P).
If you don't mind, I can ping you when that PR is live here, so you can check that everything is fine and provide your own patches if you want (as probably a lot of other stuff won't still be reactive and having more eyes while reviewing is always a good idea :P)

Thank you very much for your effort and desire to contibute!

HazilMohamed:
@ferferga no problem. It will be better and efficient if we can update the store directly as you mentioned. Keep up the good work!

PR #1736, opened 2022-03-07
docs: add cf page back

### LINKED ISSUE #1648: Is Cloudflare Pages still active
I always used https://jellyfin-vue.pages.dev but in the last commit, I can not find the cloudflare pages build status? Did you disable cloudflare pages? If yes, why? So I did not have to deploy it on my own or do you now manually somewhere else?

ThibaultNocchi:
fixes #1648

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@
 
 This is an experimental web client for Jellyfin based on Vue.js. We welcome all contributions and pull requests! If you have a larger feature in mind please open an issue so we can discuss the implementation before you start.
 
+<p align="center">
+    <a href="https://jf-vue.pages.dev/">View demo</a>
+</p>
+
 # Build Process
 
 ## Dependencies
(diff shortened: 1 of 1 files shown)


ferferga:
Now that you're on it, would you mind removing the Husky line and replacing it with a text that just refers to the conventional commit guidelines? (As we don't have Commitzen or Husky anymore)

ThibaultNocchi:
Done :)

README.md
```diff
@@ -72,17 +76,17 @@ This is an experimental web client for Jellyfin based on Vue.js. We welcome all
...
 
-   ```bash
-   npm run build
-   ```
+```bash
+npm run build
+```
 
-   Build output will be available under the `src/dist` folder.
+Build output will be available under the `src/dist` folder.
```
ferferga, line 89: Is the removal of the tabulation intended?

ThibaultNocchi: yes that's the markdown linter that did it. And it seems to make sense to me

ThibaultNocchi: I also did it in a specific commit :)

ferferga: @ThibaultNocchi How we can keep this layout then?

**master**
image

**Your branch**
image

Imo it makes sense to keep the "Build for production" part inside "Getting Started" as it's the bare minimum an user needs to build their own client version. The rest of the sections cover more "advanced" use cases and are good as they're now, but imo, these two initial sections should be kept together.

Issue #962, opened 2021-04-05
Opening the queue should display currently playing song


neopc10:
Currently when opening the queue view, the first ~8 tracks are displayed. If playing a relatively long queue (x>20 tracks) it'll take a bit of scrolling to show what is currently playing.

Preferred behavior would be that if the currently playing track is item number 40, the queue view skips down to the 40th item upon opening it.

PR #1746, opened 2022-04-11
refactor(pinia): migrate all stores to pinia and other improvements

### LINKED ISSUE #971: Arrow keys controlling music playback
Currently the left and right arrow keys will go to the previous and next song queued on the fullscreen music player or anywhere the bottom controls are visible.

Also currently the left and right arrow keys (in addition to pg up+down) will go to the previous and next swiper pages on the home screen.

They shouldn't do both simultaneously -- I think the best would be to enable left/right arrow keys for music controls only in the fullscreen music player.

### LINKED ISSUE #980: Implement the new Items store in all the client

ferferga:
* Migrated all the stores to Pinia: fully typed, with the ability to add comments to actions and really a breeze to work with.
    * Also seems much more performant than Vuex, to the point where throttling updates in ``ShakaPlayer`` and ``HlsPlayer`` was less performant in Edge 100.0.1185.36 and Firefox 99 than directly committing the updates. This still has potential to be improved in the future, as Pinia allows us to do two-way data-binding computed properties, which should remove the need for action subscriptions in players altogether

* Items returned in Items enpoint by the Axios instance are automatically added to the item store and, thus, are reactive everywhere in the client now. *Fixes #980?. See point 1 in 'ToDo' for more details*

    * Also removed nuxtjs/axios module, so we're using upstream axios directly, and yet another step in being a bit less framework-dependent. 

* Remove nuxtjs/auth with our custom implementation, to simplify the logic of our auth process and make it more independent from nuxt. It's also much clearer for new developers, as before we had the store, the module and the plugin. Now it's just a matter of the store and the middleware.
    * The user is logged out automatically after a session is revoked in another device and a request is done. *See point 2 in 'ToDo' for more details*
    * Rearchitected the stores to allow multiple logged-in users in multiple servers and seamless switch between those.

* Add ``vue-tsc`` and ``npm run typecheck`` command, which does a typecheck to the SFC syntax, something that we didn't have before. This doesn't work very well with Vue 2, as vue-tsc is a Volar's extension (for Vue 3), but I'm going to keep an eye on this during the transition period.

* Fix scroll not allowed while playing music, #971, #962 and media session implementation

* Added a buffering state to display and track animations while content is loading.

## To Do

1. This in practice isn't working: everywhere you check in the client, those objects returned by the ``$api`` plugin have the ``__ob__`` property automatically added, as all the Vue observables. However, when the socket dispatches an update, the item is updated in the store, but Vue components relying on these doesn't update for some reason (even though they're watching it, as can be seen in the ``__ob__.deps`` object). This still needs some investigation from my side.

2. As soon as a request is made after the session has been revoked, not immediately. Also, **depending in the situation, the user might be greeted with an error message if navigation is attempted**. This is [impossible to remove and a Nuxt's limitation](https://stackoverflow.com/questions/59292573/how-do-i-disable-nuxt-default-error-redirection)

3. Adding all the relevant tests for this would be too much for this already single PR (where changes couldn't be isolated). Those are going to be readded in the future, but Im gearfing towards switching to Cypress for more accurate E2E testing, as I think unit testing doesn't cover well enough the scope of this project.


frontend/plugins/nuxt/appInit.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Add default servers from environment variables
+   */
+  const defaultServers = parseServerListString(process.env.server_url_list);
+  const missingServers = defaultServers.filter((serverUrl) => {
+    const server = auth.servers.find(
+      (lsServer: ServerInfo) => lsServer.PublicAddress === serverUrl
+    );
+
+    return !isNil(server);
```
doc-code-hub, line 32:
```suggestion
    return isNil(server);
```
### Related to #1698

If you are searching the Missing servers, the check should be on isNil intead of it's opposite.

If the result of the find() is undefined then the server is missing.

Ps. i have tryied out in my local env before telling you this

doc-code-hub:
@ferferga,  2 more things i will dive into later but in the meanwhile you can take a look if you are on it.

### Collections page
In the list there is also a playlist, but it's not a BoxSet...

### Related to #1680 
The BoxSet details always show "This collection is empty", but they are not empty. Like the children are lost in the middle

I saw you updated the way collections are stored and retrieved ...
Later I will check better to find out the real problem! ( Tell me if you figure it out first :) )

frontend/store/items.ts
```diff
@@ -1,193 +1,174 @@
...
+        await this.$nuxt.$api.items.getItems({
+          userId: auth.currentUserId,
+          parentId,
+          fields: allFields
+        })
+      ).data;
 
-        children[item.Type].push(item);
+      if (childItems.Items) {
+        return this.add(childItems.Items) as BaseItemDto[];
```
doc-code-hub, line 126: You created the method above but didn't use it later:
```suggestion
        const parent = this.getItemById(parentId);
      if (childItems.Items && parent) {
        return this.addCollection(parent, childItems.Items) as BaseItemDto[];
      }
```

frontend/components/Item/CollectionTabs.vue
```diff
@@ -41,23 +42,22 @@ export default Vue.extend({
     };
   },
   computed: {
-    ...mapGetters('items', ['getBoxsetChildren']),
+    ...mapStores(itemsStore),
     children(): BaseItemDto[] {
-      return this.getBoxsetChildren(this.item.Id);
+      return this.items.getChildrenOfParent(this.item.Id) || [];
```
doc-code-hub, line 47: I'm afraid this is not enough, as the component tabs elements based on type.

In order not to have regressions, it would be necessary to reintroduce the logic that I had inserted at the time.

```suggestion
     children(): ItemsByType[] {
      const list = this.items.getChildrenOfParent(this.item.Id);

      if (list) {
        const c: { [key: string]: BaseItemDto[] } = {};

        list.forEach((item: BaseItemDto) => {
          if (item.Type && typeof item.Type === 'string') {
            if (!c[item.Type]) c[item.Type] = [];

            c[item.Type].push(item);
          }
        });

        const perTypeList: ItemsByType[] = Object.keys(c).map((type) => {
          return { Type: type, Items: c[type] };
        });

        return perTypeList;
      } else return [];
    }
```
It needs this, declared somewhere

```typescript
interface ItemsByType {
  Type: string;
  Items: BaseItemDto[];
}
```

doc-code-hub (changes requested):
#1680 
After some local tests, i suggest some modifications in order to avoid regressions in relation to this feature.

frontend/plugins/nuxt/appInit.ts
ferferga: Ooops, good catch!

frontend/components/Item/CollectionTabs.vue
ferferga: I fixed this with a [simpler logic](https://github.com/jellyfin/jellyfin-vue/pull/1746/commits/ebd85f3721f15137582a37066feb950005ce92d2#diff-4edbc1f6f9c1a1e8f29ee928046706a4d29a182a30b7309f00dbb3a3e5bcdbd6R46-R61)

The tabs still need proper translation support though

frontend/store/items.ts
ferferga: Good catch!

frontend/components/Item/CollectionTabs.vue
ferferga: I fixed this with a [much more easier logic](https://github.com/jellyfin/jellyfin-vue/blob/pinia/frontend/components/Item/CollectionTabs.vue#L43)

The tabs still need proper translation support though

doc-code-hub: Well done!

I absolutely have to see what lodash offers, because I see rather comfortable methods around here! 😄

doc-code-hub:
Hi @ferferga,, I'm trying to figure out how to solve the regression that I encounter unlike the master, in streaming video.

Do you have any ideas?

I am trying to clarify my ideas, but honestly it is not perfectly clear to me, from the code, where everything should start from, and what is the entity that makes all the calls to the individual .ts chunks or to the stream.mp4.

Something is preventing the stream from starting properly as it did before...

Could you clarify what the heart of streaming is?
I see the playerManager status set but I don't understand exactly what responds to it and where.

ferferga:
@jellyfin-bot rebase

ferferga:
@jellyfin-bot rebase

frontend/store/auth.ts
```diff
@@ -0,0 +1,297 @@
...
+        ).data;
+
+        this.rememberMe = rememberMe;
+
+        if (authenticateResponse.User?.Id && authenticateResponse.AccessToken) {
+          Vue.set(
+            this.accessTokens,
+            authenticateResponse.User.Id,
+            authenticateResponse.AccessToken
+          );
```
ThibaultNocchi, line 116: What does it do?

ThibaultNocchi (commented):
A few comments for my personal knowledge ^^

frontend/store/auth.ts
ferferga: [This is a caveat in Vue 2 reactivity system](https://v2.vuejs.org/v2/guide/reactivity.html#For-Objects)

Basically, if you have an object like this:

```js
mainKey: {
    key1: 'string',
    key2: 3,
    key3: ['hello', 1]
}
```

Vue won't react to changes if you directly add the key:

```js
mainKey.key4 = 6;

## Which transforms the object to:
mainKey: {
    key1: 'string',
    key2: 3,
    key3: ['hello', 1],
    key4: 6
}
``` 
Or delete it:

```js
delete mainKey.key3;

## Which transforms the object to:
mainKey: {
    key1: 'string',
    key2: 3
}
```

This is why ``Vue.set`` and ``Vue.delete`` helper functions exist, so Vue can correctly proxy the "transformed" objects as observables.

Vue 3 with composition API has some improvements in this regards iirc, but I can't explain them in detail right now.

frontend/components/Buttons/PlayButton.vue
```diff
@@ -76,22 +77,21 @@ export default Vue.extend({
     };
   },
   computed: {
-    ...mapState('playbackManager', ['status'])
+    ...mapStores(playbackManagerStore)
   },
   watch: {
-    status(): void {
-      if (this.status === PlaybackStatus.Playing) {
+    'playbackManager.status'(): void {
```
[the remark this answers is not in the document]
ferferga: This kind of syntax allows us to watch for changes for a specific property in an object. So, if we have:

```js
mainKey: {
    key1: 'string',
    key2: 3,
    key3: ['hello', 1]
}
```

If we set the watcher as:

```js
mainKey(): void {
    ...
}
```

The handler will run if any of the **values** (not keys, see [here](https://github.com/jellyfin/jellyfin-vue/pull/1746#discussion_r862503703)) changes. If we declare the watcher as:

```js
'mainKey.key2'(): void {
    ...
}
```

We will be watching **value** changes just in ``key2``.

Keys can also be accessed through string, both these statements are equal:

```js
mainKey.key1
mainKey['key1']
```

This is why the handler is between quotes, because that syntax is also valid to access values in an object. Doing it without quotes **makes VSCode turn everything red (although, to be honest, I didn't bother checking if compilation was still successful or not, because in that case it's a problem we can fix by revisiting our linting config)**.

I can recheck this, but perhaps it's not a linter issue and **dots are not supported in a function declaration** in the ECMAScript specification

frontend/components/Buttons/PlaybackSettingsButton.vue
```diff
@@ -119,14 +122,7 @@ export default Vue.extend({
...
   computed: {
-    ...mapGetters('playbackManager', [
-      'getCurrentItemAudioTrack',
-      'getCurrentItemSubtitleTracks'
-    ]),
-    ...mapState('playbackManager', [
-      'currentAudioStreamIndex',
-      'currentSubtitleStreamIndex'
-    ])
+    ...mapStores(playbackManagerStore)
```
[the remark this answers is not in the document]
ferferga: The older ``mapX`` functions are also present in Pinia, but honestly I think it's better to make the store instance available with all the properties in the Component's instance and call it a day. No need for us to bother declaring new properties or actions, just type ``this.store`` and have everything autocompleted is much better!

frontend/components/Layout/Backdrop.vue
```diff
@@ -13,16 +13,14 @@
...
   computed: {
-    ...mapState('page', ['backdrop']),
-    blurhash(): string {
-      return this.backdrop.blurhash;
-    },
-    opacity(): number {
-      return this.backdrop.opacity;
+    ...mapStores(pageStore),
+    blurhash(): string | null | undefined {
+      return this.page.backdrop.blurhash;
```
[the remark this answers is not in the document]
ferferga: I'll check this!

frontend/components/Buttons/PlayButton.vue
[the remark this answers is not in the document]
ThibaultNocchi: Thanks! Seems you're right https://v2.vuejs.org/v2/api/#watch

Issue #975, opened 2021-04-07
No option to remove individual music from the queue.


Bitwolfies:
While I'm sure its known, I didn't see any issues for it, and I feel its a pretty important thing to have.

akkisagiraju:
I'm working on this. Will create a PR soon.

Bitwolfies:
still needs an addition

PR #1753, opened 2022-05-03
feat(Queue): items can now be deleted from the queue

LPkkjHD:
As there hasn't been any activity related to https://github.com/jellyfin/jellyfin-vue/issues/975 for almost a year I'm allowing myself to add a PR to this feature. Closes https://github.com/jellyfin/jellyfin-vue/issues/975.

I am claiming the space from the `itemMenu` which really has no use in the queue panel in order to make space for the button which removes an item from the queue.

image

Looks like the screenshot above. If there is a specific reason why the `itemMenu` button was there we could also work out another solution for the placement of the `remove` button.

### DIFF
--- a/frontend/components/Players/DraggableQueue.vue
+++ b/frontend/components/Players/DraggableQueue.vue
@@ -41,7 +41,9 @@
             <like-button :item="item" />
           </v-list-item-action>
           <v-list-item-action v-if="!isPlaying(index)" class="mr-2">
-            <item-menu :item="item" />
+            <v-btn icon @click="playbackManager.removeFromQueue(item.Id)">
+              <v-icon>mdi-playlist-minus</v-icon>
+            </v-btn>
           </v-list-item-action>
         </v-list-item>
       </v-hover>
--- a/frontend/store/playbackManager.ts
+++ b/frontend/store/playbackManager.ts
@@ -98,6 +98,11 @@ export const playbackManagerStore = defineStore('playbackManager', {
       queue.push(...translatedItem);
       this.queue = queue;
     },
+    removeFromQueue(itemId: string) {
+      if (this.queue.includes(itemId)) {
+        this.queue.splice(this.queue.indexOf(itemId), 1);
+      }
+    },
     clearQueue(): void {
       this.queue = [];
     },
--- a/frontend/store/playbackManager.ts
+++ b/frontend/store/playbackManager.ts
@@ -92,11 +92,9 @@ export const playbackManagerStore = defineStore('playbackManager', {
   },
   actions: {
     async addToQueue(item: BaseItemDto) {
-      const queue = Array.from(this.queue);
       const translatedItem = await this.translateItemsForPlayback(item);
 
-      queue.push(...translatedItem);
-      this.queue = queue;
+      this.queue.push(...translatedItem);
     },
     removeFromQueue(itemId: string) {
       if (this.queue.includes(itemId)) {
@@ -181,12 +179,10 @@ export const playbackManagerStore = defineStore('playbackManager', {
      * @param item
      */
     async playNext(item: BaseItemDto): Promise<void> {
-      const queue = Array.from(this.queue);
       const translatedItem = await this.translateItemsForPlayback(item);
 
       if (this.currentItemIndex !== null) {
-        queue.splice(this.currentItemIndex + 1, 0, ...translatedItem);
-        this.queue = queue;
+        this.queue.splice(this.currentItemIndex + 1, 0, ...translatedItem);
       }
     },
     pause(): void {


frontend/components/Players/DraggableQueue.vue
```diff
@@ -41,7 +41,9 @@
             <like-button :item="item" />
           </v-list-item-action>
           <v-list-item-action v-if="!isPlaying(index)" class="mr-2">
-            <item-menu :item="item" />
+            <v-btn icon @click="removeFromQueue(item)">
+              <v-icon>mdi-delete-outline</v-icon>
```
ferferga, line 45: I mention this already the review's main message, but I think ``mdi-playlist-minus`` fits better.

ferferga:
Don't worry about checks not passing. The lint checks failing is an upstream issue with npm and workspaces that is already reported, Tauri is also known. 

As for the branch not up to date with master message, rebases with force push are accepted here, so you're free to keep your branch up to date with master this way.

frontend/components/Players/DraggableQueue.vue
```diff
@@ -84,6 +86,13 @@ export default Vue.extend({
     }
   },
   methods: {
+    removeFromQueue(item: BaseItemDto): void {
+      const newQueue = this.queue;
+      const index = newQueue.indexOf(item);
+
+      newQueue.splice(index, 1);
+      this.queue = newQueue;
+    },
```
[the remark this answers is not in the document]
LPkkjHD: This seems very reasonable and I totally missed this. It is addressed by https://github.com/jellyfin/jellyfin-vue/pull/1753/commits/c0ca5780b2e79e655ba00f6c00d84ea4ec7ee165

frontend/components/Players/DraggableQueue.vue
LPkkjHD: I do agree with the choice, the trash can seems very out of place, however, I didn't really find a good alternative before. The icon is addressed in https://github.com/jellyfin/jellyfin-vue/pull/1753/commits/16249a688a65bad8bf7d6afa60609867d6114c06.

LPkkjHD:
@ferferga this idea looks very promising to me. I did address your change requests, however, as I'm currently working on getting the DLNA settings page implemented I might only be able to try my hand at your idea later this week.

Maybe this goes out of scope for this PR too and we should first merge this and create another one for that feature once someone got some free time at their hands.

Should definitely move this draft/design over to the discussions to have it saved.

frontend/store/playbackManager.ts
```diff
@@ -98,6 +98,13 @@ export const playbackManagerStore = defineStore('playbackManager', {
       queue.push(...translatedItem);
       this.queue = queue;
     },
+    async removeFromQueue(item: BaseItemDto) {
+      const queue = Array.from(this.queue);
+      const translatedItem = await this.translateItemsForPlayback(item);
+
+      queue.splice(queue.indexOf(translatedItem[0]), 1);
```
ferferga, line 105: Why fetching items from server and building up an entire queue is needed for just removing an item from the queue? Ooooops! 😅

frontend/components/Players/DraggableQueue.vue
```diff
@@ -41,7 +41,9 @@
             <like-button :item="item" />
           </v-list-item-action>
           <v-list-item-action v-if="!isPlaying(index)" class="mr-2">
-            <item-menu :item="item" />
+            <v-btn icon @click="removeFromQueue(item)">
```
ferferga, line 44:
```suggestion
            <v-btn icon @click="playbackManager.removeFromQueue(item)">
```

LPkkjHD:
> I think we can have a menu with this layout there:
> 
> ```
> mdi-arrow-collapse-up - Move up and play next
> mdi-playlist-minus - Remove from queue
> mdi-arrow-collapse-down - Move to the end
> <v-divider />
> mdi-dots-horizontal - More - mdi-chevron-right
> ```

I don't think that the `More` menu point here is actually necessary nor that it is any good. By the time the user reaches this point they will have already tapped onto: `Queue`->`More` adding another layer on top of that would not be good for UX. Also I reckon there is not really anything which we might want to show there.

As we already have the full `More` menu in the player UI - where other desirable options such as editing metadata can be done - I think there is no need to implement the last two options leaving us with something like:

```
> mdi-arrow-collapse-up - Move up and play next
> mdi-playlist-minus - Remove from queue
> mdi-arrow-collapse-down - Move to the end
```

frontend/store/playbackManager.ts
```diff
@@ -98,6 +98,13 @@ export const playbackManagerStore = defineStore('playbackManager', {
...
+      this.queue = queue;
+    },
```
ferferga, line 107: For making this function more modular, I think it would be a nice touch to accept the item.Id directly. If we have the BaseItemDto in the calling code, it's as easy as doing ``removeFromQueue(item.Id)``. But if we have the id as an string, we need to fetch the item to call the function, so TypeScript doesn't complain.

```suggestion
    removeFromQueue(itemId: string) {
      if (this.queue.includes(itemId)) {
          const queue = Array.from(this.queue);

          queue.splice(queue.indexOf(item, 1);
          this.queue = queue;
      }
    },
```

Remember to update the call in DraggableQueue accordingly

frontend/components/Players/DraggableQueue.vue
ferferga: With what I suggest above:

```suggestion
            <v-btn icon @click="playbackManager.removeFromQueue(item.Id)">
```

frontend/components/Players/DraggableQueue.vue
```diff
@@ -84,6 +86,9 @@ export default Vue.extend({
     }
   },
   methods: {
+    async removeFromQueue(item: BaseItemDto): Promise<void> {
+      await this.playbackManager.removeFromQueue(item);
+    },
```
ferferga, line 91: Unnecessary if you do it like I did it above
```suggestion
```

ferferga (changes requested):

ferferga:
> I don't think that the `More` menu point here is actually necessary nor that it is any good. By the time the user reaches this point they will have already tapped onto: `Queue`->`More` adding another layer on top of that would not be good for UX. Also I reckon there is not really anything which we might want to show there.

Right now doesn't seem very useful, but that's because half of the options that the menu should have are missing. Check jf-web's menu:

image

> As we already have the full `More` menu in the player UI - where other desirable options such as editing metadata can be done - I think there is no need to implement the last two options leaving us with something like:

What if I want to edit the metadata of the 25th item in queue while I'm playing the 3rd?

frontend/store/playbackManager.ts
LPkkjHD: Yes this is a rather expensive option I used in order to get the item id. With your suggested changes this is also not relevant anymore as we take the itemid directly from the item

ferferga:
LGTM now, but can you squash all the commits into one please? There is plenty of commits for such a simple change 😅.

LPkkjHD:
> Right now doesn't seem very useful, but that's because half of the options that the menu should have are missing.
Seeing the difference with jf-web I see the point in having more options here. However, I still think that options which are even remotely often used shouldn't be hidden behind 3 UI clicks.

I'd propose (based on jf-web's options) to have the following structure:
```
Move up and play next
Remove from queue
Move to the end
>> divider
Add to playlist
Download
>> divider
View Album
View Album Artist
>> divider
More -> Media Info
     -> Edit Images
     -> Edit Metadata
     -> Refresh Metadata
     -> Copy StreamURL
```
Like this the items shouldn't be too many to be overbearing. Comparing things to Spotify they seem to have a similar amount of options in their More menu. 
image

As for the Instant Mix option I'm still unsure whether we should actually include it as it seems to be more of a "initiate Playback" action than something you would want to do with a song in their current queue (it clears the queue in favor or the created mix).

How does this sound to you @ferferga?

ThibaultNocchi:
I think Instant Mix should be available. I can just randomly play an album/my entire collection, find a song I'm in the mood for and start a mix from there.

Also for instance, Deezer has the same song options in the list view than in the queue view.

Thanks for your help! :)

ferferga:
@LPkkjHD Seems reasonable. My vote for the "More" submenu is basically for this (in case I wasn't that clear before):

* Separate options that are handy to have, but they have different contexts: the context of **queue management** and the context of **item management**

* Don't make the menu so overwhelming in a mobile version of the queue (after all the queue management interface is ported to a responsive design, of course). Spotify, for instance, doesn't have a menu in the queue in their mobile apps, but I don't like giving less choice to users just based on what device they're using.

All in all, we can discuss this better in the relevant PR that implements the feature or in a discussion thread, as all of this will probably get lost in the clutter, so let's not take it further than a simple brainstorming.

For now, I'm going to merge this :)

ferferga (approved):

Issue #1754, opened 2022-05-03
ShakaPlayer doesn't initiate playback


LPkkjHD:
### Description of the bug

When opening any music library and pressing play on an album the player will be created, however, there will be an infinite loading loop instead of playback. Pressing the `next` or `previous` media buttons will skip to the next song where playback will occur without any problems.

### Steps to reproduce

* refresh the browser
* navigate to a music library (which uses shakaplayer)
* initiate playback onto an album
* endless loading cycle

### Expected behavior

Playback of the selected albums first song.

### Platform

Linux

### Browser

Chrome

### Jellyfin server version

10.7.7

LPkkjHD:
After looking at the values via an inspector it seems that the `src` tag isn't properly populated.

PR #1755, opened 2022-05-03
fix(ShakaPlayer): first element in queue won't start playback

LPkkjHD:
With this patch ShakaPlayer will get a valid playback URL after creation of the player. Closes https://github.com/jellyfin/jellyfin-vue/issues/1754.

As `shakaPlayer` is structured differently as the `HlsPlayer` I'm unsure whether this is actually the right approach to do this and would appreciate some feedback on the way it's done.


ThibaultNocchi (approved):

ThibaultNocchi:
Thank you :)

PR #1759, opened 2022-05-07
feat(DLNA): create DLNA settings

LPkkjHD:
This PR tracks the progress of the implementation of the DLNA settings page. Required for this to be complete the following things have to be done:

- [x] create page
- [ ] allow creation of DlnaProfiles (currently only saving profiles with a new name/saving server profiles creates a new custom one)
- [ ] allow editing for DlnaProfiles
    - [x] Info
    - [x] Identification
    - [ ] Subtitle
    - [ ] Direct Play
    - [ ] Transcoding
    - [ ] Containers
    - [ ] Codecs
    - [ ] Responses
    - [ ] XML
- [ ] adding translation strings

For the visual design I tried sticking to the scheme which is defined by the pages of `settings` as well as the `devices` and `api keys`. With this card based approach I'm planning to unify the "devices" and "settings" pages which used to be separated by a drawer into a single one which should provide a better user experience.

### `SelectedDlnaProfile`'s template is massive
Normally I'd split the template into multiple subcomponents, however, this approach doesn't seem very feasible here as the `currentProfile` holding the information we want to edit/bind to components cannot be passed around easily and working with `$emit` hooks doesn't seem very elegant to me.

### API interaction is a mess
Currently every component I wrote calls the API and processes the data as it pleases. I think this should be centralized somewhere, but I'm uncertain of the way to do this properly.

### DLNA settings are visually weird
As it stands the DLNA settings (the ones on the left side) change the mouse to a weird pointer (and would highlight an entry if clicked) which is undesirable in my opinion. What I was trying to achieve was an effect similar to the one where the server information is displayed (light up on mouse hover, however, not clickable). Also the text boxes aren't aligned properly. Maybe there is another component from Vuetify which makes more sense here.

## Undocumented deprecation and removal of X-DLNA cap and X-DLNA doc
Those two flags can still be set by the stable client of jellyfin-web, however, they don't appear in any api documentation and setting them doesn't actually result in anything (the server doesn't process them). I dropped them as a result.

### DIFF
--- a/frontend/components/System/SelectedDlnaProfile.vue
+++ b/frontend/components/System/SelectedDlnaProfile.vue
@@ -1,0 +1,924 @@
+<template>
+  <v-card class="selected-dlna-profile">
+    <v-row v-if="isDialog" class="ma-0 justify-space-between align-center">
+      <v-card-title>{{ selectedProfile.Name }}</v-card-title>
+      <v-btn icon class="mr-2" @click="$emit('close-dialog')">
+        <v-icon>mdi-close</v-icon>
+      </v-btn>
+    </v-row>
+    <v-card-title v-else>{{ currentProfile.Name }}</v-card-title>
+    <v-tabs v-model="tab" centered dark icons-and-text>
+      <v-tabs-slider />
+      <v-tab v-for="i in 9" :key="i" :href="'#tab-' + i">
+        {{
+          $t(
+            'settings.dlna.profile.' +
+              categories[i - 1].toLocaleString() +
+              '.title'
+          )
+        }}
+      </v-tab>
+    </v-tabs>
+    <v-tabs-items v-model="tab">
+      <v-tab-item value="tab-1">
+        <v-list>
+          <!-- General -->
+          <v-list-group :value="true">
+            <template #activator>
+              <v-list-item-title>
+                {{ $t('settings.dlna.profile.info.general.title') }}
+              </v-list-item-title>
+            </template>
+            <v-list-item>
+              <v-list-item-content>
+                <v-list-item-title
+                  v-text="$t('settings.dlna.profile.info.general.name.text')"
+                />
+                <v-list-item-subtitle
+                  v-text="$t('settings.dlna.profile.info.general.name.info')"
+                />
+              </v-list-item-content>
+              <v-list-item-action>
+                <v-text-field v-model="currentProfile.Name" outlined dense />
+              </v-list-item-action>
+            </v-list-item>
+            <v-list-item>
+              <v-list-item-content>
+                <v-list-item-title
+                  v-text="
+                    $t('settings.dlna.profile.info.general.userLibrary.text')
+                  "
+                />
+                <v-list-item-subtitle
+                  v-text="
+                    $t('settings.dlna.profile.info.general.userLibrary.info')
+                  "
+                />
+              </v-list-item-content>
+              <v-list-item-action>
+                <v-select
+                  v-model="currentProfile.UserId"
+                  :items="users"
+                  item-text="Name"
+                  item-value="Id"
+                  outlined
+                  dense
+                  clearable
+                  :label="$t('settings.dlna.DefaultUserId.selector')"
+                  reverse
+                />
+              </v-list-item-action>
+            </v-list-item>
+            <!-- Supported Media Types -->
+            <v-list-group :value="true" no-action sub-group>
+              <template #activator>
+                <v-list-item-title>
+                  {{ $t('settings.dlna.profile.info.general.mediaTypes.text') }}
+                </v-list-item-title>
+              </template>
+              <v-list-item>
+                <v-list-item-action>
+                  <v-checkbox v-model="supportedMediaTypes" value="Audio" />
+                </v-list-item-action>
+                <v-list-item-content>
+                  <v-list-item-title
+                    v-text="
+                      $t('settings.dlna.profile.info.general.mediaTypes.audio')
+                    "
+                  />
+                </v-list-item-content>
+              </v-list-item>
+              <v-list-item>
+                <v-list-item-action>
+                  <v-checkbox v-model="supportedMediaTypes" value="Photo" />
+                </v-list-item-action>
+                <v-list-item-content>
+                  <v-list-item-title
+                    v-text="
+                      $t('settings.dlna.profile.info.general.mediaTypes.photo')
+                    "
+                  />
+                </v-list-item-content>
+              </v-list-item>
+              <v-list-item>
+                <v-list-item-action>
+                  <v-checkbox v-model="supportedMediaTypes" value="Video" />
+                </v-list-item-action>
+                <v-list-item-content>
+                  <v-list-item-title
+                    v-text="
+                      $t('settings.dlna.profile.info.general.mediaTypes.video')
+                    "
+                  />
+                </v-list-item-content>
+              </v-list-item>
+            </v-list-group>
+            <!-- back to normal -->
+            <v-list-item>
+              <v-list-item-content>
+                <v-list-item-title
+                  v-text="
+                    $t(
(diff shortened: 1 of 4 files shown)


frontend/components/System/SelectedDlnaProfile.vue
```diff
@@ -0,0 +1,924 @@
...
+              categories[i - 1].toLocaleString() +
+              '.title'
+          )
+        }}
+      </v-tab>
+    </v-tabs>
+    <v-tabs-items v-model="tab">
+      <v-tab-item value="tab-1">
+        <v-list>
+          <!-- General -->
```
ferferga, line 25: As you say, this file is massive. However, why it can't be ``v-for`` _looped_? You can loop over ``currentProfile`` with Object.entries like ``(key, value) in Object.entries(currentProfile)`` and bind the value to ``v-model``.

frontend/components/System/SelectedDlnaProfile.vue
```diff
@@ -0,0 +1,924 @@
...
+            <!-- Supported Media Types -->
+            <v-list-group :value="true" no-action sub-group>
+              <template #activator>
+                <v-list-item-title>
+                  {{ $t('settings.dlna.profile.info.general.mediaTypes.text') }}
+                </v-list-item-title>
+              </template>
+              <v-list-item>
+                <v-list-item-action>
+                  <v-checkbox v-model="supportedMediaTypes" value="Audio" />
```
ferferga, line 81: Same for this, you can loop through the enum

frontend/pages/settings/dlna.vue
```diff
@@ -0,0 +1,249 @@
...
+        {{ $t('settings.help') }}
+      </v-btn>
+    </template>
+    <template #content>
+      <v-col cols="12" md="6" lg="5" class="py-4">
+        <v-card class="mb-4">
+          <v-list two-line flat>
+            <v-list-item-group>
+              <v-list-item
+                v-for="dlnaSettingName in Object.keys(bindToRelevantConfig)"
```
ferferga, line 18:
```suggestion
                v-for="(key, value) in Object.entries(bindToRelevantConfig)"
```

Soy you don't need to do ``dlnaSettings[dlnaSettingName]`` down below

frontend/pages/settings/dlna.vue
```diff
@@ -0,0 +1,249 @@
...
+                  />
+                  <v-list-item-subtitle
+                    v-text="
+                      $t('settings.dlna.' + dlnaSettingName + '.description')
+                    "
+                  />
+                </v-list-item-content>
+                <v-list-item-action>
+                  <v-switch
+                    v-if="typeof dlnaSettings[dlnaSettingName] === 'boolean'"
```
ferferga, line 33: Why we need typeof here? Do we know if the API can return something else that it's not a boolean?

frontend/components/System/SelectedDlnaProfile.vue
```diff
@@ -0,0 +1,924 @@
+<template>
```
ferferga, line 1: ``DLNAProfileEditor`` seems a more suitable name?

frontend/pages/settings/dlna.vue
```diff
@@ -0,0 +1,249 @@
...
+      this.dlnaProfiles = (await this.$api.dlna.getProfileInfos()).data;
+      this.closeDialog();
+    },
+    async deleteSelectedProfile(): Promise<void> {
+      await this.$api.dlna.deleteProfile({
+        profileId: this.currentProfileId
+      });
+      this.dlnaProfiles = (await this.$api.dlna.getProfileInfos()).data;
+      this.closeDialog();
+    },
```
ferferga, line 241: Are we giving any visual hint that these functions are running? I.E an spinning circle, an snackbar message, etc...

ferferga:
DLNA itself might be moved to a plugin, this is why the API might not be so much love. Still a good idea to have this page, just I would say not to worry so much about it looking extremely pretty, just good enough is good I think.

frontend/components/System/SelectedDlnaProfile.vue
LPkkjHD: I'm currently working on this, however, the way you suggested is not feasible as we would modify data of the loop within it. (v-modeling `value` results in this eslint warning being thrownhttps://eslint.vuejs.org/rules/valid-v-model.html)

frontend/pages/settings/dlna.vue
LPkkjHD: As mentioned above this approach isn't workable.

frontend/pages/settings/dlna.vue
LPkkjHD: We do need typeof here as we want to display different control elements (Switches for boolean values, Textfields for strings, Selectors for Enum based values) based on which type the data has we want to control.

This seemed the best approach from my perspective as it allows us to loop through all relevant data without needing to care which exact control we want to display for each field.

frontend/components/System/SelectedDlnaProfile.vue
ferferga: Try ignoring the rule, might be a false positive but work perfectly during runtime.

frontend/pages/settings/dlna.vue
```diff
@@ -0,0 +1,249 @@
...
+
+interface DlnaNamedConfiguration extends RelevantConfig {
+  EnablePlayToTracing: boolean;
+  BlastAliveMessageIntervalSeconds: number;
+  AutoCreatePlayToProfiles: boolean;
+  SendOnlyMatchedHost: boolean;
+}
+
+export default Vue.extend({
+  components: { DlnaProfileEditor },
```
ferferga, line 151: This is unnecessary as componentes are auto imported.

frontend/pages/settings/dlna.vue
LPkkjHD: We don't currently, however, I can include a snackbar message

frontend/locales/en-US.json
```diff
@@ -345,6 +345,18 @@
         "description": "Determines which user library should be displayed on connected devices. This can be overridden for each device using profiles.",
         "selector": "Select default user"
       },
+      "EnablePlayToTracing": {
+        "text": "Enable PlayTo Tracing",
+        "description": "Enables debug logging for DLNA PlayTo"
+      },
+      "BlastAliveMessageIntervalSeconds": {
```
LPkkjHD, line 352: Was wondering why we seemingly have two properties describing the same setting. Turns out the server team planned to deprecate and remove this setting a couple of years ago, however, never pulled through. https://github.com/jellyfin/jellyfin/blob/master/Emby.Dlna/Configuration/DlnaOptions.cs#L59

What puzzles me is the fact that setting the "correct" value with `AliveMessageIntervalSeconds` is not processed by the server.

I left both in, however, marked the working setting as deprecated as it seems to be looking at the backend code.

ferferga: This is basically to keep backwards compat with old clients (like Web).

The fact that AliveMessageIntervalSeconds is not processed is something that @jellyfin/server needs to chime in.

crobibero: I don't think anyone has worked on the DLNA code for quite a while... so who knows what will happen.

ferferga: @LPkkjHD What about opening a PR in server to attach the logic of ``BlastAliveMessageIntervalSeconds`` to ``AliveMessageIntervalSeconds``, so both do the same while the deprecation process is still going?

LPkkjHD: This behavior is already implemented. I worked my way around it in https://github.com/jellyfin/jellyfin-vue/pull/1759/commits/e32296183dd96d7923ec80f745963ff4c0a54a50

LPkkjHD:
I think that's enough tinkering around for the moment. As I hit a wall and thought of a better way to implement this I'll do a rewrite from scratch in the next couple of days.

PR #1762, opened 2022-05-14
Queue manager improvements

ferferga:
* Add a transition when switching the text color between non playing tracks and the playing track.
* Disable Play and shuffle buttons in library view when the view is loading or there are no items.
* Fix item jumping in the draggable queue
* Fix animation quirk when the queue was opened and closed too fast by clearing the timeout during the DOM nodes destroy process.
* Current playing item in queue has now bold text.
* Queue is automatically scrolled into the currently playing item (related to #962)
* Fix item images being blurred on all queue changes in queue manager.

### DIFF
--- a/frontend/components/Buttons/QueueButton.vue
+++ b/frontend/components/Buttons/QueueButton.vue
@@ -122,7 +122,7 @@ export default Vue.extend({
     return {
       menu: false,
       destroy: false,
-      timeout: -1
+      timeout: undefined as undefined | number
     };
   },
   computed: {
@@ -199,7 +199,7 @@ export default Vue.extend({
         }, 500);
       } else {
         window.clearTimeout(this.timeout);
-        this.timeout = -1;
+        this.timeout = undefined;
         this.destroy = false;
       }
     }
--- a/frontend/components/Buttons/QueueButton.vue
+++ b/frontend/components/Buttons/QueueButton.vue
@@ -121,7 +121,8 @@ export default Vue.extend({
   data() {
     return {
       menu: false,
-      destroy: false
+      destroy: false,
+      timeout: -1
     };
   },
   computed: {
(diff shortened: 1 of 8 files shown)


frontend/components/Buttons/QueueButton.vue
```diff
@@ -193,10 +194,12 @@ export default Vue.extend({
   watch: {
     menu(): void {
       if (!this.menu) {
-        setTimeout(() => {
+        this.timeout = window.setTimeout(() => {
           this.destroy = true;
         }, 500);
       } else {
+        window.clearTimeout(this.timeout);
+        this.timeout = -1;
```
ThibaultNocchi, line 202: maybe prefer setting undefined?

frontend/components/Players/DraggableQueue.vue
```diff
@@ -85,6 +89,23 @@ export default Vue.extend({
...
+  mounted() {
+    const ref = this.$refs.listItems as Vue[];
+    const currentItemId = this.playbackManager.getCurrentItem?.Id || '';
+
+    const el = ref.find(
+      (v) => v.$vnode.key === `${currentItemId}-${ref.indexOf(v)}`
+    );
+
+    if (el?.$el) {
+      window.requestIdleCallback(() => {
```
ThibaultNocchi, line 104: Why do we use it? I'm only curious I trust you there's a reason ^^

ferferga: Good catch!

I wanted to document it, but forgot about it while testing. I added the comment directly above the function, but here's for reference as well:

```
/**
       * As the queue opening has a transition effect, el.$el.scrollIntoView() doesn't work directly,
       * as the parent DOM node is not fully rendered while the transition is taking place
       * (so scrollIntoView() doesn't know exactly what to scroll).
       *
       * The browser always give full priority to the DOM manipulation and painting process,
       * while the idle callback runs with lower priority. This assures us that the view will be scrolled to
       * the currently playing element as soon as all the DOM operations and transitions are over.
       */
```

ferferga:
@jellyfin-bot rebase

PR #1771, opened 2022-06-02
feat(playback): improve queue options and minor refactor to player manager

ferferga:
* Move MediaSession API handling to playbackWatcher
* Improve context menu in queue manager (based on comments in #1753)
* All "Play Next" options will, by default, append the item(s) to the next position in queue, but if some items already exist in the queue, they will be moved (not added again, so duplicates can only be made by using the "Add to queue" option).

_For current playing item_
image

_For other items_
image


ThibaultNocchi (approved):
LGTM

PR #1784, opened 2022-08-14
Update en-US.json

alanazar:
As per Weblate's best practice:
"The string uses three dots (...) instead of an ellipsis character (…)"
Reference to error: https://translate.jellyfin.org/translate/jellyfin-vue/jellyfin-vue/en/?q=check%3Aellipsis+OR+dismissed_check%3Aellipsis&offset=1

### DIFF
--- a/frontend/locales/en-US.json
+++ b/frontend/locales/en-US.json
@@ -538,8 +538,8 @@
       "tasks": "Running tasks"
     },
     "tasks": {
-      "configSync": "Syncing settings...",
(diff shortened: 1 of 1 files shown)


ferferga:
Good catch!

alanazar:
Thank you! It's my 1st approved pull request :D 
Are you able to advise on this one as someone had raised a concern:
https://github.com/jellyfin/jellyfin-mpv-shim/pull/290

PR #1789, opened 2022-08-17
fix: missing home sections

ThibaultNocchi:
~~Inefficient fix to a situation where the array was never falsy so it never awaited for it to fetch.~~

~~It's inefficient cause the method is fetched 3 times across the app, and on this specific page it'll fetch every time if the user doesn't have views.~~

~~What could be done is that:~~

~~1. the store is fetched by default by the app once so that others aren't expected to fetch it~~
~~2. convert parts of the app using it to be reactive (this part in this file isn't for instance)~~

* moved the front page home sections logic to a getter to be reactive to the store changes
* nav drawer libraries were already reactive but it manually fetched them still
* fetch the user views at authentication (remembered or not)
* removed the homesection manual "libraries" setup to a getter of the userviews

### DIFF
--- a/frontend/pages/index.vue
+++ b/frontend/pages/index.vue
@@ -35,6 +35,8 @@ import {
   authStore
 } from '~/store';
 
+const VALID_SECTIONS = ['resume', 'resumeaudio', 'upnext', 'latestmedia'];
+
 export default Vue.extend({
   // TODO: Merge asyncData and fetch once we have Nuxt 3, so we can have proper Vue 3 suspense support and have all the data
   // loaded with a complete Vue instance but with the route not being rendered until the full data is loaded
@@ -70,8 +72,6 @@ export default Vue.extend({
   computed: {
     ...mapStores(clientSettingsStore, pageStore, userViewsStore),
     homeSections(): HomeSection[] {
-      const validSections = ['resume', 'resumeaudio', 'upnext', 'latestmedia'];
-
       // Filter for valid sections in Jellyfin Vue
       // TODO: Implement custom section order
       let homeSectionsArray = pickBy(
@@ -80,7 +80,7 @@ export default Vue.extend({
         (value: string, key: string) => {
           return (
             value &&
-            validSections.includes(value) &&
+            VALID_SECTIONS.includes(value) &&
             key.startsWith('homesection')
           );
         }
@@ -96,7 +96,6 @@ export default Vue.extend({
         };
       }
 
-      // Convert to an array
       homeSectionsArray = Object.values(homeSectionsArray);
 
       const homeSections: HomeSection[] = [];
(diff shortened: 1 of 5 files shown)


ferferga:
@jellyfin-bot rebase

frontend/store/homeSection.ts
```diff
@@ -152,12 +143,15 @@ export const homeSectionStore = defineStore('homeSection', {
...
   },
   getters: {
-    getHomeSectionContent:
-      (state) =>
-      (section: HomeSection): BaseItemDto[] => {
+    libraries: (): BaseItemDto[] => {
+      const userViews = userViewsStore();
+      return Array.from(userViews.views);
+    },
+    getHomeSectionContent(state) {
```
ferferga, line 150:
```suggestion
    getHomeSectionContent() {
```

Is this needed?

frontend/pages/index.vue
```diff
@@ -59,132 +59,122 @@ export default Vue.extend({
...
-              'boxsets',
-              'channels'
-            ];
-
-            for (const userView of this.userViews.views) {
-              if (
-                excludeViewTypes.includes(userView.CollectionType as string)
-              ) {
-                continue;
+      // Convert to an array
```
ferferga, line 99: This comment is a little bit meaningless imo, as it's implicit in the next line. A better comment would be one that answers the question of: Why we need this converted to an array?

frontend/pages/index.vue
```diff
@@ -59,132 +59,122 @@ export default Vue.extend({
...
-          break;
+  head() {
+    return {
+      title: this.page.title
+    };
+  },
+  computed: {
+    ...mapStores(clientSettingsStore, pageStore, userViewsStore),
+    homeSections(): HomeSection[] {
+      const validSections = ['resume', 'resumeaudio', 'upnext', 'latestmedia'];
```
ferferga, line 73: I think this kind of "hardcoded" value lives better outside the Vue component, so it's easier to locate later one (no need to revisit the whole component).

What do you think about adding it at line 37?

ferferga (changes requested):
@ThibaultNocchi Fix all lint errors as well please

frontend/store/homeSection.ts
ThibaultNocchi: Yes this is, for lines 156 up to 162 :)

frontend/pages/index.vue
ThibaultNocchi: I didn't put it in and I feel like it's unnecessary. So I think I should just let it here ^^

frontend/pages/index.vue
ThibaultNocchi: No problem, done :)

frontend/pages/index.vue
ThibaultNocchi: Done

ferferga (approved):
LGTM now

PR #1790, opened 2022-08-17
Fix/nav drawer

ThibaultNocchi:
* disable transparency on navigation drawer on mobile screens, which made it unusable when opened and transparency was enabled (when at the top of the home page for instance)
* remove negative margins on the appbar cause it made buttons too close to the extremities of the screen on mobile


ferferga:
@jellyfin-bot rebase

Issue #1794, opened 2022-09-14
Scrollbar showing in both directions


MominRaza:
### Description of the bug

In filters dropdown scrollbar showing in both directions instead of only Vertical direction.
![image](https://user-images.githubusercontent.com/42776853/190167374-e991c458-d068-4334-9a91-3f6b94433606.png)

### Steps to reproduce

Movies->Filter->Features

### Expected behavior

It should show scrollbar in vertical direction
![image](https://user-images.githubusercontent.com/42776853/190167912-211494c2-70db-44f6-b949-2f639de275bf.png)

### Platform

Windows

### Browser

Chrome

### Jellyfin server version

10.8.4

### Additional context

```text
Can easily fix by replacing
overflow: scroll;
by
overflow: hidden;
overflow-y: auto;
```

ferferga:
Check if this doesn't break anything else and open a PR, as you already have the fix.

Thank you very much!

MominRaza:
I'm not familiar with Vue and this project, but I know css and react
But I will try

sfkpmr:
#1773 Related?

PR #1795, opened 2022-09-15
Scrollbar now shows in only vertical directions

MominRaza:
#1794 Fixed


ferferga:
Thank you very much!

This is completely unrelated to your attempt to fix this, but now that you're touching this file... Would you mind checking why max-height is set to 15 rem? Maybe we can set this to a more sensible value (or remove the value altogether), so we also don't have vertical scroll? They're really ugly as well. Check also that it looks fine in mobile.

If you don't want to keep fiddling with it, don't worry, I can tackle it later.

ferferga:
@jellyfin-bot rebase

ferferga (approved):

PR #1798, opened 2022-09-24
fix(ci): remove broken windows build and rename workflows

dkanada:


ferferga:
I'm not a fan of simply removing the check just for the sake of having all green.

Before, we had a fast CI setup where the client was built outside the Docker context and then deployed to multiple architectures using a custom Dockerfile ust for CI (where the dist folder was just copied into the image). [ARM64 was broken for some time](https://github.com/jellyfin/jellyfin-vue/issues/1610) and it passed by unnoticed. Windows builds [were broken as well](https://github.com/jellyfin/jellyfin-vue/discussions/1674) and we had no idea until we also got Tauri and it got fixed in #1768

I recall having some discussions here and on Matrix with the intention to change that and build for all architectures we wanted to support, sacrificing speed for reproducible builds. I think you were going for reproducible builds, instead of CI-specific builds iirc. That's why we spend 1 hour in building every commit on push, because the process is done for all the architectures.

For instance, we could [build Tauri cross-platform](https://tauri.app/v1/guides/building/cross-platform/) on Linux, but we do it in their native OSes so we also have the chance to test how our client builts on other OSes (I think @ThibaultNocchi was the one that asked me why we didn't cross-build, and we rejected it basically for this "reproducible builds" thing. I can't find the discussion now, though). With this, we already know that Windows fails to build, which will make troubleshooting easier and avoid a #1674, where we weren't aware of anything at the beginning.

Windows build is failing just for Swiper right now, which is something that will be fixed once we migrate to Vue 3 (already working on it in my local repo). It's something that it's not blocking our workflow like https://github.com/npm/cli/issues/4618, where we needed https://github.com/jellyfin/jellyfin-vue/commit/9dba4080de5d8f314da36317b5c0d4df02951ec3 to keep having checks

TLDR: I think we should fix Windows build instead of removing it altogether.

dkanada:
First of all, I was extremely on board with removing the CI dockerfile citing "it should build the exact binaries that end up in the releases" as the specific reason. I don't know if I was part of the discussion about cross compiling for Windows and macOS on Linux but I would have been mostly indifferent on that one.

Second, I'm not arguing that we should remove the Windows build indefinitely, only that it should be disabled (or removed in this case) until the builds are working again. I didn't even notice the CloudFlare issue hiding behind the Windows failures. A broken CI is useless because as soon as you let tests fail problems start getting missed and forgotten like this one. An even worse situation would be that we ignored it since "the Windows builds are failing anyways" or actively decided to merge code that broke Windows or CloudFlare. I didn't see a specific commit that broke either workflow so I assume it was just the passage of time - the CloudFlare one specifically seems to have been intermittently failing for months, whereas the Windows failure is more recent.

dkanada:
Pinning the NPM version on CloudFlare's website fixed the issue, but my point still stands that it was completely hidden by the failing Windows builds. It makes renovate pull requests a pain too, since you can't immediately tell if they're okay to merge.

I've also seen new contributors question failing CI on countless projects in the past only to be told it was "unrelated" in those cases, adding confusion when the goal should be a very clear indication of whether or not anything needs to be addressed.

Issue #1779, opened 2022-07-31
[UX] Movie/Episode with no default video track have no default set in the Item UI


Extarys:
### Description of the bug

When selecting an episode to watch where the video track isn't set as default inside the file itself, the UI shows no video by default.

![image](https://user-images.githubusercontent.com/11581433/182004988-53ecf4d7-a58b-4451-a8f1-f9b3e7f36af4.png)

This is purely UX as the video will still play fine, but it's very confusing to see an audio track and subtitle track being selected without a video and may lead to confusion.

### Steps to reproduce

Preparation:
1. Put a video in MKVToolnix 
2. Mark the video track as not default
3. Save
4. Import into Jellyfin
5. Click the new new episode/movie

### Expected behavior

When watching a movie, an epsiode or a music video, the first video track should be selected by default, even if it's not marked as default in the file's metadata.

![image](https://user-images.githubusercontent.com/11581433/182004998-50d894a9-01f5-4ab4-8023-49db58dd5437.png)

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.8.1

### Additional context

Might be able to fix some small issues once I get my new RAM upgrade, I'm currently saturated and can't do much coding. :/

vallezw:
Is this resolved or can I try and fix it?

ferferga:
@vallezw Take in mind we're currently working in the ``vite`` branch. I think this issue might've been fixed in ``vite`` branch already

vallezw:
Alright

ferferga:
@vallezw Please check in the Cloudflare development of [this PR](https://github.com/jellyfin/jellyfin-vue/pull/1812) and let me know so we can close this issue

vallezw:
Ok I'll check

vallezw:
@ferferga I just tested it and it seems to be working now, so the issue can be closed. If you want support on any other issue please keep me in mind since I really like this project and especially this modern client and would love to help out.

ferferga:
@vallezw Thank you very much for your interest.

I cheer you up to start diving in the code, it might seem daunting at first, but with time you will end up familiarising.

If still not your cup of tea, translations to your language or triaging issues like this is really welcome!

PR #1799, opened 2022-10-08
fix: Movie/Episode with no default video track fix

maulik9898:
Fix to a situation where a Movie/Episode with no default video track has no default set in the Item UI

[1779](https://github.com/jellyfin/jellyfin-vue/issues/1779)


ferferga:
@jellyfin-bot rebase

ferferga (approved):
Thank you very much!

PR #1806, opened 2022-11-11
feat(player): add double click to fullscreen

### LINKED ISSUE #1775: Double-clicking video doesn't switch to fullscreen
### Description of the bug

Double-clicking video is expected to switch to fullscreen.

### Steps to reproduce

1. Start playback of video
2. Double-click video

### Expected behavior

Video changes to fullscreen on double-click.

### Logs

_No response_

### Screenshots

_No response_

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.8.1

### Additional context

```text
Firefox 101.0.1.
Same issue in Chromium 102.0.5005.115.
```

tmneth:
**Changes**
Fullscreen can now be toggled on doubleclick.

**Issues**
Closes #1775

### DIFF
--- a/frontend/components/Players/PlayerManager.vue
+++ b/frontend/components/Players/PlayerManager.vue
@@ -254,6 +254,7 @@ export default Vue.extend({
     return {
       showFullScreenOverlay: false,
       fullScreenOverlayTimer: null as number | null,
+      clickTimer: null as number | null,
       keepOpen: false,
       playbackData: false,
       isUpNextVisible: false,
@@ -290,6 +291,7 @@ export default Vue.extend({
             window.addEventListener('mousemove', this.handleMouseMove);
             window.addEventListener('keyup', this.handleKeyPress);
             window.addEventListener('click', this.handleVideoClick);
+            window.addEventListener('dblclick', this.handleVideoDoubleClick);
           }
 
           break;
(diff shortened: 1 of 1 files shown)


frontend/components/Players/PlayerManager.vue
```diff
@@ -254,6 +254,7 @@ export default Vue.extend({
     return {
       showFullScreenOverlay: false,
       fullScreenOverlayTimer: null as number | null,
+      clickTimer: null as number | null,
```
ferferga, line 257: Why we need this?

tmneth: To prevent dblclick from also firing single click event. So basically, If we do not get another click during that time span, we're firing pause, otherwise fullscreen.

ferferga: @tmneth But have you tested without it? Which problems did you encounter without it? 

Browser's double click event (which you're using) should already handle all of that for you.

https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event

tmneth: @ferferga I have tried to use dblclick and click without a timeout, however, in such case the click handler is triggered when i doubleclick. So doubleclick would both pause and trigger fullscreen at the same time.

ferferga (approved):
There's a bit of lag when clicking pause, but hopefully we can revisit this later on Vue 3 migration, where (probably) a huge chunk of this file is going to be rewritten.

Thank you very much for your first contribution!

PR #1808, opened 2022-11-11
Added method to show Albums, EPs, and Singles separately

BrianCArnold:
Added method to show Albums, EPs, and Singles separately from complete discography.

I'm not a Vue dev, so I might have done things wrong. I wanted the breakpoints between Album/EP/Single to be configurable to a user's preference in the future if desired, so it's stored in a state store so it's not just hardcoded.

Right now it defaults to showing the Albums only if it discovers at least one release that's of Album length.

### DIFF
--- a/frontend/pages/artist/_itemId/index.vue
+++ b/frontend/pages/artist/_itemId/index.vue
@@ -128,6 +128,7 @@ import { getImageInfo, getBlurhash, ImageUrlInfo } from '~/utils/images';
 import { getItemDetailsLink, isValidMD5 } from '~/utils/items';
 import { pageStore, authStore } from '~/store';
 import ArtistTab from '~/components/Layout/Artist/ArtistTab.vue';
+import { msToTicks } from '~/utils/time';
 
 export default Vue.extend({
   components: { ArtistTab },
@@ -140,8 +141,8 @@ export default Vue.extend({
   },
   async asyncData({ params, $api }) {
     const albumBreakpoints = {
-      singleTicksMaxLength: 6000000000,
-      epTicksMaxLength: 18000000000
+      singleMsMaxLength: 600000,
+      epMsMaxLength: 1800000
     };
     const auth = authStore();
     const itemId = params.itemId;
@@ -168,19 +169,19 @@ export default Vue.extend({
     const singles = discography?.filter(
       (album: BaseItemDto) =>
         (album?.RunTimeTicks || album?.CumulativeRunTimeTicks || 0) <=
-        albumBreakpoints.singleTicksMaxLength
+        msToTicks(albumBreakpoints.singleMsMaxLength)
     );
     const eps = discography?.filter(
       (album: BaseItemDto) =>
         (album?.RunTimeTicks || album?.CumulativeRunTimeTicks || 0) >
-          albumBreakpoints.singleTicksMaxLength &&
+          msToTicks(albumBreakpoints.singleMsMaxLength) &&
         (album?.RunTimeTicks || album?.CumulativeRunTimeTicks || 0) <=
-          albumBreakpoints.epTicksMaxLength
+          msToTicks(albumBreakpoints.epMsMaxLength)
     );
     const albums = discography?.filter(
       (album: BaseItemDto) =>
         (album?.RunTimeTicks || album?.CumulativeRunTimeTicks || 0) >
-        albumBreakpoints.epTicksMaxLength
+        msToTicks(albumBreakpoints.epMsMaxLength)
     );
 
     const appearances = (
--- a/frontend/pages/artist/_itemId/index.vue
+++ b/frontend/pages/artist/_itemId/index.vue
@@ -127,11 +127,9 @@ import { sanitizeHtml } from '~/utils/html';
 import { getImageInfo, getBlurhash, ImageUrlInfo } from '~/utils/images';
 import { getItemDetailsLink, isValidMD5 } from '~/utils/items';
 import { pageStore, authStore } from '~/store';
-import ArtistTab from '~/components/Layout/Artist/ArtistTab.vue';
 import { msToTicks } from '~/utils/time';
 
 export default Vue.extend({
-  components: { ArtistTab },
   meta: {
     backdrop: true,
     transparentLayout: true
--- a/frontend/pages/artist/_itemId/index.vue
+++ b/frontend/pages/artist/_itemId/index.vue
@@ -40,13 +40,22 @@
             <v-tab :key="0" :disabled="!discography.length">
               {{ $t('item.artist.discography') }}
             </v-tab>
-            <v-tab :key="1" :disabled="!appearances.length">
+            <v-tab :key="1" :disabled="!albums.length">
+              {{ $t('item.artist.albums') }}
+            </v-tab>
+            <v-tab :key="2" :disabled="!eps.length">
+              {{ $t('item.artist.eps') }}
+            </v-tab>
+            <v-tab :key="3" :disabled="!singles.length">
+              {{ $t('item.artist.singles') }}
+            </v-tab>
+            <v-tab :key="4" :disabled="!appearances.length">
               {{ $t('item.artist.appearsOn') }}
             </v-tab>
-            <v-tab :key="2" :disabled="!musicVideo.length">
+            <v-tab :key="5" :disabled="!musicVideo.length">
               {{ $t('item.artist.videos') }}
             </v-tab>
-            <v-tab :key="3" :disabled="!artistBackdrop.tag && !overview">
+            <v-tab :key="6" :disabled="!artistBackdrop.tag && !overview">
               {{ $t('item.artist.information') }}
             </v-tab>
           </v-tabs>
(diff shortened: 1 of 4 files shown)


BrianCArnold:
Here's what the UI looks like right now, with Albums, EPs, Singles added after Discography:

image

frontend/pages/artist/_itemId/index.vue
```diff
@@ -174,6 +292,24 @@ export default Vue.extend({
...
+          albumBreakpoints.singleTicksMaxLength &&
+        (album?.RunTimeTicks || album?.CumulativeRunTimeTicks || 0) <=
+          albumBreakpoints.epTicksMaxLength
+    );
+    const albums = discography?.filter(
+      (album: BaseItemDto) =>
+        (album?.RunTimeTicks || album?.CumulativeRunTimeTicks || 0) >
+        albumBreakpoints.epTicksMaxLength
+    );
+
```
reedstrm, line 312: Feels to me like this business logic should live in a separate ts file, perhaps? Not an expert in either Vue or how this project is using it, just interested Jellyfin user. :)

BrianCArnold: Not a Vue expert either, but it feels more like display logic than business logic to me.

Maxplosion:
I'm curious, would a "Mini Album" in this case then also be categorized as an EP? 
This naming is a standard e.g. in K-Pop

BrianCArnold:
> I'm curious, would a "Mini Album" in this case then also be categorized as an EP? This naming is a standard e.g. in K-Pop

I'm only doing checks on the total length of the release, so everything under Single, EP, or Album based on if it's less than 10 minutes, longer than 10 minutes but less than 30, or longer than 30 minutes, as those are generally considered to be the length breakpoints in those categories. If those mini-albums are between 10 and 30 minutes long, they would fall under "EP", yes.

If you have an idea how to categorize them differently, I can see if there's a good way to work it in.

Maxplosion:
> > I'm curious, would a "Mini Album" in this case then also be categorized as an EP? This naming is a standard e.g. in K-Pop
> 
> I'm only doing checks on the total length of the release, so everything under Single, EP, or Album based on if it's less than 10 minutes, longer than 10 minutes but less than 30, or longer than 30 minutes, as those are generally considered to be the length breakpoints in those categories. If those mini-albums are between 10 and 30 minutes long, they would fall under "EP", yes.
> 
> If you have an idea how to categorize them differently, I can see if there's a good way to work it in.

roger that, I just did a quick check of some minis and in general those are between 4 and 7 tracks, less than 30min and least time was around 14min, so at least those should be categorized into EP then. 

for 'single albums' that will be less ideal, as those often have e.g. 2 tracks and then those two as instrumentals which can push them (with 4 tracks in total) above 10min. (Example Album NMIXX - AD MARE, 11min 22sec)

But I don't know how to fix that, just wanted to let you know. 
Maybe a setting where you can adjust the thresholds (e.g. single until 13min and ep until 25min and so on) might be a nice middle ground for most users with more special libraries.

But anyway, thanks for the PR, imo it increases the "beauty" of the library significantly if it is more structured.

BrianCArnold:
Another option may be to combine EPs and Singles into a single tab, and keep Albums separate. My main use case is to keep 30+ minute releases separate from < 30 minute releases, as longer releases are somewhat generally the "main" releases.

On the other hand, it's actually written to be configurable as to the time breakpoints, I just haven't written a configuration UI to set that. If we want it to be server wide instead of client specific, I would need to make it an environment variable during build, so you can change it by building your own container. 

The other option would be to move the configuration up to Jellyfin itself, which may require changing API options available to clients, including JF Vue.

Maxplosion:
> Another option may be to combine EPs and Singles into a single tab, and keep Albums separate. My main use case is to keep 30+ minute releases separate from < 30 minute releases, as longer releases are somewhat generally the "main" releases.

Keeping singles and EPs separated would be appreciated, it's a special case with K-Pop (with their singles/single albums) and with other genres the separation really can give a nice benefit for structuring the library.

> 
> On the other hand, it's actually written to be configurable as to the time breakpoints, I just haven't written a configuration UI to set that. If we want it to be server wide instead of client specific, I would need to make it an environment variable during build, so you can change it by building your own container.
> 
> The other option would be to move the configuration up to Jellyfin itself, which may require changing API options available to clients, including JF Vue.

Generally I'm a fan of having such things on the JF server as this allows other clients to also implement this logic (e.g. the old web client and other third party clients). And in general the server needs more music features/granularity anyway, so that is always a plus in my book.

frontend/store/index.ts
```diff
@@ -19,3 +19,4 @@ export * from './auth';
 export * from './items';
 export * from './playbackManager';
 export * from './taskManager';
+export * from './albumRuntimeBreakpoints';
```
ferferga, line 22: Where is this file in the PR?

Also, don't create an store just for this. For now, I think you can leave the values as const on top of the ``export default Vue.extend`` in the page content. Nothing is user-configurable at the moment, this kind of stuff should come up later and people that desire to make this kind of nitpicky changes can build the client themselves at the moment.

Adding configuration for every single thing right now is not feasible. Only major stuff that would live outside a convential config menu (like the history mode for Vue router or the default servers) are welcome in JSON config files (as they should be) or env vars (as they're now).

ferferga (changes requested):

ferferga:
Beside my main comment in this PR, thank you very much for contributing towards [this](https://github.com/jellyfin/jellyfin-vue/discussions/1679)

This kind of attention to the detail that could make Jellyfin feel "more native" for all the media types are the ones I'm really looking for the future of this client :)

frontend/store/index.ts
BrianCArnold: I must have failed to include it in my commit? I thought I did. I'll change it to your recommendation.

ferferga:
@jellyfin-bot rebase

Rebasing so I can test changes

BrianCArnold:
Force push to bring up to date with main? I'm away from my computer, can't easily see the push.

ferferga:
@BrianCArnold Yes, done with GitHub UI.
It's a requirement we have for all merged pull requests, that they're up to date with master. I also needed it to deploy to Cloudflare pages (fixed by me after you opened the PR) so I can test this.

Sorry for all the spam 😅.

BrianCArnold:
No worries! Also, I noticed that songs that aren't in a subdirectory under the artist name aren't shown under an artist at all.

So for the following: 

- Artist/Name of Single/01 - A Side.mp3
- Artist/Name of Single/01 - B Side.mp3
- Artist/Individual Song.mp3

`A Side` and `B Side` show up under discography (and now Singles), but `Individual Song` doesn't show up anywhere under the Artist.

If I search for the song itself, it does show up, and with the right Artist. And if I play all songs by an artist, they are in that playlist. My guess is that Jellyfin simply doesn't have a way to display songs that aren't considered part of an album/release. 

I'm seeing this in JFVue before and after my patch, the default Jellyfin UI, and even in other apps like Finamp. Is this an issue with how I have my library set up, or an actual bug I need to file and fix in the main Jellyfin repo?

Maxplosion:
Just quickly chiming in, as I have email notifs still on. 

In general files just in the artist root should work (https://jellyfin.org/docs/general/server/media/music). 

If that actually is true, I don't know myself as I don't have single files like this in all of my collection -- but before opening an issue, have you tested with the current master? 
There were a few music related changes merged to the server in the past weeks, so one of the might help here -- just throwing this out.

BrianCArnold:
I'll have to merge up to it, my local container has changes to support Intel Arc GPUs.

Maxplosion:
please see my edit of the former post, sorry.

Btw. just as input how I handle these singles: I move single files into a generic folder that I name the same as the file if it has no album itself, otherwise foldername = albumname - can be quite some extra work, but yeah, I see it as a hobby to play around with the file structures so it does not bother me that much.

BrianCArnold:
That's simple enough, I can experiment with it. And I'm on ZFS, so I can snapshot and play with the structure.

frontend/pages/artist/_itemId/index.vue
```diff
@@ -88,6 +97,114 @@
               </v-row>
             </v-tab-item>
             <v-tab-item :key="1">
+              <v-row no-gutters>
```
ferferga, line 100: Can you move the tab's content to a new component? That way we have less duplication. A fitting place is ``components/Layout/Pages/Artist/ArtistTab.vue`` in my opinion

frontend/pages/artist/_itemId/index.vue
```diff
@@ -152,6 +269,10 @@ export default Vue.extend({
     return isValidMD5(ctx.route.params.itemId);
   },
   async asyncData({ params, $api }) {
+    const albumBreakpoints = {
+      singleTicksMaxLength: 6000000000,
+      epTicksMaxLength: 18000000000
+    };
```
ferferga, line 275: Can we store these as ms and convert between both using [our helper](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/utils/time.ts#L14)?

Ticks is specific to .NET world and we talked some times about moving to seconds in the v2 API, besides is user friendlier for new developers to find most code references to milliseconds instead of "ticks".

ferferga (changes requested):
In one of the rebases, I took the liberty to format your commits following the [Conventional commit guide](https://www.conventionalcommits.org/en/v1.0.0/)

Don't hesitate to ping me if you are not versed in git and have troubles while rebasing your local changes :)

ferferga:
@BrianCArnold Any news about this?

BrianCArnold:
Haven't had time to go back and refactor it out to a component, with work and interview stuff.

I'm actually quite well versed in Git, it's just Vue that I'm not terribly familiar with, being an Angular/React-C# fullstack dev.

frontend/pages/artist/_itemId/index.vue
BrianCArnold: I'll switch it over.

frontend/pages/artist/_itemId/index.vue
BrianCArnold: I've moved it over, though I didn't use the exact path suggested.

frontend/pages/artist/_itemId/index.vue
BrianCArnold: I've switched to using ms instead of ticks and converted for comparison.

BrianCArnold:
Requested changes have been applied

frontend/pages/artist/_itemId/index.vue
```diff
@@ -142,8 +127,11 @@ import { sanitizeHtml } from '~/utils/html';
 import { getImageInfo, getBlurhash, ImageUrlInfo } from '~/utils/images';
 import { getItemDetailsLink, isValidMD5 } from '~/utils/items';
 import { pageStore, authStore } from '~/store';
+import ArtistTab from '~/components/Layout/Artist/ArtistTab.vue';
+import { msToTicks } from '~/utils/time';
 
 export default Vue.extend({
+  components: { ArtistTab },
```
ferferga, line 134:
```suggestion
import { msToTicks } from '~/utils/time';

export default Vue.extend({
```

@BrianCArnold Nuxt automatically imports components for us, so we don't need this.
Feel free to commit this suggestion through GitHub UI or do it locally if you want to squash it with another commit or place it somewhere else in your history.

ferferga (approved):
One minor thing though!

Notify me when I can hit merge!

frontend/pages/artist/_itemId/index.vue
BrianCArnold: Thanks, years of experience with Angular, but brand new to Vue.

BrianCArnold:
Everything should be ready to merge.

Maxplosion:
Works great!
Just one input for the future, it might be more convenient to default to "DISCOGRAPHY" instead of "ALBUMS" when opening an Artist page.

BrianCArnold:
> Works great!
> 
> Just one input for the future, it might be more convenient to default to "DISCOGRAPHY" instead of "ALBUMS" when opening an Artist page.

I considered that, however most EPs and Singles consist strictly of songs which are on albums, or are remixes of those songs. My thinking was that it would be best if all the songs are visible with the fewest things displayed.

Maxplosion:
I see yes that can make sense if the collection is like this. 
Maybe in the future a toggle could be added to set the default? as ~85% of my EPs and Singles are tracks that are not on albums so for some artists there might only be 1 album and 6EPs and 4 Singles, so the default view on album is rather empty as it omits most of the artists tracks.

BrianCArnold:
Ah, I see what you mean. If you can give me a few artist names where that's the case, I can try to research it and find some sort of threshold that would indicate that's the case for an artist.

Now that I think about it, this may present a bug.

An artist with no albums and only EPs and Singles would display nothing under "Albums" with the tab disabled.

Maxplosion:
I'll try to come up with a short list!

regarding the last point, yes then the Albums tab is greyed out and it defaults to Discography (I had this case with at least on artist when I quickly checked yday)

BrianCArnold:
Regarding setting a default, I can think of a couple different ways to do it. It can either be a server-wide setting using a --build-arg, like how admins build a docker image with a custom URI, or it could be a per-browser setting and saved in the browser's localStorage. That would mean that the same user switching to a differently browser, computer, etc would have to set it again.

Would you say that in your case, one of those is closer to your needs?

I'm also looking at extending the server API to provide the Album/EP/Single separation, along with configuration options to control where the breakpoints are between the three. Another thing I'm considering is doing the breaking up between the three categories based on number of tracks, or some combination of the current approach and that.

Maxplosion:
Personally I'd prefer a browser based setting, having to set it once for each browser (in general people use only one and that on maybe 2 devices) should not be an issue and removes the need for flags which a regular user cannot really use. 

I think having the possibility to adjust the breakpoints (e.g. time & number of tracks, or just time, or just tracks) would be the best case ofc. It allows people to adjust it to their collection (which can be very different as we can see just with our collections). Everybody has their special needs (but to cater to all is also imo not a possibility, so to find a middle ground which allows some customization but not so much that it hinders developers would be best). 

(but overall I'm quite satisfied with the current 10-30minute approach, it might not be perfect but it works in a majority of cases so that is already a big win in my book! Just the default is the one thing that makes it harder as I have to always do an extra click when I just want to quickly select a song from an artists collection)

As promised here's an example of an artist that has lots and lots of EPs and Singles that often times are only contained in their format (so in the EP or in the 'Single Album') and cannot be found in any of their full length albums: https://en.wikipedia.org/wiki/Red_Velvet_discography

PR #1814, opened 2022-11-15
docs: add comment in README to a node option for the SSL error

ThibaultNocchi:


README.md
```diff
@@ -74,6 +74,11 @@ This is an experimental web client for Jellyfin based on Vue.js. We welcome all
 
    ```bash
    npm start
+
+   # Run with the following env variable if you get an error such as this:
+   # Error: error:0308010C:digital envelope routines::unsupported
+   # It's _probably_ due to a conflict between recent OpenSSL versions and Node 16-17
```
ferferga, line 80: This doesn't get rendered as italic, but I guess that wasn't your intention and you're just using underscores to highlight it

ferferga (approved):

PR #1812, opened 2022-11-14
chore: upgrade to Vue 3

ferferga:
**THIS PR WILL BE MERGED ON MONDAY 3rd**

**TENTATIVE CHECKLIST**

*Might change depending on current situation, but I'll try to keep this up to date*

*I don't see all these tasks as mandatory for the merge, but I feel the most of them we had, better experience*

- [x] #1889
- [x] #1884
- [x] #1878 
- [x] https://github.com/vuetifyjs/vuetify/issues/16604
- [x] Create an entrypoint script for Docker, so environment variables are written to the new ``config.json`` file upon every container spin up: https://github.com/jellyfin/jellyfin-vue/blob/vite/frontend/public/config.json #1901
- [x] https://github.com/npm/cli/issues/6099 or https://github.com/actions/setup-node/issues/712 (related to #1886)
- [x] Get the maximum amount of components/pages refactored into Composition API and a lot of logic flaws (specially regarding Metadata editor) fixed #1892 
- [ ] Clean up the maximum amount of unused CSS classes and style hacks (Vuetify 2 needed a lot of those, but they no longer are necessary in a lot of places, so we should revisit every component to check that).
- [ ] Migrate to `MediaCapabilities` API for video and audio profile generation. This will remove a lot of legacy code and improve compatibility, since our profile generation is spaghetti. 
- [x] Get DraggableQueue working again (investigate VueUse's draggables or neodrag.dev, which seems a modern alternative to Sortable.JS, with better maintenance than the Sortable.JS Vue 3 version). #1909 
- [x] Fix the biggest amount of type and lint errors
- [x] Fix fullscreen music player style: #1899 
- [ ] Item store reactivity
- [x] Draft and tag a 0.2.0 release for master branch
- [ ] Merge this and all the renovate PRs in the following order. Tag 0.3.0 release afterwards:
       * #1738 
       * #1740 
       * #1770
       * #1765
       * #1838

PR #1827, opened 2022-12-14
fix(lang): Fixed pluralization for russian language

SoundwaveUwU:
This changes Vue-i18n configuration so that `$tc` function can be used for correct pluralization in russian (and other languages).

Function options cannot be stringified (as per [nuxt-18n docs](https://i18n.nuxtjs.org/options-reference#vuei18n)), so I had to move these options to separate file.

I'm not sure about place for options file, suggestions are welcome.

### DIFF
--- a/frontend/components/Buttons/QueueButton.vue
+++ b/frontend/components/Buttons/QueueButton.vue
@@ -47,7 +47,7 @@
             <v-list-item-subtitle>
               {{ getTotalEndsAtTime(playbackManager.getQueueItems) }} -
               {{
-                $t('playback.queueItems', {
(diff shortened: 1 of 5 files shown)


ferferga:
Blocked by #1812

The implementation in Vue 3 will be cleaner than the trick you're using in nuxt config, since we configure Vue i18n directly.

Take a look at the branch if you want to get accustomed, but don't PR anything to it as I rebase it constantly.

When that is merged (hopefully on Christmas, although a lot of that depends on Vuetify as well) we can get this merged 👍🏻.

ferferga:
@SoundwaveUwU Why did you close this?

PR #1845, opened 2023-01-15
Vite tnocchi 0113

ThibaultNocchi:


frontend/src/components/Layout/AppBar/AppBar.vue
```diff
@@ -58,12 +58,14 @@
...
 import { computed, inject, Ref } from 'vue';
 import { useRoute } from 'vue-router';
+import { useDisplay } from 'vuetify';
 import { useNetwork, useWindowScroll } from '@vueuse/core';
 import { clientSettingsStore } from '@/store';
 import { useResponsiveClasses } from '@/composables';
 
 const clientSettings = clientSettingsStore();
 const route = useRoute();
+const display = useDisplay();
```
ferferga, line 68: Unnecessary

frontend/src/components/Layout/AppBar/Buttons/TaskManagerButton.vue
```diff
@@ -68,128 +61,113 @@ interface TaskInfo {
...
+ * Set a timeout to hide the button
+ */
+function setTimeout(): void {
+  scheduledTimeout.value = window.setTimeout(clearState, props.timeout);
+}
+
+/**
+ * Locally retrieve and parse the tasks list
+ */
+function getTaskList(): void {
```
ferferga, line 104: Maybe this can be a computed property instead?

ThibaultNocchi: I let it that way cause there's a timeout that clears it after 5s or so, and I wasn't sure how it interacted with the task state. So I preferred to keep it that way for the migration.

PR #1850, opened 2023-01-16
Fix/cards

ThibaultNocchi:
* Convert the card item and its overlays to composition API
* Fix some styling
* Fix userViews never being reloaded once user is staying logged in by loading them on every index page load

### DIFF
--- a/frontend/src/components/Item/Card/Card.vue
+++ b/frontend/src/components/Item/Card/Card.vue
@@ -250,11 +250,6 @@ export default defineComponent({
   bottom: 0.5em;
 }
 
-.card-box {
-  cursor: pointer;
-  padding: 0;
-}
-
 .card-margin {
   margin: 0.6em;
 }
@@ -302,12 +297,6 @@ export default defineComponent({
 .card-image {
   width: 100%;
   height: 100%;
-}
-
-.card-chip {
-  position: absolute;
-  top: 1em;
-  right: 1em;
 }
 
 .card-progress {
--- a/frontend/src/components/Item/Card/Card.vue
+++ b/frontend/src/components/Item/Card/Card.vue
@@ -26,7 +26,7 @@
             v-if="item.UserData && item.UserData.UnplayedItemCount"
             color="primary"
             class="card-chip"
-            small>
+            size="small">
             {{ item.UserData.UnplayedItemCount }}
           </v-chip>
           <v-progress-linear
@@ -37,10 +37,11 @@
             "
             v-model="progress"
             color="primary-accent-4"
-            class="card-progress" />
+            absolute
+            location="bottom" />
         </div>
         <div
-          v-if="overlay && isFinePointer()"
+          v-if="overlay && isFinePointer"
           class="card-overlay d-flex justify-center align-center">
           <play-button fab :item="item" />
           <div
@@ -55,26 +56,32 @@
     </component>
     <div v-if="text" class="card-text">
       <router-link
-        class="link d-block card-title mt-1 text-truncate"
+        class="link d-block font-weight-medium pa-0 mt-1 text-truncate"
         :to="cardTitleLink">
         {{ cardTitle }}
       </router-link>
       <router-link
         v-if="cardSubtitleLink"
-        class="link d-block card-subtitle text--secondary text-truncate"
+        class="link d-block v-card-subtitle text-truncate"
         :to="cardSubtitleLink">
         {{ cardSubtitle }}
       </router-link>
-      <div v-else class="card-subtitle text--secondary text-truncate">
+      <div v-else class="v-card-subtitle text-truncate">
         {{ cardSubtitle }}
       </div>
     </div>
   </div>
 </template>
 
-<script lang="ts">
-import { defineComponent } from 'vue';
-import { BaseItemDto, ImageType } from '@jellyfin/sdk/lib/generated-client';
+<script lang="ts" setup>
+import { computed } from 'vue';
+import {
+  BaseItemDto,
+  BaseItemKind,
+  ImageType
+} from '@jellyfin/sdk/lib/generated-client';
+import { useI18n } from 'vue-i18n';
+import { useMediaQuery } from '@vueuse/core';
 import {
   CardShapes,
   getShapeFromItemType,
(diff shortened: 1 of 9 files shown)


frontend/src/components/Item/WatchedIndicator.vue
```diff
@@ -1,17 +1,9 @@
...
 </template>
 
 <script setup lang="ts" />
-
-<style lang="scss" scoped>
-.card-chip {
-  position: absolute;
-  top: 1em;
-  right: 1em;
-}
```
ferferga, line 16: Are we sure this class is needed both in Card component and here? I think it should only be necessary here (but one never knows with CSS!), so we don't need to decouple it at all (leaving it here only)

frontend/src/components/Item/Card/Card.vue
```diff
@@ -250,10 +250,10 @@ export default defineComponent({
...
 }
 
-.card-box {
-  cursor: pointer;
-  padding: 0;
-}
+// .card-box {
+//   cursor: pointer;
+//   padding: 0;
+// }
```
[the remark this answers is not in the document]
ThibaultNocchi: My bad, it's out

frontend/src/components/Item/WatchedIndicator.vue
ThibaultNocchi: Well the Card.vue file got this:

```vue
<v-progress-circular
            v-if="refreshProgress !== undefined"
            class="card-chip"
            rotate="-90"
            :model-value="refreshProgress"
            :indeterminate="refreshProgress === 0"
            color="white"
            size="24" />
```
and I checked without it, and I can assure you the class is needed lol

frontend/src/pages/index.vue
```diff
@@ -58,6 +58,8 @@ const carouselItems = (
 const userLibraries = userLibrariesStore();
 const clientSettings = clientSettingsStore();
 
+userLibraries.refresh();
```
ferferga, line 61: This must live in default layout (although I agree it makes more sense here).

With this here, the nav drawer, search etc... will appear for a while when content is loading. When the await is in default, we're blocking the page rendering at all, so the only thing that needs to load when reaching index is the carousel items (which is soemthing we can move to the store if we want content to be loaded all at once).

frontend/src/components/Item/ItemMenu.vue
```diff
@@ -57,6 +54,7 @@ import IMdiPencilOutline from 'virtual:icons/mdi/pencil-outline';
 import IMdiShuffle from 'virtual:icons/mdi/shuffle';
 import IMdiReplay from 'virtual:icons/mdi/replay';
 import IMdiRefresh from 'virtual:icons/mdi/refresh';
+import mdiDotsHorizontal from 'virtual:icons/mdi/dots-horizontal';
```
ferferga, line 57:
```suggestion
import IMdiDotsHorizontal from 'virtual:icons/mdi/dots-horizontal';
```

**PascalCase**. Told you in Matrix, but leaving this here as a note as well. Do the same for all your icon imports please (also, maybe we can investigate if we can enforce this through ``eslint-plugin-imports``

frontend/src/components/Item/Card/Card.vue
```diff
@@ -83,176 +85,134 @@ import {
...
+
+/**
+ * Gets the library update progress
+ */
+const refreshProgress = computed(
+  () => taskManager.getTask(props.item.Id || '')?.progress
+);
+
+const isFinePointer = computed(
+  () => window.matchMedia('(pointer:fine)').matches
```
ferferga, line 206: Not sure this is reactive, to be sure maybe we can switch to https://vueuse.org/core/usemediaquery/#usemediaquery ?

frontend/src/components/Item/Card/Card.vue
```diff
@@ -83,176 +85,134 @@ import {
...
-        }
-      }
+/**
+ * Returns either a string representing the production year(s) for the current item
+ * or the episode name of an item (SX EY - Episode Name)
+ * or the album artist
+ */
+const cardSubtitle = computed(() => {
+  switch (props.item.Type) {
+    case 'Episode': {
```
ferferga, line 126: Not sure if applicable here, but don't we have a ``BaseItemKind`` enum exposed in the API? Should be available now that we're using unstable.

Revisit all the uses of item types in this file if that enum exists please (I also see it in the computed property above this statement for example).

frontend/src/pages/index.vue
ferferga: This also must be awaitable, we want to block navigation to render the full UI at first load.

Fetching data like this in all pages will allow us to do some cool stuff with ``Suspense``  like showing a progress bar at the top: https://vuejs.org/guide/built-ins/suspense.html#events

ThibaultNocchi: Fixed in the meantime by always awaiting a refresh on the default layout and watching for page changes to the index page to refresh.

frontend/src/components/Item/Card/Card.vue
ThibaultNocchi: Used useMediaQuery

frontend/src/components/Item/ItemMenu.vue
ThibaultNocchi: Done

frontend/src/components/Item/Card/Card.vue
ThibaultNocchi: Done

frontend/src/layouts/default.vue
```diff
@@ -11,16 +11,26 @@
...
-}
+await userLibraries.refresh();
+
+watch(
+  () => route.name,
+  (newValue) => {
+    // @ts-expect-error - index isn't part of route values
+    if (newValue === 'index') {
+      userLibraries.refresh();
+    }
```
ferferga, line 31:
```suggestion
  async (newValue) => {
    // @ts-expect-error - index isn't part of route values
    if (newValue === 'index') {
      await userLibraries.refresh();
    }
```

I'd also add a comment like this:

```
Loading recent items from libraries can take a long time from server. To improve UX, we block navigation to the default layout when loading the client, so all the content is present when the page is finally rendered, avoiding content jumping.

We need to use a top-level await to block navigation, but we also want to ensure content is up-to-date when accessing the page (and that's why we trigger a refetch when accessing index as well).
```

frontend/src/layouts/default.vue
```diff
@@ -11,16 +11,26 @@
...
 import { userLibrariesStore } from '@/store';
 
 const display = useDisplay();
+const route = useRoute();
 const userLibraries = userLibrariesStore();
 const navDrawer = ref(!display.mobile.value);
 
-if (!userLibraries.isReady) {
-  await userLibraries.refresh();
-}
```
ferferga, line 23: This check must be in place because we also [load the content on login ](https://github.com/jellyfin/jellyfin-vue/blob/vite/frontend/src/components/Forms/LoginForm.vue#L108)

(Forgot to tell when we discussed this in Matrix, sorry!)

ThibaultNocchi: If I keep the check, the libraries won't be refreshed when your account is remembered and you first load the app.

ferferga: No, because we're using sessionStorage, so the data at userLibraries will be at it's defaults (entering this statement into effect when an user is logged in and you access the app) :)

ferferga (approved):

PR #1855, opened 2023-01-19
fix: Use VItemGroup to display settings list again

Janbong:
v-list-item-group seems to be not available anymore in Vuetify 3. I can't find any information on it. We use v-item-group (not v-list-group which was causing the settings not to show up) instead.
We use templates with v-slot:prepend and v-slot:append to place the icons correctly before or behind the actual settings name and description.

### DIFF
--- a/frontend/src/components/System/AboutLinks.vue
+++ b/frontend/src/components/System/AboutLinks.vue
@@ -7,7 +7,7 @@
         rel="noreferrer noopener"
         :href="linkItem.link"
         target="_blank">
-        <template v-slot:prepend>
+        <template #prepend>
           <v-avatar>
             <v-icon :icon="linkItem.icon" />
           </v-avatar>
@@ -15,7 +15,7 @@
         <v-list-item-title>
           {{ linkItem.name }}
         </v-list-item-title>
-        <template v-slot:append>
+        <template #append>
           <v-list-item-action>
             <v-icon>
               <i-mdi-open-in-new />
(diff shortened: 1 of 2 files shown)


frontend/src/components/System/AboutLinks.vue
```diff
@@ -1,25 +1,29 @@
...
       <v-list-item
         v-for="linkItem in linkItems"
         :key="linkItem.name"
         rel="noreferrer noopener"
         :href="linkItem.link"
         target="_blank">
-        <v-avatar>
-          <v-icon :icon="linkItem.icon" />
-        </v-avatar>
+        <template v-slot:prepend>
```
ferferga, line 10:
```suggestion
        <template #prepend>
```

This is the first occurence in the PR, but you used this syntax everywhere else. Revisit those as well please.

ferferga (changes requested):
One small thing and it's ready to go. 

This should have been reported as a warning by your code editor. If it didn't, tell us so we can try to help you configure VSCode appropiately :) (Codespaces are a quick and good way to start up quickly as well).

The vite branch is full of linter and type errors right now, but it's not that we don't care about them and adding code with warnings is good, it's just that it's a work in progress :).

frontend/src/components/System/AboutLinks.vue
Janbong: Done.

My VSCode did not warn me about it.
I am using the development container and installed all suggested extensions.
Something else I should have configured?

ferferga: @Janbong Did you wait for npm install to complete?

PR #1857, opened 2023-01-20
refactor(mediastreamselector): composition api

ThibaultNocchi:


frontend/src/components/Item/MediaStreamSelector.vue
```diff
@@ -1,204 +1,163 @@
...
-}
+const props = withDefaults(
+  defineProps<{
+    mediaStreams: MediaStream[];
+    type: string;
+    defaultStreamIndex?: number;
+  }>(),
+  { mediaStreams: () => [], defaultStreamIndex: undefined }
+);
+const emits = defineEmits(['input']);
```
ferferga, line 42: I believe this is not the most type-safe way to define an emit type. 

Check this for example: https://github.com/jellyfin/jellyfin-vue/blob/vite/frontend/src/components/Layout/Carousel/Carousel.vue#L79

ThibaultNocchi: True, I missed that it was available in Vue3. Fixed

PR #1858, opened 2023-01-20
fix: FilterButton

Janbong:
This fixes the Filter button. At least the UI part. 

A couple of problems:

- Filtering itself works (you can see the number of displayed items updating, 280 in the screenshot), but the displaying of filtered items acts weird sometimes. In the screenshot, underneath "280", you can see a small part of the title of one of the items.
image

- After applying a filter, I cannot change page without fully refreshing the full app. Not sure what's causing that. Help welcome.
- I wanted to use `dense` lists, like before. That prop has been removed in favor or `density="compact"`. But that does not seem to change anything. I compared the list of filters with the one in `master` branch and the difference is negligible, although more compact would be better if you ask me.

  | master  | this branch  | 
  |---|---|
  | image  |  image|

### DIFF
--- a/frontend/src/components/Buttons/FilterButton.vue
+++ b/frontend/src/components/Buttons/FilterButton.vue
@@ -1,54 +1,44 @@
 <template>
-  <v-menu :offset-y="true" :close-on-content-click="false" max-width="250px">
-    <template #activator="{ on, attrs }">
+  <v-menu :close-on-content-click="false" max-width="250px">
+    <template #activator="{ props: menuProps }">
       <v-btn
         v-if="!$vuetify.display.smAndDown"
-        :disabled="disabled"
         class="ma-2"
         variant="text"
         rounded
-        v-bind="attrs"
-        v-on="on">
+        v-bind="mergeProps(menuProps, props)">
         {{ $t('filter') }}
         <v-icon end>
           <i-mdi-menu-down />
         </v-icon>
       </v-btn>
-      <v-btn
-        v-else
-        :disabled="disabled"
-        class="my-2"
-        icon
-        v-bind="attrs"
-        v-on="on">
+      <v-btn v-else class="my-2" icon v-bind="mergeProps(menuProps, props)">
         <v-icon>
           <i-mdi-filter-variant />
         </v-icon>
       </v-btn>
     </template>
-    <v-expansion-panels accordion flat focusable class="dropdown">
+    <v-expansion-panels variant="accordion" class="dropdown">
       <v-expansion-panel :title="$t('status')">
-        <v-list dense class="filter-content">
-          <v-list-group
-            v-model="selectedStatusFilters"
-            multiple
-            @change="emitFilterChange">
+        <v-expansion-panel-text>
+          <v-list
+            v-model:selected="selectedStatusFilters"
+            select-strategy="leaf"
+            class="filter-content"
+            @update:selected="emitFilterChange">
             <template
               v-for="(status, statusIndex) in statusFilters"
               :key="`status-${statusIndex}`">
-              <v-list-item :value="status.name">
-                <template #default="{ active }">
-                  <v-list-item-title>
-                    {{ status.label }}
-                  </v-list-item-title>
-                  <v-list-item-action>
-                    <v-checkbox :model-value="active" />
+              <v-list-item :value="status.name" :title="status.label">
+                <template #append="{ isActive }">
+                  <v-list-item-action end>
+                    <v-checkbox-btn :model-value="isActive" />
                   </v-list-item-action>
                 </template>
               </v-list-item>
             </template>
-          </v-list-group>
-        </v-list>
+          </v-list>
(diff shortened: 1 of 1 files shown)


frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -1,182 +1,180 @@
...
+        <v-expansion-panel-text>
+          <v-list
+            select-strategy="multiple"
+            class="filter-content"
+            v-model:selected="selectedStatusFilters"
+            @update:selected="emitFilterChange">
             <template
               v-for="(status, statusIndex) in statusFilters"
               :key="`status-${statusIndex}`">
               <v-list-item :value="status.name">
```
ThibaultNocchi, line 33:
```suggestion
              <v-list-item :value="status.name" :title="status.label">
```
I'd rather you use the `title` attribute instead of the `<v-list-title>` element (same for the other occurrences)

ThibaultNocchi (changes requested):

frontend/src/components/Buttons/FilterButton.vue
Janbong: That's better indeed. Done.

ferferga (commented):
LGTM, but can we have this migrated to Composition API as well, please?
Same hints as I mentioned in #1865 apply here

Janbong:
> LGTM, but can we have this migrated to Composition API as well, please? Same hints as I mentioned in #1865 apply here

Working on it :)

frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -1,324 +1,299 @@
...
+      yearFilters.value = response.Years;
     }
+  } catch {
+    useSnackbar(t('filtersNotFound'), 'error');
   }
-});
+}
+function emitFilterChange(): void {
+  console.log('GENRE');
+  console.log(selectedGenreFilters);
```
ferferga, line 281:
```suggestion
```

frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -1,324 +1,299 @@
...
+    label: t('themeSong'),
+    name: 'HasThemeSong'
+  },
+  {
+    label: t('themeVideo'),
+    name: 'HasThemeVideo'
+  }
+];
+const selectedFeatureFilters = ref([]);
+var genreFilters = ref([]);
```
ferferga, line 235: refs must be always be consts. Verify this for all the occurences. Do you have IDE warnings?

frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -1,324 +1,299 @@
...
-              includeItemTypes: [this.itemsType]
-            })
-        ).data;
+  { label: t('liked'), name: ItemFilter.Likes },
+  { label: t('unliked'), name: ItemFilter.Dislikes }
+];
+var selectedStatusFilters = ref([]);
+const featureFilters = [
+  {
+    label: t('subtitles'),
```
ferferga, line 217: So when changing languages the labels are also translated, even if not in a computed property? Nice!

frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -1,324 +1,299 @@
...
-    },
-    disabled: {
-      type: Boolean,
-      required: false,
-      default: false
-    }
+const route = useRoute();
+const remote = useRemote();
+const { t } = useI18n();
+const emit = defineEmits(['change']);
```
ferferga, line 178: Define emits properly. See how I did it here: https://github.com/jellyfin/jellyfin-vue/blob/vite/frontend/src/components/Layout/Carousel/Carousel.vue#L79

ferferga (changes requested):

ferferga:
@Janbong I commented it in your other PR: let's try to stick with defining refs at the top of script setup (just like we did with data).
With a bit of consistency, things are more manageable :).

frontend/src/components/Buttons/FilterButton.vue
Janbong: I just double checked this. Not the case. How can we fix this then?

Janbong: OK, labels are translated, but only after a full page refresh. Not dynamically. Do we find this good enough?

ferferga: No, they must be translated instantaneously for the best UX.

Fixing it is as simple as wrapping in ``computed`` whatever object you declared that contains calls to ``t``.

frontend/src/components/Buttons/FilterButton.vue
Janbong: I changed this. the `library/_viewid` page expects a `Record<string, string[]>` when receiving the event.
Is there a better way to do this? Can we change the expected type of object in the `library/_viewid` page? to different `string[]`s?

ferferga (approved):

PR #1860, opened 2023-01-21
Fix persons page & date localisation

ThibaultNocchi:


frontend/src/pages/person/_itemId/index.vue
```diff
@@ -157,116 +124,96 @@ import { getBlurhash } from '@/utils/images';
...
-        this.$route.meta.title = value.Name || '';
+if (movies?.length) {
+  activeTab.value = 0;
+} else if (series?.length) {
+  activeTab.value = 1;
+} else if (books?.length) {
+  activeTab.value = 2;
+} else if (photos?.length) {
+  activeTab.value = 3;
+}
```
ferferga, line 201: Please, add a comment specifying this is for rendering the first tab that has content

frontend/src/pages/person/_itemId/index.vue
```diff
@@ -157,116 +124,96 @@ import { getBlurhash } from '@/utils/images';
...
-    if (movies?.length) {
-      activeTab = 0;
-    } else if (series?.length) {
-      activeTab = 1;
-    } else if (books?.length) {
-      activeTab = 2;
-    } else if (photos?.length) {
-      activeTab = 3;
-    }
+const activeTab = ref(4);
```
ferferga, line 191: Why 4?

frontend/src/utils/time.ts
```diff
@@ -19,8 +19,25 @@ import { usei18n } from '@/composables';
...
+    ];
+  } else if (
+    datefnslocales[
+      locale.split('-')[0] as keyof typeof import('date-fns/locale')
+    ]
+  ) {
+    return datefnslocales[
+      locale.split('-')[0] as keyof typeof import('date-fns/locale')
+    ];
+  }
```
ferferga, line 40: Why we need this?
Please also import the type at the top-level or using ES6 import syntax, don't need to duplicate the import literal this many times.

frontend/src/utils/time.ts
```diff
@@ -161,7 +178,7 @@ export function dateFnsFormat(
   return computed(() => {
     const i18n = usei18n();
 
-    args[args.length - 1] = merge(args, {
+    args[2] = merge(args[2], {
```
ferferga, line 181: Why? What if datefns change the amount of params this function has?

ferferga (changes requested):

frontend/src/pages/person/_itemId/index.vue
ThibaultNocchi: Because before my refactor it was 4. I guess that's because there is always something to display in the informations tab :p

frontend/src/utils/time.ts
ThibaultNocchi: We need this due to the differences between the vue i18n and date-fns locales.

* 1st case: when there is a date-fns locale equals to the i18n, it's all good
* 2nd case: vue i18n got locales such as `fr-FR` but date-fns got them as `frFR`. That's to match those
* 3rd case: if date-fns doesn't have the `frFR` locale, it's to check if `fr` exists

Without it, we strongly limit those matching

frontend/src/utils/time.ts
ThibaultNocchi: Cause this doesn't work. Calling this method with `dateFnsFormat(new Date(item.PremiereDate), 'PPP')` as done in the persons page, `args[args.length-1]` just gives you the `PPP` argument. Hence, without my fix, it just erased that parameter.

Also, the merge logic was wrong. It was:

```typescript
args[args.length - 1] = merge(args, {
      locale: getDateFnsLocale(i18n.locale.value)
    });
```

So, it merged the locale into `[date, format]`. I don't remember the details, but it made an infinite object.

frontend/src/utils/time.ts
ferferga: Copy and paste this into the code comments please

ferferga (approved):

PR #1861, opened 2023-01-22
Fixed webos useragent

gibahjoe:
Web OS User agent is actually with a Web0S (with a zero) for some reason. This PR fixes this.

Find reference in LG docs here 

https://webostv.developer.lge.com/develop/specifications/web-api-and-web-engine


ferferga (approved):
Thank you very much!

Can you please verify this is fixed as well in Jellyfin Web? This code is sourced from there, so probably it contains the same bug.

PR #1865, opened 2023-01-24
Fix type and sort buttons

Janbong:
This fixes the type and sort buttons at the top of the library page.

Type button functionality seems to work fine.
Sort functionality acts weird on my library. Sorting seems random.
Probably something wrong with the logic behind it.

Because we cannot use `v-list-item-group` anymore, we cannot use the index of the selected item anymore. We pass on the actual selected item when emitting the change event. For the labels in the button and list to work, some testing with a ternary operator needs to be done. Not the prettiest, I admin. If you have a tidier solution, please let me know and I will implement it.

### DIFF
--- a/frontend/src/components/Buttons/SortButton.vue
+++ b/frontend/src/components/Buttons/SortButton.vue
@@ -1,63 +1,87 @@
 <template>
-  <v-menu offset-y>
-    <template #activator="{ on, attrs }">
+  <v-btn @click="changeOrder">
+    <v-icon v-if="ascendingOrder">
+      <i-mdi-sort-ascending />
+    </v-icon>
+    <v-icon v-else>
+      <i-mdi-sort-descending />
+    </v-icon>
+  </v-btn>
+  <v-menu>
+    <template #activator="{ props: menuProps }">
       <v-btn
         v-if="!$vuetify.display.smAndDown"
         class="my-2"
         variant="text"
         rounded
-        :disabled="disabled"
-        v-bind="attrs"
-        v-on="on">
-        {{ $t('sortByType', { type: items[model].name }) }}
+        v-bind="mergeProps(menuProps, props)">
+        {{
+          $t('sortByType', {
+            type:
+              model.length === 0
+                ? items[0].title
+                : items.find((x) => x.value === model[0])?.title
+          })
+        }}
         <v-icon end>
           <i-mdi-menu-down />
         </v-icon>
       </v-btn>
-      <v-btn
-        v-else
-        :disabled="disabled"
-        class="my-2"
-        icon
-        v-bind="attrs"
-        v-on="on">
-        <v-icon>
-          <i-mdi-sort-alphabetical-ascending />
+      <v-btn v-else class="my-2" icon v-bind="mergeProps(menuProps, props)">
+        <v-icon v-if="model.length === 0 || model[0] == items[0].value">
+          <i-mdi-sort-alphabetical-variant />
+        </v-icon>
+        <v-icon v-else-if="model[0] == items[1].value">
+          <i-mdi-numeric-9-plus-box-outline />
+        </v-icon>
+        <v-icon v-else-if="model[0] == items[2].value">
+          <i-mdi-calendar-range />
         </v-icon>
       </v-btn>
     </template>
-    <v-list dense>
-      <v-list-group
-        v-model="model"
-        @change="$emit('change', items[model].value)">
-        <v-list-item v-for="item in items" :key="item.value">
-          <v-list-item-title>{{ item.name }}</v-list-item-title>
-        </v-list-item>
-      </v-list-group>
-    </v-list>
+    <v-list
+      v-model:selected="model"
+      :items="items"
+      class="filter-content"
+      @update:selected="emit('change', model[0], ascendingOrder)" />
   </v-menu>
 </template>
 
-<script lang="ts">
-import { defineComponent } from 'vue';
+<script setup lang="ts">
+import { computed, ref, defineProps, mergeProps } from 'vue';
+import { useI18n } from 'vue-i18n';
 
-export default defineComponent({
-  props: {
-    disabled: {
-      type: Boolean,
-      required: false,
-      default: false
-    }
-  },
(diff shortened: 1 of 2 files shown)


frontend/src/components/Buttons/TypeButton.vue
```diff
@@ -42,6 +35,11 @@
 import { defineComponent } from 'vue';
 
 export default defineComponent({
+  methods: {
+    log() {
+      console.log(this.model);
+    }
+  },
```
ferferga, line 42: Unnecessary

ferferga (changes requested):
Also, can you please attempt to migrate those to Vue 3's composition API?

Look at the codebase for ``<script setup>`` and compare the differences between what's on master and what we migrated.

Composition API really helps with troubleshooting type errors, so it will surely help you a lot to find errors.

One hint: maybe we don't need the computed property for ``items`` at all: Try to do it in a non-reactive way (just a simple ``const = [...]``) and change languages while testing: If the language also changes, all great! (we saved the computed overhead at runtime!)

frontend/src/components/Buttons/SortButton.vue
```diff
@@ -1,63 +1,56 @@
...
+    required: false,
+    default: false
   }
 });
+const items = [
+  { title: t('name'), value: 'SortName', order: false },
+  { title: t('rating'), value: 'CommunityRating', order: false },
+  { title: t('releaseDate'), value: 'PremiereDate', order: false }
+];
+var model = ref(0);
```
ferferga, line 55:
```suggestion
const model = ref(0);
```

This is not strictly enforced by the linter, but please let's try to stick to define all the refs and variables at the top of the script setup section

frontend/src/components/Buttons/TypeButton.vue
```diff
@@ -1,94 +1,82 @@
...
-          return [];
-        }
-      }
+  disabled: {
+    type: Boolean,
+    required: false,
+    default: false
+  }
+});
+var model = ref(0);
```
ferferga, line 50: Same as above with the const, refs never can be reassigned (or reactivity will be lost along the way)

ferferga (changes requested):

frontend/src/components/Buttons/TypeButton.vue
```diff
@@ -1,94 +1,84 @@
...
+    required: false,
+    default: false
+  }
+});
+const emit = defineEmits<{
+  (e: 'change', types: Record<string, string[]>): void;
+}>();
+var model = ref(0);
+
+const items = computed((): Array<Record<string, string>> => {
```
ferferga, line 54: Type inference is automatic, so this is redundant. However, in case TS can't infer the types properly, it's this way:
```suggestion
const items = computed<Array<Record<string, string>>>(()  => {
```

ferferga (changes requested):

frontend/src/components/Buttons/SortButton.vue
```diff
@@ -1,63 +1,82 @@
...
+const { t } = useI18n();
+const props = defineProps({
+  disabled: {
+    type: Boolean,
+    required: false,
+    default: false
   }
 });
+const emit = defineEmits<{
+  (e: 'change', types: Record<string, string[]>, sortOrder: boolean): void;
```
ferferga, line 64:
```suggestion
  (e: 'change', types: Record<string, string[]>, ascendingOrder: boolean): void;
```

frontend/src/components/Buttons/TypeButton.vue
```diff
@@ -1,94 +1,84 @@
...
-        @change="$emit('change', items[model].value)">
-        <v-list-item v-for="item in items" :key="item.value">
-          <v-list-item-title>{{ item.name }}</v-list-item-title>
-        </v-list-item>
-      </v-list-group>
-    </v-list>
+    <v-list
+      v-model:selected="model"
+      :items="items"
+      @update:selected="emit('change', model[0])" />
```
ferferga, line 29: I think this is wrong. Model is not an array according to what you have defined.

frontend/src/components/Buttons/SortButton.vue
```diff
@@ -1,63 +1,82 @@
...
+}>();
+const model = ref(0);
+const ascendingOrder = ref(true);
+const items = computed(() => [
+  { title: t('name'), value: 'SortName' },
+  { title: t('rating'), value: 'CommunityRating' },
+  { title: t('releaseDate'), value: 'PremiereDate' }
+]);
+
+function changeOrder() {
```
ferferga, line 74: This function needs a comment according to our linting rules

frontend/src/components/Buttons/SortButton.vue
```diff
@@ -1,63 +1,82 @@
...
-        <v-list-item v-for="item in items" :key="item.value">
-          <v-list-item-title>{{ item.name }}</v-list-item-title>
-        </v-list-item>
-      </v-list-group>
-    </v-list>
+    <v-list
+      v-model:selected="model"
+      :items="items"
+      class="filter-content"
+      @update:selected="emit('change', model[0], descendingOrder)" />
```
ferferga, line 47: As below, this seems wrong

ferferga (changes requested):

frontend/src/components/Buttons/TypeButton.vue
Janbong: Because the items get loaded in the `v-list` using `:items="items"`, the selected item gets put in an array, when passing it to the parent component, we need to take the first object in that array. Not sure why it does that, but without the `[0]` behind it, the parent component does not recognize the selected item. Might have something to do with https://next.vuetifyjs.com/en/api/v-list/#props-select-strategy . But it is not documented by Vuetify at all.

frontend/src/components/Buttons/SortButton.vue
Janbong: see https://github.com/jellyfin/jellyfin-vue/pull/1865#discussion_r1087739438

ferferga (approved):

PR #1866, opened 2023-01-25
feat: first pass of playback refactoring

ferferga:
Doesn't work in development because https://github.com/vuejs/core/issues/7567

* This PR lays the groundwork and adds some hints to some files, with ideas of how I see those features reimplemented for playback.
* Only music works at this point, but without AudioContext or any fancyness.


frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -0,0 +1,97 @@
...
+const mediaElement = computed<'audio' | 'video' | undefined>(() => {
+  if (playbackManager.currentlyPlayingMediaType === 'Audio') {
+    return 'audio';
+  } else if (playbackManager.currentlyPlayingMediaType === 'Video') {
+    return 'video';
+  }
+});
+
+/**
+ * If the player is a video element and we're
```
ThibaultNocchi, line 49: And we're what? :p

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -0,0 +1,97 @@
...
+  return mediaElement.value === 'audio' ? 'body' : '.video-container';
+});
+
+const posterUrl = computed<string>(() => {
+  return !isNil(playbackManager.currentItem) && mediaElement.value === 'video'
+    ? getImageInfo(playbackManager.currentItem, {
+        preferBackdrop: true
+      }).url || ''
+    : '';
+});
```
ThibaultNocchi, line 61:
```suggestion
const posterUrl = computed<string>(() =>
  !isNil(playbackManager.currentItem) && mediaElement.value === 'video'
    ? getImageInfo(playbackManager.currentItem, {
        preferBackdrop: true
      }).url || ''
    : ''
);
```

frontend/src/store/items.ts
```diff
@@ -19,12 +20,62 @@ const defaultState: ItemsState = {
...
+  /**
+   * == GETTERS ==
+   */
+  public getItemById = (id: string | undefined): BaseItemDto | undefined => {
+    return computed(() => {
+      if (id) {
+        return state.value.byId[id];
+      }
+    }).value;
+  };
```
ThibaultNocchi, line 38:
```suggestion
  public getItemById = (id: string | undefined): BaseItemDto | undefined =>
    computed(() => (id ? state.value.byId[id] : undefined)).value;
```

Wdyt of this? Same for others :)

frontend/src/store/playbackManager.ts
```diff
@@ -101,88 +111,89 @@ const defaultState: PlaybackManagerState = {
...
+const mediaMetadata = new MediaMetadata();
 
 class PlaybackManagerStore {
   /**
    * == GETTERS ==
    */
-  public get status(): typeof state.status {
+  public get status(): PlaybackStatus {
     return state.status;
   }
```
ThibaultNocchi, line 131: I don't remember, can getters be arrows functions?

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -0,0 +1,97 @@
...
+
+const isLoopingOnce = computed(
+  () => playbackManager.repeatMode === RepeatMode.RepeatOne
+);
+
+watch(mediaElement, async () => {
```
ThibaultNocchi, line 67: Shouldn't this be immediate?

frontend/locales/en-US.json
```diff
@@ -69,7 +69,8 @@
       "errorCode": "Error code: {errorCode}",
       "videoPlayerError": "The video player encountered an unrecoverable error."
     },
-    "unauthorized": "You're not authorized to access this page"
+    "unauthorized": "You're not authorized to access this page",
+    "cantPlayItem": "This item can't be played"
```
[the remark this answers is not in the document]
ferferga: I've been toying with the same idea for a while (check sort-keys eslint rule, and even found its automatic fix at https://www.npmjs.com/package/eslint-plugin-sort-keys-fix)

However, it was a pain to deal with in some places, like Vite config. 

We could do this for JSON locales only maybe.

frontend/src/components/Layout/AudioControls.vue
```diff
@@ -28,25 +28,32 @@
...
                     <router-link
-                      tag="span"
-                      class="text--secondary text-caption text-truncate link"
-                      :to="getItemDetailsLink(artist, 'MusicArtist')">
-                      {{ artist.Name }}
+                      v-slot="{ navigate }"
+                      :to="getItemDetailsLink(artist, 'MusicArtist')"
+                      custom>
+                      <span
+                        class="text--secondary text-caption text-truncate link"
```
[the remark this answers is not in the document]
ferferga: I didn't change that, it's probably an automatic eslint fix. You're the Vuetify style expert right now, I didn't take too much time on it :)

frontend/src/components/Layout/TimeSlider.vue
```diff
@@ -1,85 +1,51 @@
...
-.time-slider .v-slider__track-background,
-.time-slider .v-slider__track-fill {
-  transition: none !important;
+/**
+ * onMouseUp event handler
+ *
+ * Once the user releases the slider, change the time of the playbackManager with whatever
+ * input value was provided by the user
+ */
+function releaseMouse(): void {
```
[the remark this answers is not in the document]
ferferga: It's indicated in the comments though. Which name do you suggest?

frontend/src/store/items.ts
ferferga: If reactivity works, I'm fine

frontend/src/store/playbackManager.ts
ferferga: It can't, at leat it gave me syntax errors.

frontend/src/components/Layout/TimeSlider.vue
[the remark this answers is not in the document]
ThibaultNocchi: I don't know, something more related to what its role is, not its event.

frontend/src/components/Layout/VolumeSlider.vue
```diff
@@ -28,6 +30,16 @@ import IMdiVolumeLow from 'virtual:icons/mdi/volume-low';
 import { playbackManagerStore } from '@/store';
 
 const playbackManager = playbackManagerStore();
+
+const sliderValue = computed({
+  get() {
+    return playbackManager.currentVolume;
```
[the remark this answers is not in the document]
ThibaultNocchi: Before your commit, the slider had this prop: `:model-value="playbackManager.isMuted ? 0 : playbackManager.currentVolume"`. Is it intended you dropped the fact  that the slider is (virtually) at 0 when muted?

frontend/src/components/Playback/PlayerElement.vue
ThibaultNocchi:
```
/**
 * If the player is a video element and we're
 */
const teleportTarget = computed<'body' | '.video-container'>(() => {
  return mediaElement.value === 'audio' ? 'body' : '.video-container';
});
```

if that can help you :)

frontend/src/store/items.ts
ThibaultNocchi: Ok so it seems it has tradeoffs: `This will use more memory, because each class instance will have its own copy of each function defined this way`. So we'll keep it this way for the getter, I'll just change the computed return value

frontend/src/store/playbackManager.ts
ThibaultNocchi: Np :)

frontend/locales/en-US.json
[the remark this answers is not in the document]
ThibaultNocchi: I did it manually for now. For objects in the code I don't really mind. For the locales there are so many so I think it's interesting.

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -0,0 +1,97 @@
...
+import { playbackManagerStore, RepeatMode } from '@/store';
+import { mediaElementRef } from '@/store/playbackManager';
+import { getImageInfo } from '@/utils/images';
+
+const playbackManager = playbackManagerStore();
+let isShakaAttached = false;
+const shaka = new shakaPlayer.Player();
+
+window.muxjs = muxjs;
+window.player = shaka;
```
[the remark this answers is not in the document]
ThibaultNocchi: Will drop Shaka so I don't mind

frontend/src/store/items.ts
ferferga: We only have one instance on the app though.

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -64,33 +52,19 @@ const isLoopingOnce = computed(
...
-watch(
-  () => playbackManager.currentItemIndex,
-  async () => {
-    await shaka.unload();
-  }
-);
-
 watch(
   () => playbackManager.currentSourceUrl,
   async () => {
```
ferferga, line 57: @ThibaultNocchi This is no longer async

frontend/src/components/Playback/PlayerElement.vue
ferferga: @ThibaultNocchi IIRC I wanted to continue with something like this:

``If the player is a video element and we're in the PiP player or fullscreen video playback, we need to ensure the DOM elements are mounted before the teleport target is ready``

PR #1869, opened 2023-01-27
style: remove unnecessary `as` casts

aweebs:
Just starting to play around in this codebase, and realized there were a few places where `as` casts were being used but the type could be correctly specified and checked instead.

### DIFF
--- a/frontend/components/Buttons/__tests__/MarkPlayedButton.spec.ts
+++ b/frontend/components/Buttons/__tests__/MarkPlayedButton.spec.ts
@@ -1,4 +1,3 @@
-import { BaseItemDto } from '@jellyfin/client-axios';
 import { createLocalVue, mount, Wrapper } from '@vue/test-utils';
 import Vuetify from 'vuetify';
 import Vue from 'vue';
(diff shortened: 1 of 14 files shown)


ferferga:
@aweebs master is currently in maintenance mode. Do you mind checking ``vite`` branch an retarget your PR there, please?

We have much better type checking there (thanks to the improved TS support of Vue 3) and I recall removing a lot of useless casts as well, but a double-check doesn't hurt. Linting and type fixes in that branch are also welcome.

Additionally, if you can make sure we have some linter rule to avoid adding them, even better (I think we do, but better when more than two eyes 👀 are in)

aweebs:
@ferferga ah, thanks for letting me know about this branch being in maintenance. I've switched over to start looking at the `vite` branch, I'll have a PR up targeting that branch soon, and can look into linting options to keep things clean.

In the meantime, this PR is 0 regression risk and makes things slightly better, should we merge it in any way?

aweebs:
See #1874 for PR to `vite`

ferferga:
@aweebs I'd rather not, because that means spending extra time fixing merge conflicts for a really small benefit (benefits we will have in a much better way in vite branch)

ferferga:
Closing in favor of #1874

PR #1870, opened 2023-01-27
fix: SeasonTabs component

Janbong:
This fixed the SeasonTabs component that shows episodes per season in the `series/_viewid` page.

### DIFF
--- a/frontend/src/components/Item/SeasonTabs.vue
+++ b/frontend/src/components/Item/SeasonTabs.vue
@@ -5,34 +5,41 @@
         {{ season.Name }}
       </v-tab>
     </v-tabs>
-    <v-tabs v-model="currentTab" class="bg-transparent">
-      <v-tab v-for="season in seasons" :key="season.Id">
+    <v-window v-model="currentTab" class="bg-transparent">
+      <v-window-item v-for="season in seasons" :key="season.Id">
         <v-list
           v-if="seasonEpisodes && season.Id"
-          lines="two"
-          color="transparent">
+          :lines="false"
+          bg-color="transparent">
           <v-list-item
             v-for="episode in seasonEpisodes[season.Id]"
             :key="episode.Id"
-            :to="getItemDetailsLink(episode)"
-            class="flex-column flex-md-row">
-            <v-avatar width="20em" height="12em">
-              <blurhash-image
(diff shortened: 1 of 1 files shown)


frontend/src/components/Item/SeasonTabs.vue
```diff
@@ -5,34 +5,42 @@
...
     </v-tabs>
-    <v-tabs v-model="currentTab" class="bg-transparent">
-      <v-tab v-for="season in seasons" :key="season.Id">
+    <v-window v-model="currentTab" class="bg-transparent">
+      <v-window-item v-for="season in seasons" :key="season.Id">
         <v-list
           v-if="seasonEpisodes && season.Id"
-          lines="two"
-          color="transparent">
+          lines="false"
```
ThibaultNocchi, line 12:
```suggestion
          :lines="false"
```
This prop accepts the boolean false, not the string false ;)
Also, is this needed at all?

frontend/src/components/Item/SeasonTabs.vue
```diff
@@ -5,34 +5,42 @@
...
+            :to="getItemDetailsLink(episode)">
+            <v-row class="mb-8">
+              <v-col
+                :class="{ 'py-1': $vuetify.display.smAndDown }"
+                cols="12"
+                md="4">
+                <card
+                  :class="{ 'mx-8 mt-8': $vuetify.display.smAndDown }"
+                  :item="episode" />
+                <watched-indicator v-if="episode.UserData?.Played" />
```
ThibaultNocchi, line 26: Not needed anymore as the indicator is added by the card, and this just adds the indicator twice on the right of the item
```suggestion
```

ThibaultNocchi (commented):
Do you think the text could be vertically centered? I think you could use an `align` prop on the v-row

frontend/src/components/Item/SeasonTabs.vue
Janbong: Jeezz, I couldn't figure out why it kept complaining. thanks

It is needed because otherwise it puts the subtitle (episode overview in this case) on one line and shortens the rest with "..."

ThibaultNocchi: Oh great then! And no problem, once you know that happens you'll see it more quickly next time :)

PR #1874, opened 2023-01-28
style: remove unnecessary as casts

aweebs:
Removes unnecessary `as` casts, along with a few other typing updates. Mostly low-hanging fruit.
See #1869 for original effort on master branch.

Some of my own notes from going through `as` casts in code base:
1. `as` casts are being used to type component data structures, rather than using interfaces. I'll look into converting these next.
2. `as` casts are being used as not-null assertions fairly often. This should either be changed to actual not-null-assertions operators (`!`), which are still not ideal and are currently sent to error with linting, or actually have null checks performed. This is difficult because of the DTO types from the api.
3. At the moment, there are likely too many `as` casts for a linting rule to be useful, but we should consider adding them as these narrow down.

### DIFF
--- a/frontend/src/components/Item/ItemMenu.vue
+++ b/frontend/src/components/Item/ItemMenu.vue
@@ -57,7 +57,7 @@ import IMdiRefresh from 'virtual:icons/mdi/refresh';
 import IMdiDotsHorizontal from 'virtual:icons/mdi/dots-horizontal';
 import { useRemote, useSnackbar } from '@/composables';
 import { canResume } from '@/utils/items';
-import { TaskType, RunningTask } from '@/store/taskManager';
+import { TaskType } from '@/store/taskManager';
 import { playbackManagerStore, taskManagerStore } from '@/store';
 
 type MenuOption = {
@@ -119,7 +119,7 @@ const playNextAction = {
  * Options to show when the item menu is invoked in a queue item
  */
 function getQueueOptions(): MenuOption[] {
-  const queueOptions = [] as MenuOption[];
+  const queueOptions: MenuOption[] = [];
 
   if (
     menuProps.queue &&
@@ -169,7 +169,7 @@ function getQueueOptions(): MenuOption[] {
  * Playback options for the items
  */
 function getPlaybackOptions(): MenuOption[] {
-  const playbackOptions = [] as MenuOption[];
+  const playbackOptions: MenuOption[] = [];
 
   if (canResume(menuProps.item)) {
     playbackOptions.push({
@@ -221,7 +221,7 @@ function getPlaybackOptions(): MenuOption[] {
  * Library options for libraries
  */
 function getLibraryOptions(): MenuOption[] {
-  const libraryOptions = [] as MenuOption[];
+  const libraryOptions: MenuOption[] = [];
 
   if (
     remote.auth.currentUser?.Policy?.IsAdministrator &&
@@ -244,10 +244,10 @@ function getLibraryOptions(): MenuOption[] {
             useSnackbar(t('libraryRefreshQueued'), 'normal');
             taskManager.startTask({
               type: TaskType.LibraryRefresh,
-              id: menuProps.item.Id,
-              data: menuProps.item.Name,
+              id: menuProps.item.Id || '',
+              data: menuProps.item.Name || '',
               progress: 0
-            } as RunningTask);
+            });
           } catch (error) {
             console.error(error);
 
--- a/frontend/src/components/Item/Metadata/ImageEditor.vue
+++ b/frontend/src/components/Item/Metadata/ImageEditor.vue
@@ -94,7 +94,7 @@ export default defineComponent({
   },
   computed: {
     generalImages(): boolean {
-      return this.images.filter((image: ImageInfo) => {
+      return this.images.filter((image) => {
         return (
           image.ImageType !== ImageType.Screenshot &&
           image.ImageType !== ImageType.Backdrop &&
--- a/frontend/src/components/Layout/VirtualGrid/pipeline.ts
+++ b/frontend/src/components/Layout/VirtualGrid/pipeline.ts
@@ -111,8 +111,8 @@ export function fromScrollParent(
   return computed(() => {
     const el = unrefElement(elRef);
 
-    if (el) {
-      const { vertical, horizontal } = getScrollParents(el as HTMLElement);
+    if (el && el instanceof HTMLElement) {
+      const { vertical, horizontal } = getScrollParents(el);
 
       /**
        * If the scrolling parent is the doc root, use window instead as using
@@ -320,10 +320,7 @@ export function getVisibleItems<T>(
 
   return allItems.slice(first, last).map((value, localIndex) => {
     const index = first + localIndex;
-    const { x, y } = getItemOffsetByIndex(
-      index,
-      resizeMeasurement
-    ) as ItemOffset;
+    const { x, y } = getItemOffsetByIndex(index, resizeMeasurement);
 
     return {
       index,
--- a/frontend/src/pages/index.vue
+++ b/frontend/src/pages/index.vue
@@ -66,31 +66,31 @@ if (userLibraries.isReady) {
 const homeSections = computed<HomeSection[]>(() => {
   // Filter for valid sections in Jellyfin Vue
   // TODO: Implement custom section order
-  let homeSectionsArray = pickBy(
+  const homeSectionPrefs = pickBy(
     // @ts-expect-error - No typings for this
-    clientSettings.CustomPrefs,
-    (value: string, key: string) => {
+    clientSettings.CustomPrefs as { [key: string]: string },
+    (value, key) => {
       return (
         value && VALID_SECTIONS.has(value) && key.startsWith('homesection')
       );
     }
   );
 
-  if (Object.keys(homeSectionsArray).length === 0) {
-    homeSectionsArray = {
-      homeSection0: 'librarytiles',
-      homeSection1: 'resume',
-      homeSection2: 'resumeaudio',
-      homeSection3: 'upnext',
-      homeSection4: 'latestmedia'
-    };
-  }
+  const homeSectionKeys = Object.keys(homeSectionPrefs);
 
-  homeSectionsArray = Object.values(homeSectionsArray);
+  if (homeSectionKeys.length === 0) {
+    homeSectionKeys.push(
+      'librarytiles',
+      'resume',
+      'resumeaudio',
+      'upnext',
+      'latestmedia'
+    );
+  }
 
   const homeSections: HomeSection[] = [];
 
-  for (const homeSection of homeSectionsArray as Array<string>) {
+  for (const homeSection of homeSectionKeys) {
     switch (homeSection) {
       case 'librarytiles': {
         homeSections.push({
@@ -113,7 +113,10 @@ const homeSections = computed<HomeSection[]>(() => {
           ]);
 
           for (const userView of userLibraries.libraries) {
-            if (excludeViewTypes.has(userView.CollectionType as string)) {
+            if (
+              userView.CollectionType &&
+              excludeViewTypes.has(userView.CollectionType)
+            ) {
               continue;
             }
 
--- a/frontend/src/pages/item/_itemId/index.vue
+++ b/frontend/src/pages/item/_itemId/index.vue
@@ -272,7 +272,8 @@ export default defineComponent({
   computed: {
     isPlayable(): boolean {
       // TODO: Move this to a mixin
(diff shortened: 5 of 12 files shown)


frontend/src/components/Item/ItemMenu.vue
```diff
@@ -244,10 +244,10 @@ function getLibraryOptions(): MenuOption[] {
             useSnackbar(t('libraryRefreshQueued'), 'normal');
             taskManager.startTask({
               type: TaskType.LibraryRefresh,
-              id: menuProps.item.Id,
-              data: menuProps.item.Name,
+              id: menuProps.item.Id as string,
+              data: menuProps.item.Name as string,
```
aweebs, line 248: These casts are really just not-null assertions. I figured that was better than casting the entire object below. These can be more easily cleared out later I think.

frontend/src/components/Item/Metadata/ImageEditor.vue
```diff
@@ -94,7 +94,7 @@ export default defineComponent({
   },
   computed: {
     generalImages(): boolean {
-      return this.images.filter((image: ImageInfo) => {
+      return this.images.filter((image) => {
```
aweebs, line 97: Implicit parameter typing, wasn't necessary

frontend/src/store/playbackManager.ts
```diff
@@ -270,13 +270,13 @@ class PlaybackManagerStore {
           label: sub.el.DisplayTitle || 'Undefined',
           src:
             sub.el.DeliveryMethod === SubtitleDeliveryMethod.External
-              ? sub.el.DeliveryUrl
+              ? sub.el.DeliveryUrl || undefined
```
aweebs, line 273: Adding the `||` undefined isn't great, but I definitely think it's better than casting the entire object below.

frontend/src/utils/store-sync.ts
```diff
@@ -47,7 +47,7 @@ export async function fetchSettingsFromServer<T>(
   key: string,
   target: T,
   cast = true
-): Promise<T | DisplayPreferencesDto> {
+): Promise<DisplayPreferencesDto> {
```
aweebs, line 50: After removing all the typecasts, this function always return `DisplayPreferencesDto`, so it just worked out nicely

frontend/src/pages/index.vue
```diff
@@ -66,31 +66,31 @@ if (userLibraries.isReady) {
 const homeSections = computed<HomeSection[]>(() => {
   // Filter for valid sections in Jellyfin Vue
   // TODO: Implement custom section order
-  let homeSectionsArray = pickBy(
+  const homeSectionPrefs = pickBy(
```
aweebs, line 69: Previously homeSectionsArray was untyped and started off as an object, then got re-assigned to an array later, so it's type was actually changing. The logic is identical, but I think it's just easier to follow now.

frontend/src/pages/index.vue
```diff
@@ -66,31 +66,31 @@ if (userLibraries.isReady) {
...
     // @ts-expect-error - No typings for this
-    clientSettings.CustomPrefs,
-    (value: string, key: string) => {
+    clientSettings.CustomPrefs as { [key: string]: string },
```
aweebs, line 71: This does add an `as` cast, but I think it makes sense here, since we do not have typings for `CustomPrefs` and the cast is used to indicate to the reader what the value is expected to be.

aweebs (commented):
Adding some notes for reviewers

aweebs:
See #1874 for PR to `vite`

frontend/src/components/Layout/VirtualGrid/pipeline.ts
```diff
@@ -73,9 +73,9 @@ export function getScrollParents(
     ? /(auto|scroll|hidden)/
     : /(auto|scroll)/;
 
-  let vertical: HTMLElement | undefined;
-  let horizontal: HTMLElement | undefined;
-  let parent: HTMLElement | null = element;
+  let vertical: HTMLElement | SVGElement | undefined;
+  let horizontal: HTMLElement | SVGElement | undefined;
+  let parent: HTMLElement | SVGElement | null = element;
```
ferferga, line 78: Why?

frontend/src/components/Layout/VirtualGrid/pipeline.ts
```diff
@@ -57,7 +57,7 @@ export interface InternalItem<T> {
  * Gets the appropiate scroll parent for the virtual grid DOM element
  */
 export function getScrollParents(
-  element: HTMLElement,
+  element: HTMLElement | SVGElement,
```
ferferga, line 60: Why?

frontend/src/components/Item/ItemMenu.vue
ferferga, line 248:
```suggestion
              id: menuProps.item.Id || '',
              data: menuProps.item.Name || '',
```

This is what we're doing in other places, and let us be sure this is really 100% safe at runtime.

frontend/src/store/playbackManager.ts
ferferga: Why you think it's better?

frontend/src/utils/store-sync.ts
ferferga: If data is casted, definitely not ``DisplayPreferencesDto`` is returned here

ferferga (changes requested):

ferferga:
Sorry, for the force push, just a commit was renamed, no need to rewrite any of your own history

aweebs:
Reverted the store-sync changes since one of the functions had a bad return type which was hidden. I'll dig on that later. I think what's here now is a good small start.

frontend/src/utils/store-sync.ts
ferferga: @aweebs No, the typings are correct. However, when we do the cast, we're turning them to native types.

[Server response (``DisplayPreferencesDto``)](https://api.jellyfin.org/#tag/DisplayPreferences):

```
{
  ...whatever,
  CustomPrefs: {
    hello: "0",
    test: "null"
  }
}
```

When cast is true we turn that into:
```
{
  ...whatever,
  CustomPrefs: {
    hello: 0,
    test: null
  }
}
```

Effectively matching the store's native types.
Do note the difference between TypeScript casting and runtime castings. Here we're doing a runtime casting.

frontend/src/pages/index.vue
ferferga: No worry too much about this, this whole section needs refactoring anyway (it doesn't work at all), so it's good!

frontend/src/components/Layout/VirtualGrid/pipeline.ts
ferferga: Honestly, I don't think this makes sense just for the sake of removing ``as``s. TypeScript is a way to assist in the type inference of stuff in the JavaScript world, and ``as``, although not ideal, is indeed a hint to help it in some occasions do the correct inference. Sometimes even ``any`` is necessary to do more complex inferences like [here](https://github.com/jellyfin/jellyfin-vue/blob/vite/frontend/src/composables/use-datefns.ts#L38) (and I looked a lot of libraries to see how similar typings were accomplished, with all of them using ``any``).

At no point is logical to find a CSS virtual grid inside an svg. It's silly to assume we could be dealing with this at runtime, because it's illogical to do so. And this just adds complexity (dealing with SVGElement typings) just for the sake of making the "linter" happy, in my opinion.

@ThibaultNocchi any opinions?

frontend/src/components/Layout/VirtualGrid/pipeline.ts
```diff
@@ -320,10 +320,7 @@ export function getVisibleItems<T>(
 
   return allItems.slice(first, last).map((value, localIndex) => {
     const index = first + localIndex;
-    const { x, y } = getItemOffsetByIndex(
-      index,
-      resizeMeasurement
-    ) as ItemOffset;
+    const { x, y } = getItemOffsetByIndex(index, resizeMeasurement);
```
ferferga, line 323: @aweebs Good catch!

Do you have any proposal for any linter rules that helps us avoid this kind of missteps in the future.
Essentially every rule that could avoid us having another stylistic PR like this one is welcome

frontend/src/pages/item/_itemId/index.vue
```diff
@@ -272,7 +272,8 @@ export default defineComponent({
   computed: {
     isPlayable(): boolean {
       // TODO: Move this to a mixin
-      return ['PhotoAlbum', 'Photo', 'Book'].includes(this.item.Type as string)
+      return this.item.Type &&
+        ['PhotoAlbum', 'Photo', 'Book'].includes(this.item.Type)
```
ferferga, line 276: We have ``BaseItemKind`` now

frontend/src/store/playbackManager.ts
```diff
@@ -447,7 +449,7 @@ class PlaybackManagerStore {
         seekto: (action): void => {
           this.currentTime = action.seekTime || 1;
         }
-      } as { [key in MediaSessionAction]?: MediaSessionActionHandler };
+      };
```
ferferga, line 452: As said above, any linter rule to have this enforced by default is more than appreciated!

frontend/src/components/Buttons/__tests__/MarkPlayedButton.spec.ts
```diff
@@ -1,4 +1,3 @@
-import { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
```
ferferga, line 1: Ignore this whole file for now and skip it from this commit please. We need to rewrite tests completely for vitest, they're here so they could be easily traced without git bisecting when the test suite rewrite is done (after ``vite`` branch gets merged to master).

Same applies to all the spec files, so if you had another PR in mind for all of those, ignore them.
PRs refactoring the testing suite to use vitets are welcome of course :).

frontend/src/components/Layout/VirtualGrid/pipeline.ts
aweebs: The other option would be to do add an `&& el instanceof HTMLElement` to the `if` statement on line 114. This would accomplish the same thing by only allowing an HTML element into this function. Happy to make that change if you think it's better.

I do think `as` casts have a use, but they should be used as an escape hatch rather than a convenience. I don't think that applies here since it's pretty simple to narrow the types before calling the function or expand the types of the function parameter to avoid needing one.

frontend/src/utils/store-sync.ts
aweebs: The current return value of `castDisplayPreferencesResponse` is typed as `DisplayPreferencesDto`, but the return value in the function is `data.CustomPrefs || {}` on line 37. Currently this function will pick certain properties off `DisplayPreferencesDto#CustomPrefs` and return that object, which is explicitly not `DisplayPreferencesDto`. In practice it does return type `T`. The reason there is no type error right now is because every property of `DisplayPreferencesDto` is optional which allows any object type to successfully check against it.

frontend/src/pages/item/_itemId/index.vue
aweebs: Apologies, I'm not sure what you mean by this comment. This was changed because `Type` could be undefined so it first null checks and then does the array check.

frontend/src/store/playbackManager.ts
aweebs: This specifically only defaults the value for the src property on `PlaybackTrack`. If you mean we should change the function to return `PlaybackTrack[]` instead of `PlaybackTrack[] | undefined` by always return an empty list as default, I'm happy to make that change too.

frontend/src/components/Layout/VirtualGrid/pipeline.ts
ferferga: @aweebs Sounds good to me!

frontend/src/pages/item/_itemId/index.vue
ferferga: @aweebs Basically this, but now I'm realising it doesn't solve the problem:

``[BaseItemKind.PhotoAlbum, BaseItemKind.Photo, BaseItemKind.Photo].includes...``

ferferga (approved):
LGTM! Rebase your commit to include my last requested change in the VirtualGrid's pipeline, and it's good to go!

Thank you very much for your contribution!

ferferga:
@aweebs Any update on the ``instanceof`` change? :)

aweebs:
> @aweebs Any update on the `instanceof` change? :)

Sorry, I was away for the weekend. I've rebased and pushed the `instanceof` changes. Feel free to have a final peek and then merge away!

ferferga (approved):
@aweebs Thank you very much for your PR! 😊

PR #1875, opened 2023-01-30
Fix: search

Janbong:


frontend/src/pages/search.vue
```diff
@@ -15,146 +15,145 @@
...
+    bookSearchResults.value = itemResults.filter(
+      (item: BaseItemDto) => item.Type === 'Book'
+    );
+
+    personSearchResults.value = (
+      await remote.sdk.newUserApi(getPersonsApi).getPersons({
+        userId: remote.auth.currentUserId.value,
+        searchTerm: searchQuery.value
+      })
+    ).data.Items;
```
ThibaultNocchi, line 148:
```suggestion
    ).data.Items || [];
```
Should do the trick :)

frontend/src/pages/search.vue
```diff
@@ -15,146 +15,145 @@
...
+    albumSearchResults.value = itemResults.filter(
+      (item: BaseItemDto) => item.Type === 'MusicAlbum'
+    );
+
+    trackSearchResults.value = itemResults.filter(
+      (item: BaseItemDto) => item.Type === 'Audio'
+    );
+
+    bookSearchResults.value = itemResults.filter(
+      (item: BaseItemDto) => item.Type === 'Book'
```
ThibaultNocchi, line 140: You can use the enum for the type checks, such as this:

```suggestion
      (item: BaseItemDto) => item.Type === BaseItemKind.Book
```

ferferga (changes requested):
* We should remove the ``Top results`` tab, because the Virtual Grid is not ready to show cards of different sizes (or we shouldn't use virtual grid at all here, which may even be a more sensible solution if we want to keep that tab, which I'm not a fan of because it's horrible). We need to redesign that page functionally anyway.

* I'd say we should keep the functionality of keeping tabs that don't have any items disabled.

ferferga (approved):

frontend/src/pages/search.vue
```diff
@@ -1,160 +1,189 @@
...
+);
+
+watch(
+  searchQuery,
+  async () => {
+    if (searchQuery.value !== '') {
+      await performSearchDebounce();
     } else {
-      await this.performSearch();
+      router.back();
```
ThibaultNocchi, line 120: What do you think of putting the router.back in the performSearch? That way 1. logic is centralized 2. you can quickly erase your sentence and start typing again without being redirected back to the debounced logic.

Also you may want to use watchDebounced from vueuse?

ThibaultNocchi (approved):

frontend/src/components/Layout/AppBar/SearchField.vue
```diff
@@ -22,13 +23,23 @@ const searchQuery = computed({
...
-      router.back();
-    } else if (searchQuery.value) {
-      router.replace({ path: '/search', query: { q: value } });
-    } else {
-      router.push({ path: '/search', query: { q: value } });
-    }
+    router.replace({
+      ...router.currentRoute,
+      query: {
+        q: value || undefined
```
ThibaultNocchi, line 29:
```suggestion
        q: value.trim() || undefined
```

Just an easy QOL change

PR #1876, opened 2023-01-30
style: switch button to variant text

ThibaultNocchi:
Use `variant="text"` by default on buttons to match master style

### DIFF
--- a/frontend/src/components/Buttons/ScrollToTopButton.vue
+++ b/frontend/src/components/Buttons/ScrollToTopButton.vue
@@ -2,11 +2,12 @@
   <v-fab-transition>
     <v-btn
       v-show="y > 1"
+      class="ma-4"
       color="primary"
-      fab
-      dark
+      icon
       location="bottom right"
-      fixed
+      position="fixed"
+      variant="elevated"
       @click="scrollToTop">
       <v-icon>
         <i-mdi-chevron-up />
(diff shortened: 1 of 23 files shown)


ferferga (changes requested):
Didn't test it any further, as this breaks server connect styles already:

image

ferferga:
@ThibaultNocchi We want the play buttons inside cards to stay the same way? Not full circular like in master?

frontend/src/components/Buttons/ScrollToTopButton.vue
```diff
@@ -2,11 +2,11 @@
   <v-fab-transition>
```
ferferga, line 2: This button should be rounded, like the play one (if we want to strictly stick with master styles)

frontend/src/components/Buttons/SortButton.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-btn @click="changeOrder">
+  <v-btn icon @click="changeOrder">
```
ferferga, line 2:
```suggestion
  <v-btn icon @click="changeOrder" v-bind="props">
```

This would fix the sort button not being disabled on page load, like in master. Not strictly related to this PR but a good to have fix anyway.

ferferga (changes requested):

ferferga (approved):

PR #1878, opened 2023-01-31
Implement video player in Vite branch

ThibaultNocchi:
- [x] Remove unneeded logs
- [x] Remove unneeded CSS styles and classes
- [x] Skim over the code with a clear mind

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -26,6 +26,7 @@
     "vue-i18n-extract-remove": "vue-i18n-extract --remove"
   },
   "dependencies": {
+    "@jellyfin/libass-wasm": "4.1.1",
     "@jellyfin/sdk": "0.0.0-unstable.202301101857",
     "@vueuse/components": "9.12.0",
     "@vueuse/core": "9.12.0",
@@ -36,15 +37,15 @@
     "destr": "1.1.1",
     "dompurify": "2.3.8",
     "he": "1.2.0",
+    "hls.js": "1.3.1",
     "langs": "2.0.0",
     "lodash-es": "4.17.21",
-    "screenfull": "6.0.1",
     "swiper": "9.0.3",
     "uuid": "9.0.0",
     "vue": "3.2.39",
     "vue-i18n": "9.2.2",
     "vue-router": "4.1.5",
-    "vuetify": "3.1.2"
+    "vuetify": "3.1.6"
   },
   "devDependencies": {
     "@iconify/json": "2.1.104",
@@ -85,6 +86,7 @@
     "vite": "3.2.2",
     "vite-plugin-pages": "0.27.1",
     "vite-plugin-pwa": "0.13.2",
+    "vite-plugin-static-copy": "0.13.0",
     "vite-plugin-vue-layouts": "0.7.0",
     "vue-eslint-parser": "9.1.0",
     "vue-i18n-extract": "2.0.7",
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -3,6 +3,6 @@
   <v-app>
     <router-view-transition is-root />
     <snackbar />
-    <player-element />
   </v-app>
+  <player-element />
 </template>
--- a/frontend/src/components/Buttons/PlaybackSettingsButton.vue
+++ b/frontend/src/components/Buttons/PlaybackSettingsButton.vue
@@ -1,143 +1,100 @@
 <template>
   <v-menu
-    v-model="menu"
+    v-model="menuModel"
     :close-on-content-click="false"
-    :persistent="!true"
     :transition="'slide-y-transition'"
-    location="top"
-    :nudge-top="nudgeTop"
-    offset-y
-    min-width="25em"
-    max-width="25em"
-    :z-index="500"
-    class="menu"
-    @input="$emit('input', $event)">
-    <!-- eslint-disable-next-line vue/no-template-shadow -->
-    <template #activator="{ on: menu, attrs }">
-      <v-tooltip location="top">
-        <template #activator="{ on: tooltip }">
-          <v-btn
-            class="align-self-center active-button"
-            icon
-            v-bind="attrs"
-            v-on="{ ...tooltip, ...menu }">
-            <v-icon>
-              <i-mdi-cog />
-            </v-icon>
-          </v-btn>
-        </template>
-        <span>{{ $t('playbackSettings') }}</span>
-      </v-tooltip>
+    location="top">
+    <template #activator="{ props: menu }">
+      <tooltip-button
+        class="align-self-center active-button"
+        v-bind="menu"
+        :tooltip="{ text: $t('playbackSettings'), location: 'top' }"
+        :icon="{ icon: true }">
+        <v-icon>
+          <i-mdi-cog />
+        </v-icon>
+      </tooltip-button>
     </template>
-    <v-card>
-      <v-list color="transparent">
-        <v-list-item disabled>
-          <v-row align="center">
-            <v-col :cols="4">
-              <label>{{ $t('quality') }}</label>
-            </v-col>
-            <v-col :cols="8">
-              <v-select />
-            </v-col>
-          </v-row>
-        </v-list-item>
-        <v-list-item>
-          <v-row align="center">
-            <v-col :cols="4">
-              <label>{{ $t('audio') }}</label>
-            </v-col>
-            <v-col :cols="8">
-              <media-stream-selector
-                v-if="playbackManager.currentItemAudioTracks"
-                :media-streams="playbackManager.currentItemAudioTracks"
-                type="Audio"
-                :default-stream-index="playbackManager.currentAudioStreamIndex"
-                @input="setAudio($event)" />
-            </v-col>
-          </v-row>
-        </v-list-item>
-        <v-list-item v-show="!$vuetify.display.smAndUp">
-          <v-row align="center">
-            <v-col :cols="4">
-              <label>{{ $t('subtitles') }}</label>
-            </v-col>
-            <v-col :cols="8">
-              <media-stream-selector
-                v-if="playbackManager.currentItemSubtitleTracks"
-                :media-streams="playbackManager.currentItemSubtitleTracks"
-                type="Subtitle"
-                :default-stream-index="
-                  playbackManager.currentSubtitleStreamIndex
-                "
-                @input="setSubtitle($event)" />
-            </v-col>
-          </v-row>
-        </v-list-item>
-        <v-list-item disabled>
-          <v-row align="center">
-            <v-col :cols="4">
-              <label>{{ $t('speed') }}</label>
-            </v-col>
-            <v-col :cols="8">
-              <v-select />
-            </v-col>
-          </v-row>
-        </v-list-item>
-        <v-list-item>
-          <v-row align="center">
-            <v-col :cols="4">
-              <label>{{ $t('stretch') }}</label>
-            </v-col>
-            <v-col :cols="8">
-              <v-switch v-model="stretch" />
-            </v-col>
-          </v-row>
-        </v-list-item>
-      </v-list>
+    <v-card min-width="300">
+      <v-card-text>
+        <v-row align="center">
+          <v-col :cols="4">
+            <label>{{ $t('quality') }}</label>
+          </v-col>
+          <v-col :cols="8">
+            <v-select density="comfortable" hide-details disabled />
+          </v-col>
+        </v-row>
+        <v-row align="center">
+          <v-col :cols="4">
+            <label>{{ $t('audio') }}</label>
+          </v-col>
+          <v-col :cols="8">
+            <media-stream-selector
+              v-if="playbackManager.currentItemAudioTracks"
+              :media-streams="playbackManager.currentItemAudioTracks"
+              type="Audio"
+              :default-stream-index="playbackManager.currentAudioStreamIndex"
+              @input="playbackManager.currentAudioStreamIndex = $event" />
+          </v-col>
+        </v-row>
+        <v-row v-if="!$vuetify.display.smAndUp" align="center">
+          <v-col :cols="4">
+            <label>{{ $t('subtitles') }}</label>
+          </v-col>
+          <v-col :cols="8">
+            <media-stream-selector
+              v-if="playbackManager.currentItemSubtitleTracks"
+              :media-streams="playbackManager.currentItemSubtitleTracks"
+              type="Subtitle"
+              :default-stream-index="playbackManager.currentSubtitleStreamIndex"
+              @input="playbackManager.currentSubtitleStreamIndex = $event" />
+          </v-col>
+        </v-row>
+        <v-row align="center">
+          <v-col :cols="4">
+            <label>{{ $t('speed') }}</label>
+          </v-col>
+          <v-col :cols="8">
+            <v-select density="comfortable" disabled hide-details />
+          </v-col>
+        </v-row>
+        <v-row align="center">
+          <v-col :cols="4">
+            <label>{{ $t('stretch') }}</label>
+          </v-col>
+          <v-col :cols="8" class="text-right">
+            <v-switch v-model="stretched" color="primary" hide-details />
+          </v-col>
+        </v-row>
+      </v-card-text>
     </v-card>
   </v-menu>
 </template>
 
-<script lang="ts">
-import { defineComponent } from 'vue';
-import { playbackManagerStore } from '@/store';
+<script lang="ts" setup>
+import { computed } from 'vue';
+import { useVModel } from '@vueuse/core';
+import { playbackManagerStore, playerElementStore } from '@/store';
 
-export default defineComponent({
-  props: {
-    nudgeTop: {
-      type: [Number, String],
-      default: 0
-    },
-    stretchProp: Boolean
-  },
-  setup() {
-    const playbackManager = playbackManagerStore();
+const props = defineProps<{
+  modelValue: boolean;
+}>();
+const emit = defineEmits<{
+  (e: 'update:modelValue', val: boolean): void;
+}>();
 
-    return {
-      playbackManager
-    };
+const menuModel = useVModel(props, 'modelValue', emit);
+
+const playbackManager = playbackManagerStore();
+const playerElement = playerElementStore();
+
+const stretched = computed({
+  get() {
+    return playerElement.isStretched;
   },
-  data() {
-    return {
-      menu: false,
-      stretch: this.stretchProp
-    };
-  },
-  watch: {
-    stretchProp(value) {
-      this.stretch = value;
-    },
-    stretch(value) {
-      this.$emit('stretch', value);
-    }
-  },
-  methods: {
-    setAudio(audioStreamIndex: number) {
-      this.playbackManager.currentAudioStreamIndex = audioStreamIndex;
-    },
-    setSubtitle(subtitleStreamIndex: number) {
-      this.playbackManager.currentSubtitleStreamIndex = subtitleStreamIndex;
-    }
+  set(v: boolean) {
+    playerElement.isStretched = v;
   }
 });
 </script>
--- a/frontend/src/components/Buttons/QueueButton.vue
+++ b/frontend/src/components/Buttons/QueueButton.vue
@@ -1,214 +1,160 @@
 <template>
   <v-menu
-    v-model="menu"
+    v-model="menuModel"
     :close-on-content-click="false"
     :persistent="!closeOnClick"
     :transition="'slide-y-transition'"
-    location="top"
-    :nudge-top="nudgeTop"
-    offset-y
-    min-width="35vw"
-    max-width="35vw"
-    min-height="60vh"
-    max-height="60vh"
-    :z-index="500"
-    class="menu"
-    @input="$emit('input', $event)">
-    <!-- eslint-disable-next-line vue/no-template-shadow -->
-    <template #activator="{ on: menu, attrs }">
-      <v-tooltip location="top">
-        <template #activator="{ on: tooltip }">
-          <v-btn
-            class="align-self-center active-button"
-            icon
-            v-bind="attrs"
-            v-on="{ ...tooltip, ...menu }">
-            <v-icon>
-              <i-mdi-playlist-play />
-            </v-icon>
-          </v-btn>
-        </template>
-        <span>{{ $t('queue') }}</span>
-      </v-tooltip>
+    location="top">
+    <template #activator="{ props: menu }">
+      <tooltip-button
+        class="align-self-center active-button"
+        v-bind="menu"
+        :tooltip="{ text: $t('queue'), location: 'top' }"
+        :btn="{ icon: true }">
+        <v-icon>
+          <i-mdi-playlist-play />
+        </v-icon>
+      </tooltip-button>
     </template>
     <v-card>
       <v-list>
-        <v-list-item>
-          <v-avatar>
-            <blurhash-image
-              v-if="playbackManager.initiator"
-              :item="playbackManager.initiator" />
-            <v-icon v-else :icon="modeIcon" />
-          </v-avatar>
+        <v-list-item :title="sourceText">
+          <template #prepend>
+            <v-avatar>
+              <blurhash-image
+                v-if="playbackManager.initiator"
+                :item="playbackManager.initiator" />
+              <v-icon v-else :icon="modeIcon" />
+            </v-avatar>
+          </template>
 
-          <v-list-item-title>{{ sourceText }}</v-list-item-title>
-          <v-list-item-subtitle>
-            {{ getTotalEndsAtTime(playbackManager.queue) }} -
+          <template #subtitle>
+            {{ getTotalEndsAtTime(playbackManager.queue).value }} -
             {{
               $t('playback.queueItems', {
                 items: playbackManager.queue.length
               })
             }}
-          </v-list-item-subtitle>
+          </template>
 
-          <v-list-item-action>
+          <template #append>
             <like-button
               v-if="playbackManager.initiator && playbackManager.currentItem"
               :item="playbackManager.currentItem" />
-          </v-list-item-action>
-          <v-list-item-action class="mr-1">
             <item-menu
               v-if="playbackManager.initiator && playbackManager.currentItem"
               :item="playbackManager.currentItem" />
-          </v-list-item-action>
+          </template>
         </v-list-item>
       </v-list>
       <v-divider />
       <v-list class="overflow">
         <!-- We set an special property to destroy the element so it doesn't take resources while it's not being used.
         This is especially useful for really huge queues. -->
-        <draggable-queue v-if="menu || !destroy" class="ml-4" />
+        <draggable-queue v-if="menuModel || !destroy" class="ml-4" />
       </v-list>
       <v-spacer />
-      <v-card-actions class="d-flex justify-space-between">
-        <v-tooltip location="top">
-          <template #activator="{ on: tooltip }">
-            <v-btn icon v-on="tooltip" @click="playbackManager.stop">
-              <v-icon>
-                <i-mdi-playlist-remove />
-              </v-icon>
-            </v-btn>
-          </template>
-          <span>{{ $t('playback.clearQueue') }}</span>
-        </v-tooltip>
-        <v-tooltip location="top">
-          <template #activator="{ on: tooltip }">
-            <v-btn icon disabled v-on="tooltip">
-              <v-icon>
-                <i-mdi-content-save />
-              </v-icon>
-            </v-btn>
-          </template>
-          <span>{{ $t('playback.saveAsPlaylist') }}</span>
-        </v-tooltip>
+      <v-card-actions>
+        <tooltip-button
+          :tooltip="{ text: $t('playback.clearQueue'), location: 'top' }"
+          :btn="{ icon: true }"
+          @click="playbackManager.stop">
+          <v-icon>
+            <i-mdi-playlist-remove />
+          </v-icon>
+        </tooltip-button>
+        <tooltip-button
+          :tooltip="{ text: $t('playback.saveAsPlaylist'), location: 'top' }"
+          :btn="{ icon: true, disabled: true }">
+          <v-icon>
+            <i-mdi-content-save />
+          </v-icon>
+        </tooltip-button>
         <v-spacer />
       </v-card-actions>
     </v-card>
   </v-menu>
 </template>
 
-<script lang="ts">
-import { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
-import { defineComponent } from 'vue';
+<script setup lang="ts">
+import { ref, computed, watch } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { useTimeoutFn } from '@vueuse/core';
 import IMdiShuffle from 'virtual:icons/mdi/shuffle';
 import IMdiPlaylistMusic from 'virtual:icons/mdi/playlist-music';
 import { playbackManagerStore } from '@/store';
 import { InitMode } from '@/store/playbackManager';
 import { getTotalEndsAtTime } from '@/utils/time';
 
-export default defineComponent({
-  props: {
-    nudgeTop: {
-      type: [Number, String],
-      default: 0
-    },
-    closeOnClick: {
-      type: Boolean,
-      default: false
-    }
-  },
-  setup() {
-    const playbackManager = playbackManagerStore();
+withDefaults(
+  defineProps<{
+    closeOnClick?: boolean;
+  }>(),
+  { closeOnClick: false }
+);
 
-    return { playbackManager };
-  },
-  data() {
-    return {
-      menu: false,
-      destroy: false,
-      timeout: undefined as undefined | number
-    };
-  },
-  computed: {
-    sourceText(): string {
-      /**
-       * TODO: Properly refactor this once search and other missing features are implemented, as discussed in
-       * https://github.com/jellyfin/jellyfin-vue/pull/609
-       */
-      const unknownSource = this.$t('playback.playbackSource.unknown');
+const playbackManager = playbackManagerStore();
+const { t } = useI18n();
 
-      switch (this.playbackManager.playbackInitMode) {
-        case InitMode.Unknown: {
-          return unknownSource;
-        }
-        case InitMode.Item: {
-          return this.playbackManager.currentItem?.AlbumId !==
-            this.playbackManager.initiator?.Id
-            ? unknownSource
-            : this.$t('playback.playbackSource.item', {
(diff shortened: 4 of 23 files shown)


ferferga (changes requested):
Aside from all these (quick) comments. Why you created your custom components instead of using the templates I set [here](https://github.com/jellyfin/jellyfin-vue/blob/vite/frontend/src/components/Playback/PiPVideoPlayer.vue) and [here](https://github.com/jellyfin/jellyfin-vue/blob/vite/frontend/src/pages/playback/video/index.vue)

I don't see why OsdPlayer must be a component, I don't see them being reusable at all, given that the mini player OSD is really different to the fullscreen video player and I'd rather not have another component filled with ``v-if`` and such.

I'm even in favor to break everything as much as possible, and make individual components for the play/pause, back, forward, repeat and shuffle buttons. Yes, they're really simple, but having them reimplemented multiple times is error and inconsistency prone.

frontend/src/plugins/vuetify.ts
```diff
@@ -59,7 +59,8 @@ const vuetify = createVuetify({
       color: 'primary'
     },
     VBtn: {
-      color: ''
+      color: '',
+      variant: 'text'
```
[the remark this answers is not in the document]
ThibaultNocchi: Yup, it will go away when the style branch is merged :)

frontend/src/store/playbackManager.ts
```diff
@@ -113,7 +115,7 @@ const defaultState: PlaybackManagerState = {
 
 export const mediaElementRef = ref<HTMLMediaElement>();
 const state = reactive<PlaybackManagerState>(cloneDeep(defaultState));
-const mediaControls = useMediaControls(mediaElementRef);
+let mediaControls = useMediaControls(mediaElementRef);
```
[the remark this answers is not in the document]
ThibaultNocchi: This is just my temporary hack to make the controls work in dev mode. This won't be in the updated PR

frontend/src/store/playbackManager.ts
```diff
@@ -793,7 +815,7 @@ class PlaybackManagerStore {
           !isNil(this.currentItem.Id) &&
           !isNil(remote.auth.currentUser)
         ) {
-          await this._reportPlaybackStopped(this.currentItem.Id);
+          this._reportPlaybackStopped(this.currentItem.Id);
```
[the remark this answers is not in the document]
ThibaultNocchi: If you await the reportPlayback, then when you close the player there's a non null time where nothing happens (cause it's waiting for the API call to end) before the players closes.

We could just return the promise sent by the reportPlayback then.

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -54,18 +57,76 @@ const isLoopingOnce = computed(
...
+    if (playbackManager.currentSourceUrl) {
+      // Wait for nextTick for the mediaElementRef to get filled
+      await nextTick();
+
+      try {
+        if (!mediaElementRef.value) {
+          throw new Error('No media element found');
+        }
+
+        // setNewMediaControls(useMediaControls(mediaElementRef));
```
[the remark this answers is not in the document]
ThibaultNocchi: As stated above, it's just there to have the controls work in dev mode for now while I play with it

ThibaultNocchi:
1.	Completely missed the video player page, I'll move everything to it. 
2.	I think the OSD logic being separated avoids us having a big video page script, and it being a component doesn't change its v-if needs (or not). So I don't really agree that's it can't be split, even if it won't be reused. The player element logic + template is already complicated, even if short, that we don't need to add the OSD logic, style and template to it.

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,207 @@
...
 
+/**
+ * Applies the current subtitle from the playbackManager store
+ *
+ * It first disables all the VTT subtitles
+ * If external and VTT, it shows the correct one
+ * If external and SSA, it loads it in SO
+ * If embedded, it triggers a new transcode
+ *
+ * Optionnally force destroy SSO, useful when needing its canvas to move when switching from FS context to minimized
```
ferferga, line 91:
```suggestion
 * Optionally force destroy SSO, useful when needing its canvas to move when switching from FS context to minimized
```

frontend/package.json
```diff
@@ -26,16 +26,19 @@
     "vue-i18n-extract-remove": "vue-i18n-extract --remove"
   },
   "dependencies": {
+    "@jellyfin/libass-wasm": "4.1.1",
     "@jellyfin/sdk": "0.0.0-unstable.202301101857",
     "@vueuse/components": "9.12.0",
     "@vueuse/core": "9.12.0",
+    "@vueuse/shared": "9.12.0",
```
ferferga, line 33: We don't need this, this is included in core

frontend/src/components/Buttons/PlaybackSettingsButton.vue
```diff
@@ -1,143 +1,98 @@
...
-      default: 0
-    },
-    stretchProp: Boolean
-  },
-  setup() {
-    const playbackManager = playbackManagerStore();
+const props = defineProps<{
+  modelValue: boolean;
+}>();
+const emit = defineEmits(['update:modelValue']);
```
ferferga, line 83: Type this properly please

frontend/src/components/Buttons/PlaybackSettingsButton.vue
```diff
@@ -1,143 +1,98 @@
...
-    stretchProp(value) {
-      this.stretch = value;
-    },
-    stretch(value) {
-      this.$emit('stretch', value);
-    }
+const menuModel = useVModel(props, 'modelValue', emit);
+
+const playbackManager = playbackManagerStore();
+const videoPlayer = videoPlayerStore();
```
ferferga, line 88: I would rename this store to ``playerElementStore`` instead

frontend/src/components/Buttons/QueueButton.vue
```diff
@@ -1,214 +1,164 @@
...
                 <i-mdi-playlist-remove />
               </v-icon>
             </v-btn>
           </template>
-          <span>{{ $t('playback.clearQueue') }}</span>
         </v-tooltip>
-        <v-tooltip location="top">
-          <template #activator="{ on: tooltip }">
-            <v-btn icon disabled v-on="tooltip">
+        <v-tooltip :text="$t('playback.saveAsPlaylist')" location="top">
```
ferferga, line 67: Why not migrate this to the new component?

frontend/src/components/Buttons/QueueButton.vue
```diff
@@ -1,214 +1,164 @@
...
   </v-menu>
 </template>
 
-<script lang="ts">
-import { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
-import { defineComponent } from 'vue';
+<script setup lang="ts">
+import { ref, computed, watch } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { useTimeoutFn } from '@vueuse/shared';
```
ferferga, line 85: As said above, shared is not needed. We use shared functions in VirtualGrid without any additional dependency.

frontend/src/components/Buttons/SubtitleSelectionButton.vue
```diff
@@ -1,96 +1,69 @@
...
-      type: [Number, String],
-      default: 0
-    }
-  },
-  setup() {
-    const playbackManager = playbackManagerStore();
+const props = defineProps<{
+  modelValue: boolean;
+}>();
+const emit = defineEmits(['update:modelValue']);
```
ferferga, line 50: Type this properly please

frontend/src/components/Playback/OsdPlayer.vue
```diff
@@ -0,0 +1,354 @@
...
+  window.addEventListener('mousemove', handleMouseMove);
+  window.addEventListener('touchend', handleMouseMove);
+  window.addEventListener('keyup', handleKeyPress);
+});
+
+onBeforeUnmount(() => {
+  window.removeEventListener('mousemove', handleMouseMove);
+  window.removeEventListener('touchend', handleMouseMove);
+  window.removeEventListener('keyup', handleKeyPress);
+});
```
ferferga, line 216: Vue 3 event binding changed, so you should be able to use the ``@`` syntax is components now: https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed.html

Try that first please. In case you have no luck, replace this with https://vueuse.org/core/useEventListener/

Do the same with all the occurences in the code please (this is the only one I found)

frontend/src/components/Playback/PipVideoPlayer.vue
```diff
@@ -0,0 +1,130 @@
...
+  }
+});
+
+const keyPressMappings: { [key: string]: () => void } = {
+  f: () => {
+    if (playbackManager.currentlyPlayingMediaType === 'Video') {
+      videoPlayer.toggleMinimize();
+    }
+  }
+};
```
ferferga, line 99: Why do we need the currentlyPlayingMediaType check? Previous PlayerManagr was full of those because everything was shared between video and audio playback. Now we have a clear distinction between both of them and the pipvideoplayer shouldn't be mounted at all if no video is playing.

frontend/src/pages/playback/video/index.vue
```diff
@@ -1 +1,95 @@
...
+import { playbackManagerStore, videoPlayerStore } from '@/store';
+import { mediaElementRef } from '@/store/videoPlayer';
+
+const router = useRouter();
+const playbackManager = playbackManagerStore();
+const videoPlayer = videoPlayerStore();
+
+const osd = ref(true);
+
+let fullscreen = useFullscreen(document.body);
```
ferferga, line 30: we can remove screenfull then?

frontend/src/components/Playback/PlayerElement.vue
ThibaultNocchi: Done

frontend/package.json
ThibaultNocchi: Done

frontend/src/components/Buttons/PlaybackSettingsButton.vue
ThibaultNocchi: Done

frontend/src/components/Buttons/QueueButton.vue
ThibaultNocchi: Done

frontend/src/components/Buttons/SubtitleSelectionButton.vue
ThibaultNocchi: Done

frontend/src/components/Buttons/PlaybackSettingsButton.vue
ThibaultNocchi: Done

frontend/src/components/Buttons/QueueButton.vue
ThibaultNocchi: Done

frontend/src/App.vue
```diff
@@ -3,6 +3,7 @@
   <v-app>
     <router-view-transition is-root />
     <snackbar />
-    <player-element />
+    <pip-video-player />
```
ferferga, line 6: I think this should live in the default layout. Doesn't make sense to have the pip player ready on the login and music playback layouts.

frontend/src/components/Playback/OsdPlayer.vue
ThibaultNocchi: Ended up using the vueuse listeners cause the v-overlay isn't rendered or shown (if eager) and using event on its parent was too complicated.

frontend/src/pages/playback/video/index.vue
ThibaultNocchi: Done @ferferga

frontend/src/components/Playback/OsdPlayer.vue
```diff
@@ -0,0 +1,353 @@
...
+};
+
+/**
+ * Handles the key presses
+ */
+function handleKeyPress(e: KeyboardEvent): void {
+  if (keyPressMappings[e.key]) {
+    keyPressMappings[e.key]();
+  }
+}
```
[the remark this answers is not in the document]
ThibaultNocchi: Done @ferferga

frontend/src/components/Playback/PipVideoPlayer.vue
```diff
@@ -0,0 +1,130 @@
+<template>
```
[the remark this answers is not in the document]
ThibaultNocchi: DOne @ferferga

frontend/src/components/Playback/PipVideoPlayer.vue
ThibaultNocchi: I switched the check to a check of `overlay` value, which is true when the PiP is shown. I need to render the item though anytime cause for now cause that's the default teleport location for when we're playing a video. I could use the same "isPiPMounted" as I did for FS though

frontend/src/App.vue
ThibaultNocchi: Done @ferferga

frontend/src/components/Buttons/QueueButton.vue
```diff
@@ -1,214 +1,161 @@
...
-        return this.playbackManager.initiator;
-      }
-    },
-    modeIcon(): typeof IMdiShuffle {
-      if (this.playbackManager.playbackInitMode === InitMode.Shuffle) {
-        return IMdiShuffle;
-      }
+const menuModel = ref(false);
+const destroy = ref(false);
+// const timeout = ref<undefined | number>();
```
ThibaultNocchi, line 101: Remove this

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,158 @@
...
+    playbackManager.currentlyPlayingMediaElement === 'video' &&
+    playerElement.isFullscreenMounted
+  ) {
+    return '.fullscreen-video-container';
+  } else if (
+    playbackManager.currentlyPlayingMediaElement === 'video' &&
+    playerElement.isPiPMounted
+  ) {
+    /**
+     * Defaults video to the PiP container to render it
```
ThibaultNocchi, line 68: Comment not needed anymore

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,158 @@
 <template>
-  <template v-if="mediaElement">
+  <template v-if="playbackManager.currentlyPlayingMediaElement">
     <Teleport :to="teleportTarget">
       <component
-        :is="mediaElement"
+        :is="playbackManager.currentlyPlayingMediaElement"
+        v-show="playerElement.isFullscreenMounted || playerElement.isMinimized"
```
ThibaultNocchi, line 6: Convert condition to video check

frontend/src/pages/playback/video/index.vue
```diff
@@ -1 +1,90 @@
...
+ * Toggles the fullscreen view, based on browsers supporting it or not (basically iOS or the others)
+ */
+function toggleFullscreen(): void {
+  if (fullscreen.isSupported.value) {
+    fullscreen.toggle();
+  } else if (
+    !fullscreen.isSupported.value &&
+    // @ts-expect-error - Property 'webkitEnterFullScreen' does not exist on type 'HTMLMediaElement'
+    mediaElementRef.value?.webkitEnterFullScreen
+  ) {
```
ThibaultNocchi, line 42: Add comment that this is for Safari iOS

frontend/src/store/playbackManager.ts
```diff
@@ -1124,6 +1192,34 @@ class PlaybackManagerStore {
...
+      (oldVal, newVal) => {
+        if (
+          oldVal.currentSubtitleStreamIndex !==
+            newVal.currentSubtitleStreamIndex &&
+          (oldVal.currentSubtitleTrack?.DeliveryMethod ===
+            SubtitleDeliveryMethod.Encode ||
+            newVal.currentSubtitleTrack?.DeliveryMethod ===
+              SubtitleDeliveryMethod.Encode)
+        ) {
+          this.fetchCurrentMediaSource();
```
ThibaultNocchi, line 1218: Add comment to specify that this case is when you come from or go to a situation where the sub is burnt in so that a new stream is needed

frontend/src/store/playerElement.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * Applies the current subtitle from the playbackManager store
+   *
+   * It first disables all the VTT subtitles
+   * If external and VTT, it shows the correct one
+   * If external and SSA, it loads it in SO
+   * If embedded, it triggers a new transcode
+   *
+   * Optionally force destroy SSO, useful when needing its canvas to move when switching from FS context to minimized
```
ThibaultNocchi, line 135: Correct comment

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,180 @@
...
   () => playbackManager.currentSourceUrl,
-  () => {
-    if (
-      mediaElementRef.value &&
-      playbackManager.currentSourceUrl &&
-      playbackManager.currentMediaSource?.SupportsDirectPlay &&
-      playbackManager.currentlyPlayingMediaType === 'Audio'
-    ) {
-      mediaElementRef.value.src = playbackManager.currentSourceUrl;
+  async () => {
```
ferferga, line 124: Fix SonarCloud complexity issue here please. Some ideas:

* if ``currentlyPlayingMediaType === 'Audio'`` -> do all the logic for music -> else if ``currentlyPlayingMediaType === 'Video'`` -> do all the logic for video, which includes hls.
* Do a ``const isHlsSupported = Hls.isSupported()`` at the beginning of this component. I highly doubt the truthy value of the hls.js support for the browser will ever change during the lifecycle of the application.
* Do the same with a ``isNativeHlsSupported`` variable.

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,180 @@
...
+      try {
+        if (!mediaElementRef.value) {
+          throw new Error('No media element found');
+        }
+
+        if (
+          (playbackManager.currentMediaSource?.SupportsDirectPlay &&
+            playbackManager.currentlyPlayingMediaType === 'Audio') ||
+          (mediaElementRef.value.canPlayType('application/vnd.apple.mpegurl') &&
+            playbackManager.currentlyPlayingMediaType === 'Video')
```
ferferga, line 140: With my suggestions above, this check would just be a ``if (isNativeHlsSupported)`` in the else if ``currentlyPlayingMediaType === 'Video'`` block. Much simpler!

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,180 @@
...
+           */
+          mediaElementRef.value.src = playbackManager.currentSourceUrl;
+          mediaElementRef.value.currentTime = playbackManager.currentTime;
+        } else if (
+          Hls.isSupported() &&
+          playbackManager.currentlyPlayingMediaType === 'Video'
+        ) {
+          if (!hls) {
+            hls = new Hls({
+              testBandwidth: false,
```
ferferga, line 153: Why this is false? Wouldn't this be useful?

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,180 @@
...
+      } catch {
+        useSnackbar(t('errors.cantPlayItem'), 'error');
+        playbackManager.stop();
+      }
     } else {
-      useSnackbar(t('errors.cantPlayItem'), 'error');
-      playbackManager.stop();
+      if (hls) {
+        hls.destroy();
+        hls = undefined;
```
ferferga, line 174: Make this a function as said above and use it everywhere hls must be destroyed

frontend/src/composables/use-router.ts
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable no-restricted-imports */
+import router from '@/plugins/router';
+/* eslint-enable no-restricted-imports */
+
+/**
+ *	Returns the global router instance. Needed when used outside setup functions or anywhere else outside the Vue app instance.
```
ferferga, line 6:
```suggestion
 * Returns the global router instance. Needed when used outside setup functions or anywhere else outside the Vue app instance.
```

frontend/src/store/playbackManager.ts
```diff
@@ -284,22 +312,35 @@ class PlaybackManagerStore {
...
+    ) || []) as PlaybackExternalTrack[];
+  }
 
-    return (
-      subs?.filter((sub) => sub.codec === 'ass' || sub.codec === 'ssa') || []
-    );
+  public get currentItemAssParsedSubtitleTracks(): PlaybackExternalTrack[] {
+    return (this.currentItemParsedSubtitleTracks?.filter(
+      (sub) => (sub.codec === 'ass' || sub.codec === 'ssa') && sub.src
+    ) || []) as PlaybackExternalTrack[];
```
ferferga, line 331: Same as above

frontend/src/store/playbackManager.ts
```diff
@@ -1092,25 +1157,33 @@ class PlaybackManagerStore {
...
+          router.push('/playback/video');
+        } else if (
+          router.currentRoute.value.fullPath === '/playback/video' &&
+          newIndex === undefined
+        ) {
+          router.replace(
+            typeof router.options.history.state.back === 'string'
+              ? router.options.history.state.back
+              : '/'
+          );
```
ferferga, line 1184: This should be handled by a watcher in the fullscreen page, as commented above

frontend/src/store/playbackManager.ts
```diff
@@ -1124,6 +1197,38 @@ class PlaybackManagerStore {
...
+          this.fetchCurrentMediaSource();
+        }
+      }
+    );
+
+    watch(
+      () => ({
+        currentSubtitleStreamIndex: this.currentSubtitleStreamIndex,
+        currentSubtitleTrack: this.currentSubtitleTrack
+      }),
```
ferferga, line 1213: Can't we merge those watchers together? On both we need to do the same, fetch a new media source.

frontend/src/store/playerElement.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    if (this.isMinimized) {
+      router.push('/playback/video');
+    } else {
+      router.replace(
+        typeof router.options.history.state.back === 'string'
+          ? router.options.history.state.back
+          : '/'
+      );
+    }
```
ferferga, line 87: This is the 3rd time I see this logic. Again, should be possible to have all of this handled just in a watcher in the playback page.

frontend/src/store/playerElement.ts
```diff
@@ -0,0 +1,179 @@
...
+      subtitlesOctopus.freeTrack();
+    }
+  };
+
+  /**
+   * Applies the current subtitle from the playbackManager store
+   *
+   * It first disables all the VTT and SSA subtitles
+   * It then find the potential index of the applied VTT sub
+   * Or the PlaybackExternalTracj object of the potential SSA sub
```
ferferga, line 128:
```suggestion
   * Or the PlaybackExternalTrack object of the potential SSA sub
```

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,180 @@
...
+  ) {
+    return '.fullscreen-video-container';
+  } else if (
+    playbackManager.currentlyPlayingMediaElement === 'video' &&
+    playerElement.isPiPMounted
+  ) {
+    return '.minimized-video-container';
+  } else {
+    return 'body';
+  }
```
ferferga, line 70: Either:

```suggestion
  if (
    playbackManager.currentlyPlayingMediaElement === 'video' &&
    playerElement.isFullscreenMounted
  ) {
    return '.fullscreen-video-container';
  } else if (
    playbackManager.currentlyPlayingMediaElement === 'video' &&
    playerElement.isPiPMounted
  ) {
    return '.minimized-video-container';
  } else {
    return 'body';
  }
```

Or (more a fan of this one):

```suggestion
  if (playbackManager.currentlyPlayingMediaElement === 'audio') {
    return 'body';
  } else if (playbackManager.currentlyPlayingMediaElement === 'video') {
    if (playerElement.isFullscreenMounted) {
       return '.fullscreen-video-container';
    } else if (playerElement.isPiPMounted) {
       return '.minimized-video-container';
    }
 }
```

I mention it in another comment, but I don't understand why the mediaelement check is now in playbackManager.

frontend/src/components/Playback/OsdPlayer.vue
```diff
@@ -0,0 +1,336 @@
+<template>
+  <v-overlay
+    v-model="overlay"
+    contained
+    scrim="transparent"
+    width="100%"
+    height="100%">
+    <div
+      class="d-flex flex-column justify-space-between align-center player-overlay">
+      <div class="osd-top pt-s pl-s pr-s">
```
ferferga, line 10: I believe we can use https://vuetifyjs.com/en/components/overlays/ instead of our custom overlay?

ferferga (changes requested):

frontend/src/components/Playback/PlayerElement.vue
ThibaultNocchi: This helps pick an HLS level, provided you have different ones to pick from. The server handles the "levels" on its side so this doesn't serve any purpose.

frontend/src/components/Buttons/PlaybackSettingsButton.vue
```diff
@@ -1,143 +1,100 @@
...
+          </v-col>
+        </v-row>
+        <v-row align="center">
+          <v-col :cols="4">
+            <label>{{ $t('stretch') }}</label>
+          </v-col>
+          <v-col :cols="8" class="text-right">
+            <v-switch v-model="stretched" color="primary" hide-details />
+          </v-col>
+        </v-row>
```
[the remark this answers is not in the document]
ThibaultNocchi: I initially forgot to just v-if based on the screen size, but you're right. It's not a really used feature so I'll just hide it in the playback settings.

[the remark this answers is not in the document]
ThibaultNocchi: Done

frontend/src/components/Playback/PiPVideoPlayer.vue
```diff
@@ -1 +1,105 @@
...
+onMounted(() => {
+  playerElement.isPiPMounted = true;
+});
+
+onUnmounted(() => {
+  playerElement.isPiPMounted = false;
+});
+</script>
+
+<style>
```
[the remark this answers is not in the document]
ThibaultNocchi: This is needed as the class is applied to the div by the v-overlay logic, so when scoped the compiler doesn't include the classes as it's not explicitly called in the component.

```vue
<v-overlay
    :model-value="true"
    persistent
    no-click-animation
    scrim
    scroll-strategy="none"
    content-class="minimized-overlay"
    :width="$vuetify.display.mobile ? '60vw' : '25vw'">
```

frontend/src/composables/use-router.ts
ThibaultNocchi: Done

frontend/src/pages/playback/video/index.vue
```diff
@@ -1 +1,93 @@
...
+
+  playerElement.isFullscreenMounted = false;
+});
+
+onMounted(() => {
+  playerElement.isFullscreenMounted = true;
+});
+</script>
+
+<style>
```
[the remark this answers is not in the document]
ThibaultNocchi: The <video> tag isn't explicitly rendered in the component, so if it's scoped it won't be included in the build and it doesn't work.

frontend/src/store/playbackManager.ts
```diff
@@ -284,22 +312,35 @@ class PlaybackManagerStore {
...
+   * Filters the native subtitles
+   *
+   * As our profile requires either SSA or VTT, if it's not SSA it'll be VTT.
+   * This is done this way as server sends as "Codec" the initial value of the track, so it can be webvtt, subrip, srt...
+   * This is easier to filter out the SSA subs
+   */
+  public get currentItemVttParsedSubtitleTracks(): PlaybackExternalTrack[] {
+    return (this.currentItemParsedSubtitleTracks?.filter(
+      (sub) => sub.codec !== 'ass' && sub.codec !== 'ssa' && sub.src
+    ) || []) as PlaybackExternalTrack[];
```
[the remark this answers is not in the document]
ThibaultNocchi: That's a technical you maybe could help. 

`this.currentItemParsedSubtitleTracks` is a `PlaybackTrack[]` and `PlaybackExternalTrack` is roughly the same item, with a few changes and especially a `src: string` and not `src?: string` as in `PlaybackTrack`.

As you see, with the filter I only keep defined `src` properties, but if I don't put the `as` cast, the linter thinks `src` can be undefined.

frontend/src/store/playbackManager.ts
ThibaultNocchi: I preferred not to as that'd not be an easy to understand if clause haha. But I don't mind either, as you wish.

frontend/src/store/playerElement.ts
ThibaultNocchi: Done

frontend/src/store/playbackManager.ts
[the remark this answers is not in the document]
aweebs: Apologies, I missed this mention earlier. I think this is a good use for a type predicate, the `is` keyword, in typescript.
These aren't actually much better than an `as` cast since they still override the typing. However, I think their use makes the intent of the code much more understandable. Especially when used with array filters.

In this case we can make the filter be a type predicate. But for completeness we should also make sure that `sub.codec` is explicitly defined.
```
return (this.currentItemParsedSubtitleTracks?.filter(
      (sub): sub is PlaybackExternalTrack => sub.src && sub.codec && sub.codec !== 'ass' && sub.codec !== 'ssa'
    ) ?? []);
```

frontend/src/components/Playback/OsdPlayer.vue
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// HACK: https://github.com/vuetifyjs/vuetify/issues/8436.
+// https://vuetifyjs.com/en/api/v-btn/#retain-focus-on-click prop was added
+// but it seems we're using a prop combination that it's incompatible with it: NaN;
+
+// SO link: https://stackoverflow.com/questions/57830767/is-it-default-for-vuetify-to-keep-active-state-on-buttons-after-click-how-do-yo/57831256#57831256
+.active-button:focus::before {
+  opacity: 0 !important;
+}
```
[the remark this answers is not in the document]
ThibaultNocchi: Done

frontend/src/components/Playback/OsdPlayer.vue
```diff
@@ -0,0 +1,336 @@
...
+whenever(keys.left, playbackManager.skipBackward);
+whenever(keys.j, playbackManager.skipBackward);
+whenever(keys.f, () => emit('toggleFullscreen'));
+whenever(keys.m, playbackManager.toggleMute);
+
+useEventListener(document, 'mousemove', handleMouseMove);
+useEventListener(document, 'touchend', handleMouseMove);
+</script>
+
+<style lang="scss">
```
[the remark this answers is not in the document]
ThibaultNocchi: Fixed by no more CSS

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,180 @@
...
+        // try to recover network error
+        console.log('fatal network error encountered, try to recover');
+        hls?.startLoad();
+        break;
+      }
+      case ErrorTypes.MEDIA_ERROR: {
+        console.log('fatal media error encountered, try to recover');
+        hls?.recoverMediaError();
+        break;
+      }
```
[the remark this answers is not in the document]
ThibaultNocchi: Done

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -1,66 +1,180 @@
...
+      default: {
+        // cannot recover
+        hls?.destroy();
```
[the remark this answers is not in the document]
ThibaultNocchi: DOne

frontend/src/components/Playback/PlayerElement.vue
ThibaultNocchi: Done

frontend/src/components/Playback/PlayerElement.vue
ThibaultNocchi: Done

frontend/src/components/Playback/PlayerElement.vue
ThibaultNocchi: Done

frontend/src/store/playbackManager.ts
```diff
@@ -1092,25 +1157,33 @@ class PlaybackManagerStore {
...
-            state.currentSourceUrl = this.getItemPlaybackUrl();
-          }
+          useSnackbar(t('errors.cantPlayItem'), 'error');
+          this.stop();
+        } else if (
+          this.currentlyPlayingMediaType === 'Video' &&
+          oldIndex === undefined &&
+          newIndex !== undefined
+        ) {
+          router.push('/playback/video');
```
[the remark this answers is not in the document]
ThibaultNocchi: Now that the logic is moved in PlayerElement, do I still need this check? I wouldn't think so

frontend/src/store/playbackManager.ts
ThibaultNocchi: Moved to playerElement

frontend/src/store/playbackManager.ts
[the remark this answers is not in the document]
ThibaultNocchi: Thanks a lot, it's corrected!

frontend/src/store/playbackManager.ts
ThibaultNocchi: Done

frontend/src/pages/playback/video/index.vue
```diff
@@ -1 +1,93 @@
...
+     */
+    // TODO - if entering FS this way, SSA subs won't display. So we should trigger a new encode
+    // @ts-expect-error - Property 'webkitEnterFullScreen' does not exist on type 'HTMLMediaElement'
+    mediaElementRef.value?.webkitEnterFullScreen();
+  }
+}
+
+if (!playbackManager.currentItem) {
+  router.replace('/');
+}
```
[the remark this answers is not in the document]
ThibaultNocchi: Done and moved to PlayerElement

ferferga (approved):

PR #1882, opened 2023-02-10
chore: little bit of cleanup

aweebs:
Several smaller unrelated commits that I've stacked together.
Let me know if you'd like separate PRs for these, but I figure they're small enough it isn't worth it.
I recommend reviewing the diffs per commit, since they're separated logically in that way.

### DIFF
--- a/frontend/src/components/Buttons/PlayButton.vue
+++ b/frontend/src/components/Buttons/PlayButton.vue
@@ -35,92 +35,62 @@
   </div>
 </template>
 
-<script lang="ts">
+<script setup lang="ts">
 import { BaseItemDto } from '@jellyfin/sdk/lib/generated-client';
-import { defineComponent } from 'vue';
+import { ref } from 'vue';
 import { playbackManagerStore } from '@/store';
 import { canResume, canPlay } from '@/utils/items';
 import { ticksToMs } from '@/utils/time';
 
-export default defineComponent({
-  props: {
-    item: {
-      type: Object as () => BaseItemDto,
-      required: true
-    },
-    iconOnly: {
-      type: Boolean
-    },
-    fab: {
-      type: Boolean
-    },
-    shuffle: {
-      type: Boolean
-    },
-    videoTrackIndex: {
-      type: Number,
-      default: undefined
-    },
-    audioTrackIndex: {
-      type: Number,
-      default: undefined
-    },
-    subtitleTrackIndex: {
-      type: Number,
-      default: undefined
-    },
-    disabled: {
-      type: Boolean,
-      default: false
-    }
-  },
-  setup() {
-    const playbackManager = playbackManagerStore();
+const props = withDefaults(
+  defineProps<{
+    item: BaseItemDto;
+    iconOnly?: boolean;
+    fab?: boolean;
+    shuffle?: boolean;
+    videoTrackIndex?: number;
+    audioTrackIndex?: number;
+    subtitleTrackIndex?: number;
+    disabled?: boolean;
+  }>(),
+  { disabled: false }
+);
 
-    return { playbackManager };
-  },
-  data() {
-    return {
-      loading: false
-    };
-  },
-  methods: {
-    async playOrResume(): Promise<void> {
-      this.loading = true;
+const playbackManager = playbackManagerStore();
 
-      if (this.item && canResume(this.item)) {
-        await this.playbackManager.play({
-          item: this.item,
-          audioTrackIndex: this.audioTrackIndex,
-          subtitleTrackIndex: this.subtitleTrackIndex || -1,
-          videoTrackIndex: this.videoTrackIndex,
-          startFromTime:
-            ticksToMs(this.item.UserData?.PlaybackPositionTicks) / 1000
-        });
-      } else if (this.shuffle) {
-        /**
-         * We force playback from the start when shuffling, since you wouldn't resume AND shuffle at the same time
-         */
-        await this.playbackManager.play({
-          item: this.item,
-          audioTrackIndex: this.audioTrackIndex,
-          subtitleTrackIndex: this.subtitleTrackIndex || -1,
-          videoTrackIndex: this.videoTrackIndex,
-          startShuffled: true
-        });
-      } else {
-        await this.playbackManager.play({
-          item: this.item,
-          audioTrackIndex: this.audioTrackIndex,
-          subtitleTrackIndex: this.subtitleTrackIndex || -1,
-          videoTrackIndex: this.videoTrackIndex
-        });
-      }
+const loading = ref(false);
 
-      this.loading = false;
-    },
-    canPlay,
-    canResume
+/** Begins or resumes playing of the item */
+async function playOrResume(): Promise<void> {
+  loading.value = true;
+
+  if (props.item && canResume(props.item)) {
+    await playbackManager.play({
+      item: props.item,
+      audioTrackIndex: props.audioTrackIndex,
+      subtitleTrackIndex: props.subtitleTrackIndex || -1,
+      videoTrackIndex: props.videoTrackIndex,
+      startFromTime:
+        ticksToMs(props.item.UserData?.PlaybackPositionTicks) / 1000
+    });
+  } else if (props.shuffle) {
+    // We force playback from the start when shuffling, since you wouldn't resume AND shuffle at the same time
+    await playbackManager.play({
+      item: props.item,
+      audioTrackIndex: props.audioTrackIndex,
+      subtitleTrackIndex: props.subtitleTrackIndex || -1,
+      videoTrackIndex: props.videoTrackIndex,
+      startShuffled: true
+    });
+  } else {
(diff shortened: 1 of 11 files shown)


frontend/src/components/Buttons/PlayButton.vue
```diff
@@ -35,92 +35,62 @@
   </div>
 </template>
 
-<script lang="ts">
+<script setup lang="ts">
```
aweebs, line 38: Github absolutely destroyed this diff, but the changes are actually very minimal. None of the logic was changed, just switched to composition setup api

frontend/src/components/Item/CollectionTabs.vue
```diff
@@ -19,46 +19,33 @@
   </div>
 </template>
 
-<script lang="ts">
+<script setup lang="ts">
```
aweebs, line 22: Github also destroyed this diff too. No logic changes, just conversion to composition setup api

frontend/src/store/userLibraries.ts
```diff
@@ -141,7 +141,8 @@ class UserLibrariesStore {
         state.value.homeSections.audioResumes = audioResumes;
       }
     } catch (error) {
-      useSnackbar(error as string, 'error');
+      // TODO: actual user-facing error messages
```
aweebs, line 144: This seems to be the only place in the codebase we try to snackbar an error. I'm happy to remove these in favor of a `console.error` or remove the TODOs if a real translated error isn't ever going to be a thing.

ThibaultNocchi: I think snackbars are better than silently throwing errors for users that could be meaningful. So yeah we could add some proper errors :)

ThibaultNocchi (changes requested):

frontend/src/components/Item/CollectionTabs.vue
```diff
@@ -19,46 +19,33 @@
...
+}>();
+
+const currentTab = ref(0);
+const loading = ref(false);
+const children = computed(() => {
+  return groupBy(items.getChildrenOfParent(props.item.Id), 'Type');
+});
+
+watch(
+  [props.item],
```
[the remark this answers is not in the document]
ferferga: Not preferred, it's the only one that works when watching subproperties.

frontend/package.json
```diff
@@ -26,7 +26,7 @@
     "vue-i18n-extract-remove": "vue-i18n-extract --remove"
   },
   "dependencies": {
-    "@jellyfin/sdk": "0.0.0-unstable.202301101857",
+    "@jellyfin/sdk": "0.7.0",
```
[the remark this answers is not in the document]
aweebs: I'll remove this commit for now, thanks for letting me know. There's only one real bug filed in that repo right now. Does that mean the next release should be a good time to upgrade, or are there other issues you're aware of that aren't filed which would prevent that?

frontend/src/components/Item/CollectionTabs.vue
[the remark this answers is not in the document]
aweebs: Ah right, I'm learning Vue as I go - I'm used to React with the dependency lists. I did some reading of https://vuejs.org/guide/essentials/watchers.html, and it seems that I need a getter here because props are not reactive by default, and thus a getter is needed to perform comparison. I'll update this to use a getter.

As a quick question, I had a really hard time understanding if props are made reactive by default. My current understanding is no, but they will be reactive if a reactive object is passed in as a prop value. Is that correct?

frontend/src/store/userLibraries.ts
aweebs: Follow up question: Should we SnackBar the raw JS error like in this case, or should we show the generic error message in the translation file (`errors.anErrorHappened`) and console.error the actual JS error.

ThibaultNocchi: I'd do your second option: snackbar a generic message and console.error the real one :)

frontend/src/components/Item/CollectionTabs.vue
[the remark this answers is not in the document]
ThibaultNocchi: The props item in general is reactive. With watchers, you may need to enable the `deep` attribute (it was on by default iirc, but I think they've changed it) or use a getter. But it may be easier to watch for an attribute, and those aren't reactive unless you pass a reactive object indeed. So for a single property, you need to use a getter like what you did in your latest change :)

ThibaultNocchi (approved):

ThibaultNocchi:
LGTM. You may want to snacbkar the generic message and console.error the errors before we merge.

frontend/src/store/userLibraries.ts
ferferga: Yeah, all our snackbars needs a lot of cleanup (but probably that it's more worth it for another PR, I have this pending on my mind but need to write it to the roadmap at some point).

We have a lot of messages for a lot of actions that can fail (e.g like an item, mark as watched, etc...), where all of these could be resumed to the following messages:

* An error has happened
* The operation couldn't be performed.

frontend/src/components/Item/CollectionTabs.vue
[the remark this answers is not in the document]
ferferga: @ThibaultNocchi @aweebs It wasn't by default before. Read this: https://vuejs.org/guide/essentials/watchers.html#deep-watchers and this: https://v2.vuejs.org/v2/api/#vm-watch

My experience while migrating the app and deep diving the docs tells me that in Vue 3 watchers:

*Assume getters the following syntax: ``() => whatever``*

* Refs and root reactive objects: **Don't need getters**
      * defineProps returns a proxied object with ``reactive``
      * Root reactive objects are basically this: ``const a = reactive(...)``

* Primitives inside reactive objects and reassignment of objects inside reactive objects: **A getter is enough**
* Reassignment inside objects inside reactive objects: **A getter and deep is needed**
     * E.G is not the same to do ``a.b = c`` (being b and c objects) than doing ``a.b.e = false``. In the second one, you're mutating a value inside the nested object, not the nested object itself.

frontend/package.json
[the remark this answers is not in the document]
aweebs: @ThibaultNocchi bumping my above question. I can merge without it answered, but I'm curious if there's any other issues that aren't filed which we can't upgrade because of?

aweebs:
@ThibaultNocchi snackbar updates for `userLibraries` are in. Feel free to merge when you're ready.

frontend/src/main.ts
```diff
@@ -11,10 +11,10 @@
 
 import { createApp } from 'vue';
 import Root from '@/App.vue';
-/* eslint-disable no-restricted-imports */
+// eslint-disable-next-line no-restricted-imports
 import { createRemote, i18n, router, vuetify } from '@/plugins';
+// eslint-disable-next-line no-restricted-imports
```
ferferga, line 16: I agree for the other changes, but this one being multiline, i think it's better to use the previous method to disable and reenable later

frontend/vite.config.ts
```diff
@@ -1,7 +1,7 @@
-/* eslint-disable import/no-nodejs-modules */
+// eslint-disable-next-line import/no-nodejs-modules
 import path, { dirname, resolve } from 'node:path';
+// eslint-disable-next-line import/no-nodejs-modules
 import { fileURLToPath } from 'node:url';
-/* eslint-enable import/no-nodejs-modules */
```
ferferga, line 4: Same as I mentioned above

ferferga (changes requested):

frontend/src/main.ts
aweebs: Here's my case for preferring single line vs a block:
1. Currently it's the same amount of comment lines, so there's no code savings really
2. The blocks shouldn't get much larger, since that would indicate we're making a lot of exceptions to our linting rules, which in general is bad
3. We can always switch these to a disable/enable scheme if they do get much larger

Let me know if I've swayed you or if you'd still like these reverted :)

aweebs: Thanks for that explanation @ferferga. I'd also like to clarify my position a bit:
Linting rules exist to enforce code standards. There will always be exceptions to those rules, which is why these exist. However, we should try to limit those exceptions, and adding individual comments is a clear way to show that a specific line should be ignored. Disable/enable should be saved for large sections where linting needs to be diabled, and these just don't exist in our code base.

That all being said, my opinion here that single is better than block, especially in this case, isn't strongly enough held that I'd want to prevent the rest of the PR from getting merged, so I'll revert this for now :)

I'm not sure if there's a linting rule to enforce certain line counts, but having a lint rule for lint exceptions is an interesting thought :D.

ferferga: @aweebs There are (https://mysticatea.github.io/eslint-plugin-eslint-comments/) and we use some of them. In fact, the use of enable-disable with the rule specified is enforced by that plugin.

ferferga (approved):

PR #1884, opened 2023-02-15
chore: more cleanup

aweebs:
Some more cleanup commits stacked on each other.

Goals that this PR builds towards:
1. No more `@ts-expect-error` exceptions - these can be avoided by typecasting where necessary. It's less safe to completely ignore any static compiling error on a line than to resolve an error with a typecast. Linting rule coming in future PR.
2. No more unnecessary `as` casts - linting rule incoming in later PR once these are all removed.
3. `typecheck` script runs successfully

Recommend reviewing the changes commit by commit

### DIFF
--- a/frontend/src/components/Item/Metadata/MetadataEditor.vue
+++ b/frontend/src/components/Item/Metadata/MetadataEditor.vue
@@ -78,7 +78,7 @@
             <v-combobox
               v-model="metadata.Genres"
               v-model:search-input="search"
-              :items="genders"
+              :items="genres"
               :label="$t('genres')"
               hide-selected
               multiple
@@ -95,7 +95,7 @@
             <v-combobox
               v-model="metadata.Tags"
               v-model:search-input="search"
-              :items="genders"
+              :items="genres"
               :label="$t('tags')"
               hide-selected
               multiple
@@ -231,7 +231,7 @@ export default defineComponent({
       menu: false,
       dialog: false,
       person: null as BaseItemPerson | null,
-      genders: [] as BaseItemDto[] | null | undefined,
+      genres: [] as BaseItemDto[] | null | undefined,
       search: '',
       loading: false,
       tabName: null
@@ -277,14 +277,23 @@ export default defineComponent({
     async getData(): Promise<void> {
       await this.fetchItemInfo();
 
+      if (!this.metadata.Id) {
+        return;
+      }
+
       const ancestors = await this.$remote.sdk
         .newUserApi(getLibraryApi)
         .getAncestors({
-          itemId: this.metadata.Id as string,
+          itemId: this.metadata.Id,
           userId: this.$remote.auth.currentUserId
         });
-      const libraryInfo =
-        ancestors.data.find((index) => index.Type === 'CollectionFolder') || {};
+      const libraryInfo = ancestors.data.find(
+        (index) => index.Type === 'CollectionFolder'
+      );
+
+      if (!libraryInfo || !libraryInfo.Id) {
+        return;
+      }
 
       this.getGenres(libraryInfo.Id);
     },
@@ -298,12 +307,15 @@ export default defineComponent({
 
       this.$data.metadata = itemInfo;
     },
-    async getGenres(parentId = ''): Promise<void> {
-      this.genders = (
-        await this.$remote.sdk.newUserApi(getGenresApi).getGenres({
-          parentId
-        })
-      ).data.Items?.map((index) => index.Name) as BaseItemDto[];
+    async getGenres(parentId: string): Promise<void> {
+      this.genres =
+        (
+          await this.$remote.sdk.newUserApi(getGenresApi).getGenres({
+            parentId
+          })
+        ).data.Items?.map((index) => index.Name).filter(
+          (genre): genre is string => !!genre
+        ) ?? [];
     },
     async saveMetadata(): Promise<void> {
       const item = pick(this.metadata, [
@@ -348,8 +360,13 @@ export default defineComponent({
 
       try {
         this.loading = true;
+
+        if (!this.metadata.Id) {
+          throw new Error('Expected metadata to have id');
+        }
+
         await this.$remote.sdk.newUserApi(getItemUpdateApi).updateItem({
-          itemId: this.metadata.Id as string,
+          itemId: this.metadata.Id,
           baseItemDto: item
         });
         this.$emit('save');
@@ -395,7 +412,7 @@ export default defineComponent({
       this.dialog = result;
     },
     handlePersonDel(index: number): void {
-      (this.metadata.People as BaseItemPerson[]).splice(index, 1);
+      (this.metadata.People ?? []).splice(index, 1);
     }
   }
 });
--- a/frontend/src/components/Skeletons/SkeletonHomeSection.vue
+++ b/frontend/src/components/Skeletons/SkeletonHomeSection.vue
@@ -1,6 +1,6 @@
 <template>
   <v-col>
-    <!-- TODO: Wait for Vuetify 3.1 -->
+    <!-- TODO: Wait for Vuetify 3 implementation (https://github.com/vuetifyjs/vuetify/issues/13504) -->
     <!-- <v-skeleton-loader type="heading" max-width="25em" class="ml-2" /> -->
     <v-row class="space-around ma-0">
       <skeleton-card
--- a/frontend/src/pages/settings/index.vue
+++ b/frontend/src/pages/settings/index.vue
@@ -141,9 +141,10 @@ import { version as clientVersion } from '@/../package.json';
 const { t } = useI18n();
 const route = useRoute();
 const remote = useRemote();
-let systemInfo = {} as SystemInfo;
 
 route.meta.title = t('settings.settings');
+
+let systemInfo: SystemInfo = {};
 
 if (remote.auth.currentUser?.Policy?.IsAdministrator) {
   systemInfo = (await remote.sdk.newUserApi(getSystemApi).getSystemInfo()).data;
--- a/frontend/src/store/items.ts
+++ b/frontend/src/store/items.ts
@@ -156,7 +156,7 @@ class ItemsStore {
 
     state.value.collectionById[parent.Id] = childIds;
 
-    return this.getChildrenOfParent(parent.Id) as BaseItemDto[];
+    return this.getChildrenOfParent(parent.Id) ?? [];
   };
 
   /**
@@ -196,7 +196,11 @@ class ItemsStore {
     if (childItems.Items) {
       const parent = this.getItemById(parentId);
 
-      return this.addCollection(parent as BaseItemDto, childItems.Items);
+      if (!parent) {
+        throw new Error('expected item to be found');
+      }
+
+      return this.addCollection(parent, childItems.Items);
     }
   };
 
--- a/frontend/src/store/items.ts
+++ b/frontend/src/store/items.ts
@@ -81,35 +81,19 @@ class ItemsStore {
    * @param payload
    * @returns - The reactive references
    */
-  public add = (
-    payload: BaseItemDto | BaseItemDto[]
-  ): BaseItemDto | BaseItemDto[] => {
-    const isArray = Array.isArray(payload);
-    const itemArray = isArray ? payload : [payload];
-
-    const res: BaseItemDto[] = [];
-
-    for (const item of itemArray) {
-      if (!item.Id) {
-        throw new Error("One item doesn't have an id");
-      }
-
-      state.value.byId[item.Id] = item;
-
-      const fetched = this.getItemById(item.Id);
-
-      if (!fetched) {
-        throw new Error('Expected to receive newly added item to store');
-      }
-
-      res.push(fetched);
+  public add = (item: BaseItemDto): BaseItemDto => {
+    // Without an id this cannot be cached so return a non-reactive version
+    if (!item.Id) {
+      return item;
     }
 
-    if (res.length === 1 && !isArray) {
-      return res[0];
+    state.value.byId[item.Id] = item;
+    const fetched = this.getItemById(item.Id);
+    if (!fetched) {
+      throw new Error('Expected to retrieve newly added item to store');
     }
 
-    return res;
+    return fetched;
   };
 
   /**
--- a/frontend/src/components/Buttons/SubtitleSelectionButton.vue
+++ b/frontend/src/components/Buttons/SubtitleSelectionButton.vue
@@ -79,8 +79,7 @@ export default defineComponent({
   },
   computed: {
     tracks(): PlaybackTrack[] {
-      const subs = this.playbackManager
-        .currentItemParsedSubtitleTracks as PlaybackTrack[];
+      const subs = this.playbackManager.currentItemParsedSubtitleTracks ?? [];
 
       return [
         {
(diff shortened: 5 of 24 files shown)


frontend/src/components/Layout/Carousel/CarouselProgressBar.vue
```diff
@@ -24,6 +24,10 @@
 import { defineComponent } from 'vue';
 import { useResponsiveClasses } from '@/composables';
 
+interface Data {
```
ferferga, line 27: For this file, Swiper 9 includes a new autoplay module. Would you mind trying to move this component to use Swiper's autoplay API: https://swiperjs.com/swiper-api#autoplay (which after 9.X version allows for pausing, which is something that was not possible before)

Also, using swiper native paginators: https://swiperjs.com/swiper-api#pagination

This is something I was debating wether to do after or before vite branch gets merged because the earlier it gets merged, the merrier (so we don't have master frozen for longer), but now that you took the plunge to fix the linting errors here, do you want to take the challenge? :)

frontend/src/components/Layout/Carousel/Item/ItemsCarouselTitle.vue
```diff
@@ -31,96 +31,79 @@
...
 
-            if (this.item.AlbumArtist) {
-              this.titleString = this.item.AlbumArtist;
-            }
+const { t } = useI18n();
 
-            if (this.item.Name) {
-              this.subtitle = this.item.Name;
-            }
+const itemLink = ref('');
```
ferferga, line 45: All of this can probably be computed instead of a watcher

frontend/src/components/Layout/SwiperSection.vue
```diff
@@ -44,7 +44,6 @@
 </template>
 
 <script setup lang="ts">
-import { Swiper, SwiperSlide } from 'swiper/vue';
```
ferferga, line 47: Reviewing on GitHub so probably I'm missing something, but why is this unnecessary? We're using those components in the template

frontend/src/components/Playback/PlaybackInfoCard.vue
```diff
@@ -1,427 +0,0 @@
-<template>
```
ferferga, line 1: This is not unused, it's pending the reimplementation of the video player (#1878)

frontend/src/components/Skeletons/SkeletonHomeSection.vue
```diff
@@ -13,33 +13,34 @@
   </v-col>
 </template>
 
-<script lang="ts">
-import { defineComponent } from 'vue';
+<script setup lang="ts">
+import { computed } from 'vue';
+import { useVuetify } from '@/composables';
```
ferferga, line 18:
```suggestion
import { useDisplay } from 'vuetify';
```

If you read the comments of that composable, that composable is just meant to be used outside Vue components (i.e raw TS files). Here we could use Vuetify composables as stated in their docs.

frontend/src/components/Skeletons/SkeletonItemGrid.vue
```diff
@@ -15,11 +15,8 @@ import { computed } from 'vue';
 import { CardShapes, getShapeFromItemType } from '@/utils/items';
 import { useResponsiveClasses } from '@/composables';
 
-const props = defineProps({
-  viewType: {
-    type: String,
-    default: (): string => 'Movie'
-  }
+const props = withDefaults(defineProps<{ viewType?: string }>(), {
```
ferferga, line 18: Looking at GitHub, so maybe I'm not recalling stuff correctly, but, afaik, the viewTypes are BaseItemKind, right? Doesn't make sense to use the BaseItemKind type?

frontend/src/pages/settings/index.vue
```diff
@@ -141,10 +141,11 @@ import { version as clientVersion } from '@/../package.json';
 const { t } = useI18n();
 const route = useRoute();
 const remote = useRemote();
-let systemInfo = {} as SystemInfo;
 
 route.meta.title = t('settings.settings');
 
+let systemInfo: SystemInfo = {};
```
ferferga, line 147: Why this position change?

ferferga (changes requested):
I'm not a fan at all of the changes to the data interface - to me, it seems like another way to remove ``as`` just for the sake of it. In your other PR I didn't mind it in the metadata pages because the fixes in the page were good and it was a nitpick really. However, modifying files purposefully just for this is, in my opinion, "polluting" the commit history (I'm more concerned about the quality and usefulness of our commits since our repo starting getting bigger and bigger and it's no longer a 1 second to clone 😣) without any real improvement, when the real improvement would've been to directly migrate all those files to Composition API.

I'd love to keep the rest of the runtime checks/handles you added, but please also take the extra step (which I know is cumbersome and annoying at times) to migrate them to Composition API 🙏.

ferferga:
@aweebs By the way, as mentioned here, you can chat with us in Matrix if you want to get quick replies to your questions and in a more informal way

We're bad at doing detailed roadmaps or in publicily document our preferences and conventions (for commits, for instance) because we get all of them resolved in chat (between @ThibaultNocchi and me) and right now we want to get stuff done instead of focusing in proper documentation (which is also much less appealing to work in :P). In the meantime, hopefully we can have you around in Matrix so you don't end up doing extra work that could have been saved with a question in chat beforehand :P

frontend/src/components/Layout/SwiperSection.vue
aweebs: Oof, VS Code wasn't pulling me to the uses of it so I thought they were unused types. I'll get these back in!

frontend/src/components/Layout/Carousel/Item/ItemsCarouselTitle.vue
aweebs: Definitely, I initially just copied the current behavior, but I'm happy to switch this. This could either be broken into multiple simple computed values, or a single computed value which returns an object. I'm going to go ahead with a single object since the logic for computing them all is related, but let me know if that's bad practice.

frontend/src/components/Playback/PlaybackInfoCard.vue
aweebs: Unused in the sense that the current branch didn't have any references to it :)
ThePR you referenced doesn't change it (yet) and a lot of the component logic is pretty dependent on the old shaka player types. Would you prefer I leave it untouched or should I clean it out so that it's not relying on the shaka player types?

frontend/src/components/Skeletons/SkeletonItemGrid.vue
aweebs: I didn't look at this when I converted it initially since I was scoping my changes, but I think you're right, and there's very few uses of this component. I'll add a commit for this change 🚀

frontend/src/pages/settings/index.vue
aweebs: The position was changed so that it's right above the part of the code where it gets re-assigned. I just thought it was slightly easier to follow. Happy to move it back, but heads up that the line will still have a diff since the type declaration style is changed.

frontend/src/components/Layout/Carousel/CarouselProgressBar.vue
aweebs: I'm be super happy to pick this up as my first dive into non-typing changes now that I've had some time to poke around in the project and feel like I have a grasp of what's going. I'll do that as my next task, and I'll put it in a separate PR to compartmentalize the changes.

aweebs:
> I'm not a fan at all of the changes to the data interface - to me, it seems like another way to remove as just for the sake of it. In your other PR I didn't mind it in the metadata pages because the fixes in the page were good and it was a nitpick really. However, modifying files purposefully just for this is, in my opinion, "polluting" the commit history (I'm more concerned about the quality and usefulness of our commits since our repo starting getting bigger and bigger and it's no longer a 1 second to clone 😣) without any real improvement, when the real improvement would've been to directly migrate all those files to Composition API.
> I'd love to keep the rest of the runtime checks/handles you added, but please also take the extra step (which I know is cumbersome and annoying at times) to migrate them to Composition API 🙏.

I'm conflicted about this. I think these Data interface changes do represent some amount of progress since they help with type safety, but agree that they're not significant at the moment. There's 15 components were updated, and making the full transition to the Composition API will take me a while, but I'm happy to continue forward on that path as separate tasks/PRs.
I think this is a good example of incremental progress vs perfection.
In terms of git history pollution, I don't think worrying about repo clone times is a valid concern. The git history will only increase, and this repo is still very small comparatively. I also don't believe that this obstructs the history - it's a single commit and easy to go backwards with if a git blame is necessary.

> @aweebs By the way, as mentioned here, you can chat with us in Matrix if you want to get quick replies to your questions and in a more informal way

I'll definitely join the Matrix chat, thanks for pointing that out!

frontend/src/components/Playback/PlaybackInfoCard.vue
ferferga: That should be answered by @ThibaultNocchi :)

frontend/src/components/Skeletons/SkeletonItemGrid.vue
aweebs: Update from some quick investigation: `library/_viewId` will need some large refactors, so I'll look into redoing it with composition API but that will take a while. I'll work on it but I don't think we need to block this PR on that being done.

ferferga:
@aweebs 

> I think these Data interface changes do represent some amount of progress since they help with type safety

They don't. The rest of the changes in some of those files (because even not all of the files are modified beyond the Data interface, like ``QueueButton.vue``) **add type safety** with if checks and the use of the nullish coalescing (``??``) operand. However, the data interface **itself** doesn't change anything runtime-wise. The only change is the aesthetics of how those types are defined for the compiler. And just change the aesthetic for the sake of change aesthetic without any real improvement (I don't like ``as`` either, but using it makes for much less verbose code than a new interface) while the real improvement in type safety (aside from the if checks and null coalescing) is to use Composition API, which has better type-checking support from Volar during development, cleaner definition of types with defineProps, etc etc... It's the real solution, this is, as with everything cosmetic, just a highly subjective patch. And the effort of doing it *objectively* better by migrating to Composition API is not much bigger (albeit I know it's sometimes frustrating).

And yes, I know the repo will grow, the point is not that, but the fact that the commits are as relevant as possible. And, as said before, the interface approach is sbjectively relevant, while Composition API is objectively relevant with not much more effort. When you do a git blame, you must go commit by commit. Going commit by commit takes time, and takes more time if you have even more commits. Style commits for instance are not useful at all when debugging, but commits where a file was migrated to Composition API is 100x times more useful.

I'm sorry, but this is a hard no for me. We've been carrying some technical debt issues specifically because of lazy commits and workarounds that "we can change/refactor in future PRs" (and I'm the first culprit of doing that) and we end up completely forgetting or abandoning. The metadata editing section and library view, as you saw, require heavy refactoring just for this sole reason, in my opinion. Bigger stuff of course will still be affected and will require revisits in the future and multiple PRs. But I'd like to get PRs merged in the best possible state within a reasonable scope (that's also why vite is taking so long to get merged, because I could've started working in master directly and leave the client broken for a lot of time).

It's true the LoC for this PR will get much bigger with the migration, but I don't mind reviewing them here if you don't want the extra hassle. However, if you prefer another PR, just cherrypick the data interface commit to another branch and start migrating those modified components to Composition API and amend the name afterwards. You can also do it in different commits when the scope of those components is different so, as said above, the git blame is more useful.

TL:DR: The Data interface is just a subjective middle-ground for improvements in DX when a migration to Composition API is the real deal and this just seems a lazy (with all due respects, because I know finding all the occurences of as takes some time) approach.

frontend/src/components/Skeletons/SkeletonItemGrid.vue
ferferga: @aweebs Yup of course. In fact, I wouldn't worry too much about migrating that page to Composition API for the moment. Look at this: https://github.com/jellyfin/jellyfin-vue/discussions/1679

In order to accomplish that new style of media-type centric navigation, new library pages for every kind of type (artists, genres, shows, etc) will be created, so the logic of each page will be more simple. It's probably not worth the effort to port anything when we want to rewrite that entirely in the near future.

frontend/src/pages/settings/index.vue
ferferga: @aweebs No problem, just was curious about it. I personally stick to leave all the variables at the top of the component, but I don't have any strong feelings about it either.

frontend/src/components/Playback/PlaybackInfoCard.vue
aweebs: Going to remove this from the PR for now and follow up on Matrix

aweebs:
@ferferga I've removed the Data interface change commit, as well as the commit for deleting `PlaybackInfoCard`. The rest of the feedback had been incorporated.

ferferga (approved):
A some files are conflicting with the implement video playback PR (#1878) and #1889, if you don't mind, I'm going to be cherry-picking commits that I'm sure that don't conflict with those, then merge those PRs, and then fix the conflicts of the remaining commits from this PR.

But here you have the greencheck of your changes :)

frontend/src/store/items.ts
```diff
@@ -81,35 +81,19 @@ class ItemsStore {
...
-
-      const fetched = this.getItemById(item.Id);
-
-      if (!fetched) {
-        throw new Error('Expected to receive newly added item to store');
-      }
-
-      res.push(fetched);
+  public add = (item: BaseItemDto): BaseItemDto => {
+    // Without an id this cannot be cached so return a non-reactive version
```
aweebs, line 85: This change is necessary since calls like getting devices or apikeys also end up coming through this function since, see my comment in `axios/index`, but they won't have a valid Id, and this was previously failing. Now if there no id the non-reactive value is still returned without being cached so those calls will continue working.

ferferga: @aweebs I like your fallback when the item doesn't have an id.

About the axios interceptor, ideally we should strictly check for everything before calling this method in the interceptor. Previously in the item store we had all the API methods (some, not all of them), which was a maintenance burden. The interceptor-way of working was a middle-ground I came up with to ease the maintenance burden, however it didn't end up working because the Vue 2 limitations. 

Just saying all of this if you want to take another challenge :): Investigate all the endpoints that returns BaseItemDto or BaseItemPerson and filter  them properly.

aweebs: I'll add it to my list of things to work on :)
I also added a fairly descriptive TODO to the interceptor so that anyone else who comes across it is also aware of the limitations and that there is work to be done there.

frontend/src/components/Item/Metadata/MetadataEditor.vue
```diff
@@ -298,12 +307,15 @@ export default defineComponent({
...
       this.$data.metadata = itemInfo;
     },
-    async getGenres(parentId = ''): Promise<void> {
-      this.genders = (
-        await this.$remote.sdk.newUserApi(getGenresApi).getGenres({
-          parentId
-        })
-      ).data.Items?.map((index) => index.Name) as BaseItemDto[];
+    async getGenres(parentId: string): Promise<void> {
+      this.genres =
```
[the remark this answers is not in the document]
aweebs: Ah, good catch! These changes were separated from my Data interface commit, and I missed including the rename. Pushing the fix now!

ferferga (approved):

PR #1885, opened 2023-02-15
Added support for spatial navigation (Navigation using arrow keys) and also support for tv remote keys

gibahjoe:
This PR adds support for Spatial navigation which is especially useful when using jellyfin-vue on a TV.

Also support for TV remote keys during playback (Tested with LGTV remote) was also added. 

The main focus of this PR is to let TV users select items on their screen using the arrow keys of the remote and also to allow the remote media keys (tested with LGTV magic remote, webOS 5 and 6 but it can be easily expanded for other tv types if their keycodes are different).

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = {
   ignorePatterns: [
     '.eslintrc.js',
     '*.config.js',
+    'spatialNavigation.js',
     'types/global/routes.d.ts',
     'types/global/components.d.ts'
   ],
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -1,6 +1,6 @@
 <template>
   <backdrop />
-  <v-app>
+  <v-app v-focus-section:app>
     <router-view-transition is-root />
     <snackbar />
   </v-app>
--- a/frontend/src/assets/styles/global.scss
+++ b/frontend/src/assets/styles/global.scss
@@ -99,3 +99,9 @@ html.no-forced-scrollbar {
   top: 1em;
   right: 1em;
 }
+
+button.bg-primary:focus,
+button.bg-primary:focus-visible {
+  box-shadow: 0 0 0 5px white !important;
+  outline: none !important;
+}
--- a/frontend/src/components/Buttons/Playback/PlayButton.vue
+++ b/frontend/src/components/Buttons/Playback/PlayButton.vue
@@ -2,6 +2,7 @@
   <div class="d-inline-flex">
     <v-btn
       v-if="canPlay(item) && (fab || iconOnly)"
+      v-focus
       :variant="iconOnly ? undefined : 'elevated'"
       :color="iconOnly ? undefined : 'primary'"
       icon
@@ -17,6 +18,7 @@
     </v-btn>
     <v-btn
       v-else-if="!fab"
+      v-focus
       :disabled="disabled || !canPlay(item)"
       :loading="loading"
       class="mr-2"
--- a/frontend/src/components/Item/Card/Card.vue
+++ b/frontend/src/components/Item/Card/Card.vue
@@ -1,5 +1,8 @@
 <template>
-  <div :class="{ 'card-margin': margin }">
+  <div
+    v-focus="!!link"
+    :class="{ 'card-margin': margin }"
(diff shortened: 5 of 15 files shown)


Ritchie1108:
so, anybody knows? does this project support smart tv remote?
i want to use it to replace jellyfin-web in jellyfin-tizen

ThibaultNocchi:
I haven't yet gone around to review this PR, we had a lot of work with the recent Vue 3 merge.

I don't mind adding support for keys and remotes, what I don't guarantee is that the client will work on smart TVs. We use a pretty recent software stack, with which TVs usually aren't compatible.

But adding support for keys isn't something I'll block, I just need to take some time :)

Ritchie1108:
sorry, im not good at web code, so im not not sure what the browser version this project depends on
maybe chromium m85?

ThibaultNocchi:
It seems at least to need Chromium 87, as Vite specifies.

Ritchie1108:
oops, so sad

gibahjoe:
@ThibaultNocchi The previous vue app worked on TVs (tested on LGTV) but Vite didn't as at 3 months ago when I tested it. I will take a look at it again and resolve these merge conflicts

ferferga:
@gibahjoe Still it's useful if this can be navigated through keyboard for accessibility reasons or TVs that have latest Chromium versions.

There's no harm in supporting as much use cases as possible, as long as the support is good!

gibahjoe:
Update:

I already began work on this a few days back. Unfortunately, the plug-in I used previously doesn't support Vue 3 so I have had to copy over the code and make a couple of adjustments.

Anyway, I should be done by this weekend.

ferferga:
@gibahjoe The other day I was thinking about the media type redesign and it came across to my mind again what to do for this PR.

In general, I think we should have three views for libraries: card grid (the only one right now), coverflow one (the current one for fullscreen music player, which is already useful for touch-based units like Echo Show or Android Auto but it also might be a presentation people like for books and TVs) and a list view (specially for music)

For the home sections, music and that hypothetical coverflow view, we already have full keyboard support through swiper.

Why I'm saying all of this? Maybe before you do further work here, perhaps we should gather feedback from people who might benefit from this the most. I'm not a TV user myself, so all I can do is give my review in the impl details, etc, but maybe not much insight in how it should be UX-wise

gibahjoe:
@ferferga That is fine.

endrl:
Playground for WebOS and Tizen is now open ;)
https://github.com/jellyfin/jellyfin-vue/pull/1967

endrl:
The question is should be "keyboard navigation" with auto focus and highlight a default feature which is always enabled? Or a build flag? Webos and tizen and any other client that might want to embed jellyfin-vue (tizen and webos will build the interface with compatibility flags/shims anyway).

Implementing some media keys for the player(s) might be a good solution [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
```
KeyboardEvent: key='MediaPlayPause' | code=''
KeyboardEvent: key='MediaTrackNext' | code=''
KeyboardEvent: key='MediaTrackPrevious' | code=''
KeyboardEvent: key='AudioVolumeDown' | code=''
KeyboardEvent: key='AudioVolumeUp' | code=''
```

I had a look regarding key code mapping.
- I don't know what `jellyfin-web` has implemented at this point
- `jellyfin-tizen` i found no hacks. See [Tizen docs](https://developer.samsung.com/smarttv/develop/api-references/tizen-web-device-api-references/tvinputdevice-api.html) and [jellyfin-tizen](https://github.com/jellyfin/jellyfin-tizen/blob/master/tizen.js#L182)
- `jellyfin-webos` has basic [key code mapping](https://github.com/jellyfin/jellyfin-webos/blob/master/frontend/js/index.js#L94). However the media keys and partly the navigation seems to work out of the box

webos and tizen might be able to inject something like this if really required (Not sure if this really works).
``` js
// https://stackoverflow.com/questions/8776543/how-to-catch-event-keycode-and-change-it-to-another-keycode
Object.defineProperty(KeyboardEvent.prototype, 'key', {
    get: function() {
        switch (this.keyCode) {
            case 19: return 'MediaPlayPause';
            case 415: return 'MediaPlayPause';
            default: return this.key
        }
    }
})
```

PR #1881, opened 2023-02-09
feat(vue3): create tooltippedbutton component

ferferga:

PR #1886, opened 2023-02-16
fix: fix failing github action setup

aweebs:
Currently all builds are failing the `Lint` and `Typecheck` (and other) steps due to an error with 
setting up the node environment, `ENOWORKSPACES`. From the Lint logs for recent PRs (like #1881):
image

These errors are because the npm cache option doesn't work with npm workspaces - try running `npm config get cache` locally to confirm this. This PR removes the github actions cache option to allow the github action workflows to run.

The jobs still fail now because of linting and type checking errors, but the steps actually run to completio.
This should help monitor how PRs increase/decrease lint rule and type checking violations.

### DIFF
--- a/.github/workflows/automation.yml
+++ b/.github/workflows/automation.yml
@@ -9,7 +9,7 @@ on:
     branches:
       - master
     paths-ignore:
-      - '**/*.md'
+      - "**/*.md"
   pull_request_target:
 
 jobs:
@@ -63,10 +63,10 @@ jobs:
         if: ${{ github.event_name == 'pull_request_target' }}
         continue-on-error: true
         with:
-          repo-token: '${{ secrets.JF_BOT_TOKEN }}'
+          repo-token: "${{ secrets.JF_BOT_TOKEN }}"
 
       - name: Check all PRs for merge conflicts ⛔
         uses: eps1lon/actions-label-merge-conflict@v2.1.0
         with:
-          dirtyLabel: 'merge conflict'
+          dirtyLabel: "merge conflict"
           repoToken: ${{ secrets.JF_BOT_TOKEN }}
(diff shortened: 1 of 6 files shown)


ferferga:
I don't understand what are you trying to accomplish with this PR, as workflows work perfectly. They are currently failing because ``npm run typecheck`` and ``npm run lint`` are failing, which is expected as there are still type errors and linting issues in the codebase.

Tauri is failing because there's a huge mismatch between the Rust versions and JS packages (check https://github.com/jellyfin/jellyfin-vue/pull/1738 and https://github.com/jellyfin/jellyfin-vue/pull/1770). As soon as viet gets merged and renovate dependencies go afterwards, I expect those to get fixed. 

Bear in mind that even there was a node version switch in vite (master is in 16, vite in 18)

aweebs:
> I don't understand what are you trying to accomplish with this PR, as workflows work perfectly. They are currently failing because `npm run typecheck` and `npm run lint` are failing, which is expected as there are still type errors and linting issues in the codebase.
> 
> Tauri is failing because there's a huge mismatch between the Rust versions and JS packages (check #1738 and #1770). As soon as viet gets merged and renovate dependencies go afterwards, I expect those to get fixed.
> 
> Bear in mind that even there was a node version switch in vite (master is in 16, vite in 18)

The lint and typechecking steps are actually currently failing on setup, and not on execution. I've updated the description to better explain this. Check the failure result of the Typecheck and Lint steps for any recent PR to verify.

aweebs:
@ferferga bumping this, I've updated the description to include the reason for it

ferferga:
@aweebs Just updated the checklist of the main PR to include this: https://github.com/jellyfin/jellyfin-vue/pull/1812

I just finished testing the new setup-node update and the issue is also present in master.
I want to investigate why the npm team disabled that command for workspaces, it makes no sense to me and might probably be a bug (in that case I'll file an issue).

They've been breaking workspaces in every major npm update lately, so this doesn't surprise me.

PR #1887, opened 2023-02-18
Added intro skipper button 

gibahjoe:
This PR adds a "skip intro" button which is an implementation of the intro skipper plugin https://github.com/ConfusedPolarBear/intro-skipper

Screenshot 2023-02-18 at 15 43 01

The above plugin is required for this to work.


speatzle:
Hi, i just ran across this and gave it a shot. Works Pretty well.

I would just like to know why the skip button appears on the left side as opposed to the right side like all other intro skip implementation that i know of like ConfusedPolarBear's Jellyfin-web, Plex or Netflix.

speatzle:
Just found a bug where when using the minimized player and the skip intro button is visible the middle of the play and pause symbols is not clickable. Clicking the rest of the Player still works fine for pausing or resuming playback.

gibahjoe:
> Hi, i just ran across this and gave it a shot. Works Pretty well.
> 
> I would just like to know why the skip button appears on the left side as opposed to the right side like all other intro skip implementation that i know of like ConfusedPolarBear's Jellyfin-web, Plex or Netflix.

For some reason, I remembered netflix's own being on the left. I will adjust that and move it to the right.

gibahjoe:
> Just found a bug where when using the minimized player and the skip intro button is visible the middle of the play and pause symbols is not clickable. Clicking the rest of the Player still works fine for pausing or resuming playback.

Thanks for the feedback. The button was overlapping with the other on screen buttons. I have resolved that.

Ge082:
Would be also great if there is an option to choose between "do nothing", "show skip intro button" or "automatically skip intro" like Emby does

gibahjoe:
> Would be also great if there is an option to choose between "do nothing", "show skip intro button" or "automatically skip intro" like Emby does
> 
> But this is already neat

This can be done from the admin UI (Jellyfin web) in the intro skipper plugin settings. The admin UI has not been ported to Jellyfin-vue yet.

HEPOSHEIKKI:
Works great, thank you! I think what would fit great is if the skip intro button had a hover effect that cancels the hiding of the button like [KDE does with notifications](https://imgur.com/a/3Oujg3E). Also it would be nice if we could buffer some video after the intro end so skipping would feel much smoother but this would probably require a second FFMPEG thread or something like that.

HEPOSHEIKKI:
> Works great, thank you! I think what would fit great is if the skip intro button had a hover effect that cancels the hiding of the button like [KDE does with notifications](https://imgur.com/a/3Oujg3E).

I have created a PR into gibahjoe:skip-intro from HEPOSHEIKKI:skip-intro for this, please review.

Ge082:
> > Would be also great if there is an option to choose between "do nothing", "show skip intro button" or "automatically skip intro" like Emby does
> > But this is already neat
> 
> This can be done from the admin UI (Jellyfin web) in the intro skipper plugin settings. The admin UI has not been ported to Jellyfin-vue yet.

Hmmm, true, but shouldn't this be a user-based option rather than admin (Maybe the plugin itself still doesn't support that way) like Emby does ?

HEPOSHEIKKI:
> > > Would be also great if there is an option to choose between "do nothing", "show skip intro button" or "automatically skip intro" like Emby does
> > > But this is already neat
> > 
> > This can be done from the admin UI (Jellyfin web) in the intro skipper plugin settings. The admin UI has not been ported to Jellyfin-vue yet.
> 
> Hmmm, true, but shouldn't this be a user-based option rather than admin (Maybe the plugin itself still doesn't support that way) like Emby does ?

You should submit an issue to `confusedpolarbear:intro-skipper`

As this is not yet a feature in the plugin

HEPOSHEIKKI:
Changed merge title on gibahjoe fork, should fix check failure.

ferferga:
@gibahjoe I think @ThibaultNocchi already commented this to you, but no PRs are getting merged to master until ``vite`` branch is in. You could rebase to that branch in the meantime, but this probably won't get in either after the merge as we're focusing in getting things fixed, no new features.

We also need to discuss how to approach this kind of components that are not part of the "core" server, but based on plugins.

HEPOSHEIKKI:
> @gibahjoe I think @ThibaultNocchi already commented this to you, but no PRs are getting merged to master until `vite` branch is in. You could rebase to that branch in the meantime, but this probably won't get in either after the merge as we're focusing in getting things fixed, no new features.
> 
> We also need to discuss how to approach this kind of components that are not part of the "core" server, but based on plugins.

My bad, I must've missed the comments,

gibahjoe:
@ferferga You are right, the frontend should not be implementing non-core plugins. Feel free to close.

As an aside, maybe Jellyfin should consider making a skip intro plugin a default plugin. It's a very much needed feature and may prevent users from checking out this vue project.

nielsvanvelzen:
> As an aside, maybe Jellyfin should consider making a skip intro plugin a default plugin. It's a very much needed feature and may prevent users from checking out this vue project.

We've reached out to the plugin author but they're not interested in upstreaming the plugin (https://github.com/ConfusedPolarBear/intro-skipper/issues/53). Simultaneously someone from our team started work on a core implementation for segment skipping but progress on that has slowed down (https://github.com/jellyfin/jellyfin-meta/discussions/30). In summary; we are interested in making intro skipping a core feature of the Jellyfin server so we can support it in all official clients.

ferferga:
Just a heads up for everybody participating in this: @ThibaultNocchi and me have been discussing this internally.

Although [we want to get this into Jellyfin server itself](https://github.com/ConfusedPolarBear/intro-skipper/issues/53), this is unfortunately still a plugin, hence not something that we should support *directly* into the core (as I also mentioned [here](https://github.com/jellyfin/jellyfin-vue/pull/1887#issuecomment-1465911782)). However, we want to be as flexible as possible and we understand this is a good feature and something people want to have.

I think the best approach for this would be to have it as an independent micro-frontend that is completely decoupled from `@jellyfin-vue/frontend` package (hence, having it's own npm workspace and such). If the client has been built with it, at runtime it can be attached to the video player's DOM and also import what `@jellyfin-vue/frontend` exports (only `playbackManager` comes to my mind, to watch for changes in the playback status so the intro-skipper microfrontend knows when to mount and unmount)

[single-spa's](https://single-spa.js.org/docs/getting-started-overview) documentation is a really good thing to start reading about the possibilities microfrontends could give us for "jellyfin-vue plugins" while keeping the core "on-point" with the core features of server.

Any help is appreciated! 🙌💗

alanstrok:
How can I install this on docker ?

ferferga:
Hello

I'm going to close this since 10.10 already includes the [`MediaSegments` API](https://jellyfin.org/docs/general/server/metadata/media-segments/), so this could now be reimplemented in the core's Jellyfin server way, instead of catering to unofficial plugins.

PR #1890, opened 2023-02-23
Rewrite library view

aweebs:
Generally works the same as the previous page, with the bonus of having a watching on the viewId parameter for when libraries change.

### DIFF
--- a/frontend/src/components/Buttons/FilterButton.vue
+++ b/frontend/src/components/Buttons/FilterButton.vue
@@ -1,11 +1,14 @@
 <template>
-  <v-menu :close-on-content-click="false" max-width="250px">
+  <v-menu
+    :disabled="disabled"
+    :close-on-content-click="false"
+    max-width="250px">
     <template #activator="{ props: menuProps }">
       <v-btn
         v-if="!$vuetify.display.smAndDown"
         class="ma-2"
         v-bind="mergeProps(menuProps, props)">
-        {{ $t('filter') }}
+        {{ t('filter') }}
         <v-icon end>
           <i-mdi-menu-down />
         </v-icon>
@@ -17,7 +20,7 @@
       </v-btn>
     </template>
     <v-expansion-panels variant="accordion" class="dropdown">
-      <v-expansion-panel :title="$t('status')">
+      <v-expansion-panel :title="t('status')">
         <v-expansion-panel-text>
           <v-list
             v-model:selected="selectedStatusFilters"
@@ -38,12 +41,7 @@
           </v-list>
         </v-expansion-panel-text>
       </v-expansion-panel>
-      <v-expansion-panel
-        v-if="
-          collectionInfo.CollectionType === 'movies' ||
-          collectionInfo.CollectionType === 'tvshows'
-        "
-        :title="$t('features')">
+      <v-expansion-panel v-if="isMovieOrTvShow" :title="t('features')">
         <v-expansion-panel-text>
           <v-list
             v-model:selected="selectedFeatureFilters"
@@ -64,7 +62,7 @@
           </v-list>
         </v-expansion-panel-text>
       </v-expansion-panel>
-      <v-expansion-panel v-if="genreFilters.length > 0" :title="$t('genres')">
+      <v-expansion-panel v-if="genreFilters.length > 0" :title="t('genres')">
         <v-expansion-panel-text>
           <v-list
             v-model:selected="selectedGenreFilters"
@@ -87,7 +85,7 @@
       </v-expansion-panel>
       <v-expansion-panel
         v-if="ratingFilters.length > 0"
-        :title="$t('parentalRatings')">
+        :title="t('parentalRatings')">
         <v-expansion-panel-text>
           <v-list
             v-model:selected="selectedRatingFilters"
@@ -108,12 +106,7 @@
           </v-list>
         </v-expansion-panel-text>
       </v-expansion-panel>
-      <v-expansion-panel
-        v-if="
-          collectionInfo.CollectionType === 'movies' ||
-          collectionInfo.CollectionType === 'tvshows'
-        "
-        :title="$t('videoTypes')">
+      <v-expansion-panel v-if="isMovieOrTvShow" :title="t('videoTypes')">
         <v-expansion-panel-text>
           <v-list
             v-model:selected="selectedTypeFilters"
@@ -134,7 +127,7 @@
           </v-list>
         </v-expansion-panel-text>
       </v-expansion-panel>
-      <v-expansion-panel v-if="yearFilters.length > 0" :title="$t('years')">
+      <v-expansion-panel v-if="yearFilters.length > 0" :title="t('years')">
         <v-expansion-panel-text>
           <v-list
             v-model:selected="selectedYearFilters"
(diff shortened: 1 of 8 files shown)


frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -110,10 +112,9 @@
       </v-expansion-panel>
       <v-expansion-panel
         v-if="
-          collectionInfo.CollectionType === 'movies' ||
-          collectionInfo.CollectionType === 'tvshows'
+          item.CollectionType === 'movies' || item.CollectionType === 'tvshows'
```
ferferga, line 115: This check is repeated. Can't we move it to a computed property?

frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -249,18 +266,16 @@ const typeFilters = [
  * applying filters and sorting
```
ferferga, line 266: Fix the ``refesh`` typo above this please (GH doesn't let me select that)

frontend/src/components/Buttons/FilterButton.vue
```diff
@@ -292,12 +307,7 @@ function emitFilterChange(): void {
     years: selectedYearFilters.value
   });
 }
-watch(
-  () => props.itemsType,
-  () => {
-    refreshItems();
-  }
-);
+watch(() => props.item, refreshItems);
```
ferferga, line 310: The logic behind the library/filter/sort buttons have always seemed so tangled to me that I never tried to understood it properly (not your fault ofc, in fact you're brave by trying to fix past messes!)
However, better be safe than worry: are you sure we don't need to call ``refreshItems`` at mount? It's truly only needed when filters are applied?

In case you discover we need it at mount, make this watcher immediate please.

frontend/src/components/Buttons/SortButton.vue
```diff
@@ -1,20 +1,24 @@
...
       <v-btn
         v-if="!$vuetify.display.smAndDown"
         class="my-2"
         v-bind="mergeProps(menuProps, props)">
         {{
-          $t('sortByType', {
+          t('sortByType', {
             type:
               model.length === 0
                 ? items[0].title
```
ferferga, line 24: Maybe this can be computed too?

frontend/src/components/Buttons/TypeButton.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-  <v-menu>
+  <v-menu :disabled="disabled">
     <template #activator="{ props: menuProps }">
       <v-btn
-        v-if="!$vuetify.display.smAndDown && (model.length === 0 || model[0])"
```
ferferga, line 5: Why? You're cluttering the setup with this change imo. At the past, @ThibaultNocchi and me discussed about the different methods of importing icons and we agreed that having icons both defined as prop or as components is fine, since the templates can get very cluttered with how Vuetify uses icon as slots. However, in this case, we take even more space (when icons use around 1:1 of the space)

frontend/src/components/Buttons/TypeButton.vue
```diff
@@ -28,29 +28,35 @@
...
-    type: string;
+    /** CollectionType */
+    type: string | undefined;
     disabled?: boolean;
   }>(),
   { disabled: false }
 );
 const emit = defineEmits<{
-  (e: 'change', types: Record<string, string>): void;
+  (e: 'change', types: BaseItemKind | undefined): void;
```
ferferga, line 45: Thank you for replacing the remnants of not having ``BaseItemKind`` properly typed!

frontend/src/utils/__tests__/items.spec.ts
```diff
@@ -46,23 +46,23 @@ describe('getShapeFromCollectionType', () => {
...
   it('returns the correct card shape based on item type', () => {
     expect(getShapeFromItemType('Audio')).toEqual(CardShapes.Square);
-    expect(getShapeFromItemType('folder')).toEqual(CardShapes.Square);
-    expect(getShapeFromItemType('musicalbum')).toEqual(CardShapes.Square);
-    expect(getShapeFromItemType('musicartist')).toEqual(CardShapes.Square);
-    expect(getShapeFromItemType('musicgenre')).toEqual(CardShapes.Square);
-    expect(getShapeFromItemType('photoalbum')).toEqual(CardShapes.Square);
-    expect(getShapeFromItemType('playlist')).toEqual(CardShapes.Square);
-    expect(getShapeFromItemType('video')).toEqual(CardShapes.Square);
+    expect(getShapeFromItemType('Folder')).toEqual(CardShapes.Square);
```
ferferga, line 49: I don't care about getting this change in or out, but I was actually thinking on removing all the test files before merging and do proper testing from the start using Vitest or Cypress (which was the next major thing I wanted to tackle after Vue 3).

If you are going to be around and want to help in the testing suite, I'll leave this up to you (maybe you prefer to incrementally refactor, instead of a complete redo?) :).

frontend/src/pages/library/_viewId.vue
```diff
@@ -21,34 +21,42 @@
...
+        :disabled="loading"
         @change="onChangeFilter" />
       <v-spacer />
-      <play-button :item="collectionInfo" shuffle :disabled="noContent" />
-      <play-button :item="collectionInfo" :disabled="noContent" />
+      <play-button
+        v-if="library"
+        :item="library"
+        shuffle
+        :disabled="loading || items.length === 0" />
```
ferferga, line 37: Same as above, maybe we can have this computed?

frontend/src/pages/library/_viewId.vue
```diff
@@ -57,258 +65,224 @@ import { getMusicGenresApi } from '@jellyfin/sdk/lib/utils/api/music-genres-api'
...
-      if (this.collectionInfo.Name) {
-        this.$route.meta.title = this.collectionInfo.Name;
-      }
+const library = ref<BaseItemDto>();
+const loadingLibrary = ref(false);
+watch(library, (lib) => {
+  if (lib && lib.Name && lib.Type && validLibraryTypes.includes(lib.Type)) {
+    route.meta.title = lib.Name;
+  }
+});
```
ferferga, line 80: Please, let's try to keep all the concerns together. The order is not that relevant to me, but ideally all definitions (refs or computed), functions and then watchers at the bottom.

Why? Besides looking more clear and easier to find them if we stick to some convention at this, we avoid ``ReferenceError`` when an immediate watcher could be fired before a variable is assigned.

Also, maybe this can be computed too?

frontend/src/pages/library/_viewId.vue
```diff
@@ -57,258 +65,224 @@ import { getMusicGenresApi } from '@jellyfin/sdk/lib/utils/api/music-genres-api'
...
-          break;
-        }
-        default: {
-          await this.refreshItems();
-          break;
-        }
+const items = ref<BaseItemDto[]>([]);
+const loadingItems = ref(false);
+watch(() => (route.params as { viewId: string }).viewId, fetchLibrary);
+fetchLibrary((route.params as { viewId: string }).viewId);
```
ferferga, line 85: Make this an immediate watcher

frontend/src/pages/library/_viewId.vue
```diff
@@ -57,258 +65,224 @@ import { getMusicGenresApi } from '@jellyfin/sdk/lib/utils/api/music-genres-api'
...
+    console.error(error);
+    useSnackbar(t('failedToRefreshItems'), 'error');
+  } finally {
+    loadingLibrary.value = false;
+  }
+}
 
-      try {
-        let itemsResponse;
+async function refreshItems(): Promise<void> {
```
ferferga, line 186: Said the ``ReferenceError`` thing above and then I discover this function is defined after the watcher :)
See what I mean?

ferferga (changes requested):

frontend/src/components/Buttons/TypeButton.vue
aweebs: I like the idea of the code part of the template explicitly declaring all the requirements that the template will need, rather than have "global" variables in the template. But if there's been a decision to not do this then I will revert it.

frontend/src/utils/__tests__/items.spec.ts
aweebs: I'd prefer to keep it in, I'm a fan of incremental changes in the right direction. I think having these old test cases around will help when the test suite is upgraded. I hope I'll still be around to help with that!

frontend/src/pages/library/_viewId.vue
aweebs: I'll re-organize like that. I was organizing these by the parts that they impacted.

Not sure this can be computed though since there's an assignment as part of it which means it's a side effect and should be a watch.

frontend/src/components/Buttons/TypeButton.vue
ferferga: @aweebs I thought this was the reasoning behind, so somehow we agreed beforehand :).

And yes, I agree some things like this are confusing in Vue (like the ref unwrapping). No strong feelings about it, just wanted to be sure if it was conscious and not because you thought it wasn't working, for instance.

frontend/src/utils/__tests__/items.spec.ts
ferferga: @aweebs Okay. Be aware anyway that removing them will still make them available in history, so we're not losing them really. It's just a matter of deciding between how easy it will be to review new files vs or incrementally modify them (because if the modifications are going to be huge, maybe it's not worth it).

I looked at vitest API and it looks really similar to jest, so I'm not sure what to think (this is why I'm asking :P)

frontend/src/pages/library/_viewId.vue
ferferga: @aweebs My point is to have a computed ``libName`` for the if checking and then assign it like ``route.meta.title = libName.value``.

Though now that I'm thinking about this... I'm not sure if maybe we don't need the watcher or the computed property at all since lib should already be reactive? Can you check it please?

aweebs: Using just an assignment without a computed value doesn't work. Neither does assigned it to a computed value like so:
```
const title = computed(() => {
  if (
    library.value &&
    library.value.Name &&
    library.value.Type &&
    validLibraryTypes.includes(library.value.Type)
  ) {
    return library.value.Name;
  }
});

route.meta.title = title.value;
```

I think this will need to stay as a watcher

aweebs: I'm going to simplify this too so that it always just assigned the Name, since only libraries should be loaded here I don't think there's any point in checking if it's a valid library type.

aweebs:
I think the feedback has largely been addressed, the last thing that's needed before merge is to debounce the item search and also check the filter type button since I think it's triggering another search once the filters load.

frontend/src/pages/library/_viewId.vue
ferferga: @aweebs Fine then.

The check could be moved to the middleware level instead, it doesn't hurt really. Doing that in this PR is up to you though.

aweebs:
> @aweebs This is good to go as soon as the unused debounce import is removed and the comment is fixed imo. The refresh is probably triggered by the type button watcher that I mentioned above.

@ferferga I've removed the unused import for now and update the multiline comment to docblock which I think was the other blocker. We can add debouncing at another time.

ferferga (approved):

PR #1897, opened 2023-03-05
feat: implement jassub instead of libass

ThibaultNocchi:
Blocked by:

* ThaUnknown/jassub#19 - font problem on iOS/iPadOS
* ThaUnknown/jassub#20 - see if we can drop the static copy of the .wasm file

### DIFF
--- a/frontend/src/store/playerElement.ts
+++ b/frontend/src/store/playerElement.ts
@@ -6,14 +6,16 @@
  */
 import { cloneDeep } from 'lodash-es';
 import { nextTick, reactive, watch } from 'vue';
-// @ts-expect-error - No types on libass-wasm
-import SubtitlesOctopus from '@jellyfin/libass-wasm';
-import subtitlesOctopusWorkerUrl from '@jellyfin/libass-wasm/dist/js/subtitles-octopus-worker.js?url';
+import JASSUB from 'jassub';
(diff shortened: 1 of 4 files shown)


frontend/src/store/playerElement.ts
```diff
@@ -94,26 +95,35 @@ class PlayerElementStore {
...
-        prescaleFactor: 0.5,
-        renderAhead: 90
+        workerUrl: jassubWorker,
+        legacyWorkerUrl: jassubLegacyWorker,
+        availableFonts: { 'liberation sans': jassubDefaultFont },
+        libassMemoryLimit: 40,
+        libassGlyphLimit: 40,
+        prescaleFactor: 0.8,
+        maxRenderHeight: 2160,
+        onDemandRender: false
```
ThibaultNocchi, line 113: I need to add a comment to say this value is needed to work on Safari iOS/iPadOS without drawbacks on other browsers

ferferga (approved):

frontend/src/store/playerElement.ts
ThaUnknown: @ThibaultNocchi mind elaborating? to my knowledge onDemandRender has no issues on any browser outside of specifically webOS2 which lies about API support

PR #1899, opened 2023-03-11
refactor: merge fullscreen player and other refactors

ferferga:
* Removed Vuetify hacks around Vuetify 2
* Better back behaviour with Vue router (thanks to @ThibaultNocchi for the original implementation in the video player)
* Middleware for playback pages (which ensures the page doesn't get mounted and destroyed back again. Saves a few CPU cycles and we also don't temporarily see the page after the loading splashscreen is over)
* Merge redirection behaviour of music and video (needed when the user is into a playback page and playback is over) into playerElement store
* Factorize playback buttons into its own components (color doesn't work because a Vuetify bug. Vuetify 3.1.2 works perfectly, but recent versions doesn't. I'm still investigating)
* Minor CSS refactoring
* Decouple the fullscreen music playback page from the audio controls. Resemble a more similar look to Spotify's design ([see here](https://github.com/jellyfin/jellyfin-vue/pull/520#issuecomment-756805747))


frontend/src/assets/styles/global.scss
```diff
@@ -13,13 +13,9 @@
...
-.v-app-bar.v-app-bar--fixed {
-  z-index: 100 !important;
-}
-
-.v-btn {
-  overflow: hidden;
-}
+// .v-app-bar.v-app-bar--fixed {
+//   z-index: 100 !important;
+// }
```
aweebs, line 18: What's the reason for commenting instead of removing? There seems to be a bit of that in this file.

frontend/src/plugins/router/index.ts
```diff
@@ -30,6 +31,19 @@ router.beforeEach(metaGuard);
...
+
+/**
+ * Replaces the 'back' function, taking into account if there's a previous page or not.
+ * If there's no previous page in history, we ensure we want to go home
+ */
+router.back = (): ReturnType<typeof router.back> => {
+  router.replace(
+    typeof router.options.history.state.back === 'string'
+      ? router.options.history.state.back
+      : '/'
```
aweebs, line 44: For my own curiosity, what was the reason this had to be done?

frontend/src/pages/playback/video/index.vue
```diff
@@ -1,10 +1,111 @@
...
+    :class="{ 'cursor-none': !overlay }"
+    @mousemove="handleMouseMove"
+    @touchend="handleMouseMove">
+    <v-overlay
+      v-model="overlay"
+      contained
+      scrim="transparent"
+      width="100%"
+      height="100%">
+      <div
```
aweebs, line 13: Is this a strict copy of OsdPlayer.vue into this file or are there more substantial changes?

frontend/src/store/playerElement.ts
```diff
@@ -178,27 +170,29 @@ class PlayerElementStore {
   };
 
   public constructor() {
+    /**
+     * Move user out of the playback pages when
+     */
```
aweebs, line 175: Looks like you didn't finish the thought ;)

aweebs (commented):
Don't have a lot of context, but overall the changes look good! Just a few questions and comments :)

frontend/src/plugins/router/index.ts
ferferga: @aweebs Try this:

* Go to incognito window (just to be sure we don't mess with cache)
* Access ``google.com``
* Open a new tab and login at ``vite.jf-vue.pages.dev``. Close that tab after you're in the home screen
* Go back to the tab with Google and **replace** the url with ``vite.jf-vue.pages.dev/#/playback/music``
* As no playback will be happening, you'll be redirected back, going to Google instead of the bounds of the client

Repeat the same steps but with the CloudFlare deployment URL of this PR. You'll be redirected back to th home page of the client instead. This is what replacing the function does, strictly staring within the bounds of the client.

frontend/src/pages/playback/video/index.vue
ferferga: @aweebs A few changes:

* Playback buttons use the new refactored components
* Event handling between the page and the component has been removed (no longer necessary in the same page)

But for the rest, it's strictly the same thing.

frontend/src/assets/styles/global.scss
ferferga: @aweebs I originally commented the lines during the initial stages of the migration so I could come back to them quickly if needed, but at this point I agree it's better to start cleaning up the styles and use the git history if it's ever needed again.

Most are hacks to fix bugs in Vuetify 2, and all of them (at least I didn't found any that it isn't) are unnecessary now. I made a complete task for cleaning them up in the Vue 3 PR.

TLDR: Will remove them 👍🏻.

frontend/src/components/Buttons/Playback/RepeatButton.vue
```diff
@@ -0,0 +1,26 @@
...
+const props = defineProps<{ size?: string }>();
+const playbackManager = playbackManagerStore();
+
+const repeatIcon = computed(() => {
+  if (playbackManager.repeatMode === RepeatMode.RepeatOne) {
+    return IMdiRepeatOnce;
+  }
+
+  return IMdiRepeat;
+});
```
ThibaultNocchi, line 25:
```suggestion
const repeatIcon = computed(() => playbackManager.repeatMode === RepeatMode.RepeatOne ? IMdiRepeatOnce : IMdiRepeat);
```

I guess?

ThibaultNocchi (commented):
Also, it seems like that the page height is now way too big: you can scroll way below the last items on the home page.
Screenshot 2023-03-12 at 17-16-46 Home Jellyfin Vue

ferferga:
@ThibaultNocchi @aweebs Although the fixes of this PR are really relevant (conversely the key-changing fix for route paths at RouterViewTransition, which is necessary before #1892 goes in as all the watchers we have in pages might be unnecessary now), #1895 must go first since there are a lot of good fixes too that this PR depends upon and it's almost ready to be merged.

#1895 has all the requested changes, just needs a final look. This one also has the page height issue fixed and most concerns answered, so even after my rebase on top of #1895, getting this merged will be really straightforward afterwards.

ferferga:
@ThibaultNocchi @aweebs Although the fixes of this PR are really relevant (conversely the key-changing fix for route paths at RouterViewTransition, which is necessary before #1892 goes in as all the watchers we have in pages might be unnecessary now), #1895 must go first since there are a lot of good fixes too that this PR depends upon and it's almost ready to be merged.

#1895 has all the requested changes, just needs a final look. This one also has the page height issue fixed and most concerns answered, so even after my rebase on top of #1895, getting this merged will be really straightforward afterwards.

ferferga:
@ThibaultNocchi @aweebs Although the fixes of this PR are really relevant (conversely the key-changing fix for route paths at RouterViewTransition, which is necessary before #1892 goes in as all the watchers we have in pages might be unnecessary now), #1895 must go first since there are a lot of good fixes too that this PR depends upon and it's almost ready to be merged.

#1895 has all the requested changes, just needs a final look. This one also has the page height issue fixed and most concerns answered, so even after my rebase on top of #1895, getting this merged will be really straightforward afterwards.

ferferga:
Button colors has been fixed in Vuetify 3.1.9 (which has been pushed as part of this PR)

PR #1900, opened 2023-03-15
show linear progress when making network requests

ferferga:
Show a v-progress-linear when an Axios request is happening or suspense is loading content.

In a similar fashion to GitHub's and YouTube loading indicators

### DIFF
--- a/frontend/src/components/Layout/AppBar/AppBar.vue
+++ b/frontend/src/components/Layout/AppBar/AppBar.vue
@@ -3,6 +3,7 @@
     :class="useResponsiveClasses('app-bar-safe-zone')"
     :elevation="transparentAppBar ? 0 : 3"
     :color="transparentAppBar ? 'transparent' : undefined">
+    <loading-indicator />
     <v-app-bar-nav-icon
       v-if="$vuetify.display.mobile"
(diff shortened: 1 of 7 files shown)


ThibaultNocchi (commented):
Code looks good to me, but the progress bar stutters _a lot_ when loading for instance a library. I guess that's just when the view rendering stalls, so that's on our views and not this bar.

ferferga:
@ThibaultNocchi With the circular loader at the chip happens the same thing, even in Vue 2. It keeps spinning but without resizing for a glimpse (maybe because the spinning animation is less taxing in the engine so it cuts frames in the resizing part)

It's true the linear progress maybe stutters a bit more (perhaps its animations are more complex?) but I think it's good anyway in most places.

Fixing the performance hit on that page is going to require server pagination, since last time I debugged it, the processing itself of the request + item store logic to add them are expensive (virtual grid is too, but I think in a less meaningful way and probably something that can be improved by an animationFrame. The problem is that we never tried to cut the others).

ThibaultNocchi:
@ferferga I think the PR is good, it looks great on pages and the stutter isn't really problematic. Also, it's not due to the progress bar so LGTM

ThibaultNocchi (approved):

PR #1903, opened 2023-03-18
fix(icons): unset scale to see icons on iOS and avoid overflow

### LINKED ISSUE #1902: [vite] Icons not visible on iOS
### Description of the bug

Since 85eb2e0ec412b45e809af85a6a0dc660c8241ca8, having set up the scale to `0` makes the icons invisible on iOS. As this was done to fix overflows problems, we need to discuss how to fix it.

### Steps to reproduce

Run a `vite` branch on an iPhone.

### Expected behavior

See the icons

### Logs

_No response_

### Screenshots

_No response_

### Platform

iOS

### Browser

Safari

### Jellyfin server version

10.8.9

### Additional context

_No response_

ThibaultNocchi:
fix #1902

---

Unsetting the scale makes the icons work on iOS and overflow are still gone. This also removes the manual setting.


ferferga:
@ThibaultNocchi already tried this (as the default config is the same as not defining the key in the Vite config)

But this:



VS master:



Even if we can workaround it, I don't like this and I think we must keep this to an unset size because:

* It will be another reason to do the CSS hacks we had been doing as of today, where we have a lot of stuff with arbitrary margins and paddings. Thankfully we're removing those since Vuetify removed the need to those, but we don't want another source. 

* Having SVG icons defined with a size beforehand it's like having a PNG. Completely kills the reason to have SVG in the first place. Even if it could be overridden with styles for it or the parent, it's hacky as I said before. 

This needs a deeper investigation, in my opinion

PR #1905, opened 2023-03-20
More refactors and cleanups

aweebs:
- Refactor the store-sync code into smaller functions which can be composed better, even if only a few are used
- Adds much more explicit checks to satisfy static type checking for WebSocketMessage handling
- Added a parser for verifying the external config is the expected shape
- Add an eslint rule to ban `@ts-` directives, and replaces them with casts. This is minimally safer but prevents multiple unexpected errors from being hidden on a line.
- Upgrades typescript dependency to version 5
- A bunch of minor typefixes and teaks. This resolves almost every non-vuetify-bad-typing error.

Recommend reviewing commit by commit and with whitespace visualization disabled.

### DIFF
--- a/frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
+++ b/frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
@@ -8,8 +8,7 @@
       <tooltip-button
         class="align-self-center"
         v-bind="menu"
-        :tooltip="{ text: $t('playbackSettings'), location: 'top' }"
-        :icon="{ icon: true }">
+        :tooltip="{ text: $t('playbackSettings'), location: 'top' }">
         <v-icon>
           <i-mdi-cog />
         </v-icon>
--- a/frontend/src/components/Layout/Images/__tests__/BlurhashCanvas.spec.ts
+++ b/frontend/src/components/Layout/Images/__tests__/BlurhashCanvas.spec.ts
@@ -14,35 +14,26 @@ beforeEach(() => {
 
 describe('component: BlurhashCanvas', () => {
   it('has a hash property of type String', () => {
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.hash).toBeDefined();
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.hash.type).toBe(String);
   });
 
   it('has a width property of type Number', () => {
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.width).toBeDefined();
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.width.type).toBe(Number);
   });
 
   it('has a height property of type Number', () => {
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.height).toBeDefined();
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.height.type).toBe(Number);
   });
 
   it('has a punch property of type Number', () => {
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.punch).toBeDefined();
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.punch.type).toBe(Number);
   });
 
   it('requires a hash', () => {
-    // @ts-expect-error - props are not defined in the Vue type.
     expect(wrapper.vm.$options.props.hash.required).toBeTruthy();
   });
 
--- a/frontend/src/pages/index.vue
+++ b/frontend/src/pages/index.vue
@@ -21,19 +21,11 @@
 
 <script setup lang="ts">
 import { computed, onMounted } from 'vue';
-import { pickBy } from 'lodash-es';
 import { useI18n } from 'vue-i18n';
 import { useRoute } from 'vue-router';
 import { CardShapes, getShapeFromCollectionType } from '@/utils/items';
-import { clientSettingsStore, userLibrariesStore } from '@/store';
+import { userLibrariesStore } from '@/store';
 import type { HomeSection } from '@/store/userLibraries';
-
-const VALID_SECTIONS = new Set([
-  'resume',
-  'resumeaudio',
-  'upnext',
-  'latestmedia'
-]);
 
 const { t } = useI18n();
 const route = useRoute();
@@ -42,7 +34,6 @@ route.meta.title = t('home');
 route.meta.transparentLayout = true;
 
 const userLibraries = userLibrariesStore();
-const clientSettings = clientSettingsStore();
 
 /**
  * Items are fetched at logon when switching between the 'fullpage' and 'default' layout. With the help of the
@@ -62,29 +53,13 @@ onMounted(async () => {
 });
 
 const homeSections = computed<HomeSection[]>(() => {
-  // Filter for valid sections in Jellyfin Vue
-  // TODO: Implement custom section order
-  const homeSectionPrefs = pickBy(
-    // @ts-expect-error - No typings for this
-    clientSettings.CustomPrefs as { [key: string]: string },
-    (value, key) => {
-      return (
-        value && VALID_SECTIONS.has(value) && key.startsWith('homesection')
-      );
-    }
-  );
-
-  const homeSectionKeys = Object.keys(homeSectionPrefs);
-
-  if (homeSectionKeys.length === 0) {
-    homeSectionKeys.push(
-      'librarytiles',
-      'resume',
-      'resumeaudio',
-      'upnext',
-      'latestmedia'
-    );
-  }
+  const homeSectionKeys = [
+    'librarytiles',
+    'resume',
+    'resumeaudio',
+    'upnext',
+    'latestmedia'
+  ];
 
   const homeSections: HomeSection[] = [];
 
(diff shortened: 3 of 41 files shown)


frontend/src/utils/store-sync.ts
```diff
@@ -1,161 +1,206 @@
...
-): Promise<T | DisplayPreferencesDto> {
+export async function fetchDisplayPreferences(
+  displayPreferencesId: string
+): Promise<DisplayPreferencesDto> {
   const remote = useRemote();
 
+  if (!remote.auth.currentUserId) {
+    throw new Error(USER_UNAUTHENTICATED_MESSAGE);
+  }
+
```
aweebs, line 40: This check isn't really done anywhere else, we usually just default the currentUserId if it's null, so I'll likely just do that instead of this error check everywhere.

frontend/src/utils/external-config.ts
```diff
@@ -1,22 +1,57 @@
...
 /**
  * Fetch configuration at runtime from the config.json file
  * We use destr for serialization as it has better support for JS primitives.
  */
 export async function getJSONConfig(): Promise<ExternalJSONConfig> {
   if (isNil(externalConfig)) {
-    externalConfig = destr(
-      JSON.stringify(await (await fetch('/config.json')).json())
-    ) as ExternalJSONConfig;
+    const loadedConfig = await (await fetch('/config.json')).json();
```
aweebs, line 50: `fetch` would have already tried to parse the response with the regular JSON parser and not `destr`, so this second stringify and parsing was pointless.

frontend/src/utils/playback-profiles/helpers/codec-profiles.ts
```diff
@@ -26,7 +26,17 @@ import {
...
+    isTizen() &&
+    'webapis' in window &&
+    typeof window.webapis === 'object' &&
+    window.webapis &&
+    'productinfo' in window.webapis &&
+    typeof window.webapis.productinfo === 'object' &&
+    window.webapis.productinfo &&
+    'isUdPanelSupported' in window.webapis.productinfo &&
+    typeof window.webapis.productinfo.isUdPanelSupported === 'function'
+  ) {
```
ferferga, line 39: FYI, our idea is to move completely out from our custom profile detections and use MediaCapabilities API, so this file is indeed deprecated.

You can coordinate with @ThibaultNocchi, because it will be a huge step forward better playback and this is the major legacy thing we have in this client as of today.

frontend/src/utils/store-sync.ts
```diff
@@ -1,161 +1,206 @@
 import { DisplayPreferencesDto } from '@jellyfin/sdk/lib/generated-client';
 import { getDisplayPreferencesApi } from '@jellyfin/sdk/lib/utils/api/display-preferences-api';
 import destr from 'destr';
-import { isNil } from 'lodash-es';
 import { usei18n, useRemote, useSnackbar } from '@/composables';
 import { taskManagerStore } from '@/store';
 
+const CLIENT = 'vue';
+
+const USER_UNAUTHENTICATED_MESSAGE = 'Expected user to be authenticated';
```
ferferga, line 9:
```suggestion
const CLIENT = 'vue';
const USER_UNAUTHENTICATED_MESSAGE = 'Expected user to be authenticated';
```

frontend/src/utils/store-sync.ts
```diff
@@ -1,161 +1,206 @@
...
- * Pushes a new displayPreferences payload to server
+ * Uses the keys on `defaults` to extract values from the Server's CustomPrefs
+ * and de-serializes them to a value.
+ * All keys needed for default should exist on the defaults parameter.
+ * Warning: No runtime checking is performed and de-serialized data could vary in shape from what is expected.
  */
-export async function pushSettingsToServer(
-  key: string,
-  prefs: DisplayPreferencesDto
+export async function fetchDefaultedCustomPrefs<T extends object>(
```
ferferga, line 122: It's not clear to me yet why the function rename is needed

ferferga (changes requested):

frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
```diff
@@ -8,8 +8,7 @@
       <tooltip-button
```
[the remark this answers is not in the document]
aweebs: Agree with this since the component is really basic and doesn't have any real special/custom implementation. If we're all aligned on that being the way forward I can add it to this PR or do it separately. Will wait until we have consensus.

frontend/src/components/Layout/Images/__tests__/BlurhashCanvas.spec.ts
```diff
@@ -14,35 +14,26 @@ beforeEach(() => {
 
 describe('component: BlurhashCanvas', () => {
   it('has a hash property of type String', () => {
-    // @ts-expect-error - props are not defined in the Vue type.
```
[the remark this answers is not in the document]
aweebs: Yeah, I did it more for completeness :D

Agree that we should have a `tests` directory alongside `src`. Do you want that part of this PR?

frontend/src/pages/index.vue
```diff
@@ -62,29 +53,13 @@ onMounted(async () => {
...
-      'latestmedia'
-    );
-  }
+  const homeSectionKeys = [
+    'librarytiles',
+    'resume',
+    'resumeaudio',
+    'upnext',
+    'latestmedia'
+  ];
```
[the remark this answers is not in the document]
aweebs: It does use some reactive values from userLibraries on line 88

frontend/src/utils/playback-profiles/helpers/codec-profiles.ts
aweebs: Do you want to coordinate removing it in this PR or for future work?

frontend/src/utils/external-config.ts
```diff
@@ -1,22 +1,57 @@
...
+
+  if (
+    !('defaultServerURLs' in config) ||
+    !Array.isArray(config.defaultServerURLs)
+  ) {
+    throw new Error('Expected defaultServerURLS array');
+  }
+
+  if (
+    config.defaultServerURLs.some(
```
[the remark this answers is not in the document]
aweebs: Ah, this should be some but the inner check is wrong, and should check if any of them _are not_ strings. Good catch!

frontend/src/store/clientSettings.ts
```diff
@@ -143,7 +143,7 @@ class ClientSettingsStore {
       async () => {
         if (remote.auth.currentUser) {
           try {
-            const data = await fetchSettingsFromServer<ClientSettingsState>(
+            const data = await fetchDefaultedCustomPrefs(
```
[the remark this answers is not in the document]
aweebs: The rename was meant to be more clear about what it does (obviously that didn't work 😅 ).
Overall functionality is the same to how it worked in practice previously. Given a default object, it takes the properties from the default object and checks if they exist in CustomPrefs. If a property does exist in CustomPrefs, then the deserialized version of the value from CustomPrefs is used for that property. If the property does not exist in CustomPrefs, the default value is used for that property.

frontend/src/components/Layout/Images/__tests__/BlurhashCanvas.spec.ts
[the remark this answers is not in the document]
ferferga: @aweebs Yes please :)

frontend/src/utils/playback-profiles/helpers/codec-profiles.ts
ferferga: @aweebs Future work, but I would rule modifying this file altogether in this PR, it's indeed introducing more type errors (even after applying the Vuetify 3.1.11 update to this branch).

In my discussions with @ThibaultNocchi, we thought the best approach for MediaCapabilities would be to phase out the "generate a device profile of all codecs at startup" approach and move to a "test for compatibility before starting the playback", based on the codec information of the selected media streams given by the server. We should test for video and audio independently and report as compatible all the web-based subtitlte formats (WebVTT) + ASS (thanks to JASSUB).

If a file is not compatible, we should report to server that we're compatible with H264 and AAC, which is a good default combo that all browsers supports. There's no point in checking for MP3, H263, etc... Maybe we can also test for H265 as Edge on Windows and Safari supports it.

That should be done exclusively in a new PR, as removing all this logic will break playback.

frontend/src/utils/external-config.ts
ferferga: @aweebs Fair enough!

ferferga:
@aweebs This has more type errors than current vite (this branch sits at 36 vs 20 at vite), even after the Vuetify upgrade.
Please, revisit this.

frontend/src/utils/playback-profiles/helpers/codec-profiles.ts
aweebs: The added type errors are because of an outdated version of typescript. I realized I had been using 5.x.x since it was announced recently. Even upgrading to a later version of 4.x.x resolves these. I'll add the typescript update to this PR, it doesn't seem to impact anything other than our typing.

I would like to get rid of this type error though, the other cheating way to get rid of this is with an any cast
```
  if (isTizen()) {
    isTizenFhd = !(window as any).webapis.productinfo.isUdPanelSupported();
  }
```

I'd prefer to keep the above now that it won't cause additional type errors. Even if it's very verbose, we're pretty close to getting of all of them :D

aweebs:
> @aweebs This has more type errors than current vite (this branch sits at 36 vs 20 at vite), even after the Vuetify upgrade. Please, revisit this.

Type Script upgrade was all that's needed 🎉
```
Found 6 errors in 5 files.

Errors  Files
     1  src/components/Layout/Carousel/Item/ItemsCarouselTitle.vue:15
     2  src/components/Layout/Carousel/Item/ItemsCarousel.vue:37
     1  src/plugins/remote/auth/index.ts:109
     1  src/utils/images.ts:198
     1  src/utils/items.ts:303
```

ferferga (approved):

PR #1906, opened 2023-03-22
fix: remove workaround for Safari's iOS fullscreen

ferferga:
* After the merge of https://github.com/vueuse/vueuse/pull/2822, this is implemented upstream


ThibaultNocchi (approved):
Tested on iOS and iPad ✅

PR #1908, opened 2023-03-23
Remove old tests

aweebs:
Initially based on https://github.com/jellyfin/jellyfin-vue/pull/1905#discussion_r1141900209

But consensus is these are really old and not useful so there's no point in keeping them around at all.


ferferga:
> These may just be worth removing at this point too 🤷

That's what I've been aiming at since the beginning if you remember, I think there's little we can reuse.

ferferga:
@aweebs Yes, also removing the ignores in tsconfig and eslint.

aweebs:
@ferferga done 🚀

ferferga (approved):

PR #1909, opened 2023-03-23
fix: make queue sortable again

ferferga:
Vuetify's virtual scroller didn't update items properly when reordered, so we're not doing any virtual scrolling trickery for now. I will come up with a custom solution once Vue 3 is merged to master.


ThibaultNocchi (approved):
LGTM

PR #1911, opened 2023-03-26
fix: unplayed chips UI readable and globally rounded chips

ThibaultNocchi:
Before the PR and after

print_screen_2023-03-26-15-48-27

print_screen_2023-03-26-15-48-04


ferferga:
@ThibaultNocchi Did you check any other chips that might benefit from this, throughout the client

ThibaultNocchi:
@ferferga Indeed, I just checked them and seen that the rounded variant is better so I applied it globally. Also, only the unplayed chip needed the elevated variant, the others look good.

PR #1912, opened 2023-03-26
More small cleanups

aweebs:
1. Remove TooltipButton in favour of using the components directly with parent activator (review with whitespace changes off)
2. Support `data-` attributes on any vue components. These get mapped through into the component template root element correctly but were giving typing errors previously.
3. Make favicon and icon paths absolute. Previously when using the `history` router type these would not work correctly unless the user started on the home page.
4. Fix typing issues in `getImageInfo`
5. Remove the `tag` field from being returned as part of image info since it wasn't actually used anywhere and seemed incorrect to return.
6. Convert all uses of activator slots for menus to using the parent activator pattern.
7. Run lint fix on `use-fullscreen`. What happened with this file? The code style was completely different than what exists in the code base now and it introduced a ton of linting errors, many of which could not be auto fixed.

Getting close!
```
Found 2 errors in 2 files.

Errors  Files
     1  src/plugins/remote/auth/index.ts:109
     1  src/utils/items.ts:303
```


frontend/package.json
```diff
@@ -33,7 +33,6 @@
     "all-iso-language-codes": "1.0.12",
     "axios": "1.3.4",
     "blurhash": "1.1.5",
-    "comlink": "4.3.1",
```
ferferga, line 36: Why? Build is failing because of this

ferferga (changes requested):

aweebs:
Removed the last commit which removed comlink - my mistake!

ferferga:
I was thinking that the absolute path doesn't fix the problem properly (so maybe I should revert the change for the ``config.json`` as well).

The absolute paths doesn't handle cases where I want the client to be served in a subpath: ``domain.xyz/vue``

ferferga (changes requested):
Not all menus and tooltips have been changed to the new syntax.
There are also new lint errors introduced in this PR.

ferferga (changes requested):
Besides my two comments, I just noticed we could set the activators by default : https://vuetifyjs.com/en/features/global-configuration/#contextual-defaults
Can you ammend your commits to take that into account?

Sorry mate for not thinking about this earlier 😅

ferferga (approved):

PR #1913, opened 2023-03-28
feat: permanent scrollbars in default layout only

ferferga:
The splashcreen and video/music/login pages with scrollbars were really ugly

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="no-forced-scrollbar">
   <head>
     <meta charset="UTF-8">
     <link rel="icon" type="image/x-icon" href="/favicon.ico">
--- a/frontend/src/assets/styles/global.scss
+++ b/frontend/src/assets/styles/global.scss
@@ -11,6 +11,10 @@
   background: #666;
 }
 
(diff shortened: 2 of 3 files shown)


ThibaultNocchi (commented):
Maybe we should add the same class logic to the server layout?

ferferga:
@ThibaultNocchi I had the same doubt.

I don't expect that page to be scrollable too often (maybe if an user has tons and tons of servers?) and scroll itself will still be possible (so a minor jump is not as noticeable?)

The problem is when the page is constantly jumping because of the scrollbar

ThibaultNocchi:
Ok just a personal thing. I'd reverse the logic: don't apply the class by default and only apply it on the FS pages. To me it's more logic as those are the "edge" cases, not the default. But that's a preference.

ferferga:
@ThibaultNocchi Thought the same when doing this, so our thinking process is very similar :)

However, doing that requires more lines of code (to cover server and fullpage layout) plus I can also give the upside down to your statement: What you say is to keep Vuetify's defaults as our first defaults, while what I want to is to stick with browser's defaults first (removing the class that overrides Vuetify's default when accessing our default layout, and doing the reverse thing when going out)

ThibaultNocchi (approved):
LGTM

PR #1914, opened 2023-03-31
chore: post-Vue3 merge README

ferferga:
* Make the description framework agnostic, so we don't need to update it in case we switch our desktop app bundler (for example)
* Add better description for end users. We were too developer centric
* Add links to Wiki for all the pieces that are more susceptible of evolving
* Add some clarification about how jellyfin-vue works (add a note about jellyfin-web too) in regards to privacy, since by what I read on Reddit, many people assume the connections are not direct between their devices and the target Jellyfin server.
* Made that privacy note really agnostic for how the client might be served/distributed in the future (but hopefully I did it in a clear way)
* Clarify the "Getting started" and "Running a production build sections"

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,49 +5,127 @@
 
 <p align="center">
 <img alt="Logo Banner" src="https://raw.githubusercontent.com/jellyfin/jellyfin-ux/master/branding/SVG/banner-logo-solid.svg?sanitize=true"/>
-<br/>
-<br/>
 <a href="https://hub.docker.com/r/jellyfin/jellyfin-vue">
 <img alt="Docker Pulls" src="https://img.shields.io/docker/pulls/jellyfin/jellyfin-vue">
 </a>
 <a href="https://github.com/jellyfin/jellyfin-vue/pkgs/container/jellyfin-vue">
 <img alt="GHCR images" src="https://img.shields.io/badge/Github-packages-blue">
 </a>
-<br/>
 <a href="https://github.com/jellyfin/jellyfin-vue">
 <img alt="GPL 3.0 License" src="https://img.shields.io/github/license/jellyfin/jellyfin-vue.svg"/>
+</a>
+<a href="https://translate.jellyfin.org/projects/jellyfin-vue/jellyfin-vue/">
+<img src="https://translate.jellyfin.org/widgets/jellyfin-vue/-/jellyfin-vue/svg-badge.svg">
 </a>
 <a href="https://github.com/jellyfin/jellyfin-vue/releases">
 <img alt="Current Release" src="https://img.shields.io/github/release/jellyfin/jellyfin-vue.svg"/>
 </a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,13 +41,62 @@
 </a>
 </p>
 
-This is an experimental web client for Jellyfin based on Vue.js. We welcome all contributions and pull requests! If you have a larger feature in mind please open an issue so we can discuss the implementation before you start.
+This is an alternative client for Jellyfin based on Vue.js. It's not feature complete and it's constantly evolving.
 
-<p align="center">
-    <a href="https://jf-vue.pages.dev/">View demo</a>
-</p>
+We welcome all contributions and pull requests! If you have a larger feature in mind please open a discussion or chat with us in Matrix so we can discuss the implementation before you start.
```
thornbill, line 46: Maybe add a link to matrix?

README.md
```diff
@@ -97,25 +141,21 @@ This is an experimental web client for Jellyfin based on Vue.js. We welcome all
...
-npm run prod
+npm run serve
 ```
 
-_Although the build of the client is production-ready, Nuxt's own HTTP server should never be exposed directly to the internet and a proper hosting tool like Nginx should always be used instead._
-
-## Running Docker Container
-
-You can run the Docker container with the following command, replacing 8080 with the port of your choice:
+There's also an shortcut for building the production version of the client and serving it afterwards:
```
thornbill, line 150:
```suggestion
There's also a shortcut for building the production version of the client and serving it afterwards:
```

README.md
ferferga: @thornbill The badge is already a link. Do you have any other suggestion so I can make it more clear that's whrere users must click?

joshuaboniface: I think if there is a link in a badge that's fine.

README.md
```diff
@@ -5,49 +5,127 @@
...
+
+## Docker Container 📦
+
+You can run the Docker container with the following command, replacing 8080 with the port of your choice:
+
+```bash
+docker run -d -p 8080:80 ghcr.io/jellyfin/jellyfin-vue
+```
+
+You can check [Packages](https://github.com/jellyfin/jellyfin-vue/pkgs/container/jellyfin-vue) (recommended) or [DockerHub](https://hub.docker.com/r/jellyfin/jellyfin-vue) for all the tagged images.
```
ThibaultNocchi, line 91:
```suggestion
You can check [GHCR](https://github.com/jellyfin/jellyfin-vue/pkgs/container/jellyfin-vue) (recommended) or [DockerHub](https://hub.docker.com/r/jellyfin/jellyfin-vue) for all the tagged images.
```

I've never seen the GH repo being called that way, maybe GHCR isn't the best but I don't like Packages

ThibaultNocchi (commented):
Nitpicking

PR #1915, opened 2023-03-31
fix: nextup can have undefined parent id

ThibaultNocchi:
* nextUp was called for each library, while the function itself overwrote the whole state. This resulted in nextUp being populated by the latest library fetched. I reduced it to a single call cause removing the parentId gets us the classic nextUp experience from web.
* Wait the refresh promises at the end of the function, as we don't have to wait for them in the body of the function

### DIFF
--- a/frontend/src/store/userLibraries.ts
+++ b/frontend/src/store/userLibraries.ts
@@ -124,7 +124,7 @@ class UserLibrariesStore {
     useSnackbar(t('errors.anErrorHappened'), 'error');
   };
 
-  private _fetchUserViews = async (): Promise<void> => {
+  private _updateUserViews = async (): Promise<void> => {
     const remote = useRemote();
 
     try {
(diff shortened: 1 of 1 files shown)


frontend/src/store/userLibraries.ts
```diff
@@ -271,18 +272,26 @@ class UserLibrariesStore {
...
   public refresh = async (): Promise<void> => {
-    await this._fetchUserViews();
-    await this._fetchAudioResumes();
-    await this._fetchVideoResumes();
-    await this._fetchIndexCarouselItems();
+    const promises: Promise<void>[] = [];
+
+    this._state.value.homeSections.upNext = [];
+
+    promises.push(
```
ferferga, line 279: Why not add them directly to the declared array instead of pushing?

ferferga (changes requested):
As we discussed previously, the `fetchX` naming it's unfortunate, since the function in reality fetches **and** sets the state. Can you change it to `updateX` or something like that, so it's more clear?

Also, I think we can remove a lot of checks in the functions and replace them with the `this._state.X = fetchedX ?? []`

ThibaultNocchi:
@ferferga all of these are now done :)

ferferga (approved):

PR #1917, opened 2023-03-31
refactor(vue3): search

ferferga:
* Memoize previous search queries
* Use computed for data filters
* Fix old results temporarily appearing when clearing the query


ThibaultNocchi (approved):
LGTM

PR #1928, opened 2023-04-05
Some more and cleanups but also fix the app bar and search sizing

aweebs:
It looks good again! No more gaps!

image

Sidenote, the elevation property on the app bar applies a drop shadow which makes it look like there is a tiny gap between the bars when there are multiple, but it's much clearer in light mode that's a drop shadow. We could remove that entirely.


ThibaultNocchi:
I don't mind you removing the elevated attribute in this PR :)

frontend/src/components/Layout/AppBar/AppBar.vue
```diff
@@ -1,8 +1,10 @@
 <template>
   <v-app-bar
-    :class="useResponsiveClasses('app-bar-safe-zone')"
+    class="app-bar-safe-zone"
     :elevation="transparentAppBar ? 0 : 3"
-    :color="transparentAppBar ? 'transparent' : undefined">
+    :color="transparentAppBar ? 'transparent' : undefined"
+    density="compact"
+    flat>
```
ferferga, line 7: Why?

ferferga (changes requested):
The real fix for the elevation would be to use this: https://vuetifyjs.com/en/api/v-app-bar/#slots-extension

This would also allow us to remove a lot of CSS hacks that we currently have, which also borks the transitions as of now, iirc.
The problem is that last time I tried we don't have a nice way for teleporting the extended app bar's content into the extension slot. I didn't try it much, so if you can give it another look please.

Vue team is, as far as I know, improving teleport (by including utils like vue-safe-teleport natively) so that would be much easier in the future

frontend/src/components/Layout/AppBar/AppBar.vue
aweebs: The flat property removes the default drop shadow (which displays as if it had `elevation` of 4. This won't matter until we remove the elevation property though. I'm going to look into using the app-bar extension as you pointed out though.

aweebs:
I started looking into using the app bar extension and portal and also didn't get very far. Adding the app bar extension slot causes the height to shift regardless if content is rendered in it, and even then portaling to it requires a dom entry, so I'm not sure there's a great solution available right now. I've removed the drop shadows by applying the `flat` property in the meantime and it looks good now.

ThibaultNocchi (changes requested):
You remove the different padding classes styles, but they are used around the app. So either the concerned parts look the same without it and you can remove the classes on the tags, or you can't :)

aweebs:
> You remove the different padding classes styles, but they are used around the app. So either the concerned parts look the same without it and you can remove the classes on the tags, or you can't :)

I decided to remove that commit. I'm less sure about these changes after thinking on it for a bit, and before doing that I'd like  find out which devices use those `env` values and see the path through an emulator rather than potentially break the UI for someone's device.

ThibaultNocchi (approved):
LGTM, could you rebase the branch please? :)

ferferga (changes requested):
Fix the lint error as well, please

PR #1930, opened 2023-04-06
fix: locale with dashes in it used to get a locale name didn't work + linter errors + no negative conditions

ThibaultNocchi:


ThibaultNocchi:
@ferferga this is fixed :)

frontend/src/store/playbackManager.ts
```diff
@@ -540,7 +540,12 @@ class PlaybackManagerStore {
    * Updates mediasession metadata based on the currently playing item
    */
   private _updateMediaSessionMetadata = (): void => {
-    if (!isNil(this.currentItem)) {
+    if (isNil(this.currentItem)) {
```
ferferga, line 543: I think this whole block can be removed and just leave what is in the `else` part? This seems redundant to me.

ThibaultNocchi: Indeed, refactored

ferferga (approved):

PR #1932, opened 2023-04-07
fix: version selector not working

### LINKED ISSUE #1931: Version Selector is not working
### Description of the bug

Version selector for multiple versions available movies are not working. Right now available versions are showing as a single line.

### Steps to reproduce

1. Make sure you have a movie/episode with multiple versions
2. Try to select second version

### Expected behavior

Version selector should work and media should be able to switch

### Logs

_No response_

### Screenshots

Screenshot from 2023-04-07 11-50-36

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.8.9

### Additional context

_No response_

HazilMohamed:
Fixes #1931

NB: One problem I can see here is we are not resetting index of audio, video and subtitles. So, there might be a chance to have empty values while switching across versions here. I'm fairly new to vue, If you have any suggestion on how can I fix that, please let me know :sweat_smile:

### DIFF
--- a/frontend/src/pages/item/_itemId/index.vue
+++ b/frontend/src/pages/item/_itemId/index.vue
@@ -145,6 +145,7 @@
                 <v-col class="px-0" :cols="12" :sm="10">
                   <media-stream-selector
                     v-if="currentSource.MediaStreams"
+                    :key="currentSource.Id || ''"
                     :media-streams="
                       getMediaStreams(currentSource.MediaStreams, 'Video')
                     "
@@ -162,6 +163,7 @@
                 <v-col class="px-0" :cols="12" :sm="10">
                   <media-stream-selector
                     v-if="currentSource.MediaStreams"
+                    :key="currentSource.Id || ''"
                     :media-streams="
                       getMediaStreams(currentSource.MediaStreams, 'Audio')
                     "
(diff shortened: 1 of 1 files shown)


ThibaultNocchi:
Good job, I think I know how to fix the other selects.

As they pick the default value when the component `media-stream-selector` is created, the easiest way to make them pick the new default is by recreating them. By chance, the `key` prop (https://vuejs.org/api/built-in-special-attributes.html#key) makes a component re-create when its value is changed:

> It can also be used to force replacement of an element/component instead of reusing it.

I tried it locally, and it works great. So best way is to use `currentSource.Id || ''` as a key on the 3 `media-stream-selector`. :)

HazilMohamed:
@ThibaultNocchi yup that solves that issue. updated the PR too

ThibaultNocchi:
It seems like fixing the selector is great, but it doesn't play the selected version though, only the default one. Wanna look into it or we merge your PR and I check that?

HazilMohamed:
You can take it up if you want. I'll go through the code and try to figure out what's wrong

ThibaultNocchi:
Ok, it seems that it's a little bit more than just passing the value somewhere, our current logic doesn't take into account media sources: https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/store/playbackManager.ts#L1117

I'll just merge your PR and quickly implement it. :)

ThibaultNocchi (approved):

PR #1933, opened 2023-04-07
fix: properly disconnect user when server is unreachable

### LINKED ISSUE #1929: Can't reset client when auto-reconnecting to a user with a server not responding
### Description of the bug

If you load the page with a user already connected to a server not responding, there's no way to go to the "add server" page. You can log out (with a bad UI though) and get stuck in the user picker page from the unresponsive server. And there, you can manually go to /server/add by changing the URL, but that's all.

### Steps to reproduce

- Login with a user with the "remember me" set to ON
- Close the page
- Stops the JF server
- Opens the page

### Expected behavior

It should logout the user and redirect to the /server/add page.

### Logs

_No response_

### Screenshots

_No response_

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.8.9

### Additional context

_No response_

ThibaultNocchi:
fix #1929

### DIFF
--- a/frontend/src/pages/server/login.vue
+++ b/frontend/src/pages/server/login.vue
@@ -68,18 +68,26 @@ import { isEmpty } from 'lodash-es';
 import { UserDto } from '@jellyfin/sdk/lib/generated-client';
 import { getBrandingApi } from '@jellyfin/sdk/lib/utils/api/branding-api';
 import { getUserApi } from '@jellyfin/sdk/lib/utils/api/user-api';
-import { useRoute } from 'vue-router';
+import { getSystemApi } from '@jellyfin/sdk/lib/utils/api/system-api';
+import { useRoute, useRouter } from 'vue-router';
 import { useI18n } from 'vue-i18n';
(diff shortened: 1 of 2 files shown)


frontend/src/pages/server/login.vue
```diff
@@ -68,19 +68,27 @@ import { isEmpty } from 'lodash-es';
...
 const route = useRoute();
+const router = useRouter();
 const remote = useRemote();
 const api = remote.sdk.oneTimeSetup(
   remote.auth.currentServer?.PublicAddress || ''
 );
 
 route.meta.title = t('login.login');
 
+try {
```
ferferga, line 86: Why is this needed here? Doesn't the interceptor catches this already?

ThibaultNocchi: Nope, cause in this case it doesn't get a 401, just a NET_ETROR from Axios, hence the interceptor doing nothing and just throwing the error.

ferferga (approved):

PR #1936, opened 2023-04-07
feat: refresh user on page load

ThibaultNocchi:


ferferga (changes requested):
Why not do this in the plugin initialization? In the auth plugin constructor, for instance.

ThibaultNocchi:
@ferferga It's implemented now in the constructor :)

frontend/src/plugins/remote/auth/index.ts
```diff
@@ -197,6 +198,22 @@ class RemotePluginAuth {
...
 
+  /**
+   * Refreshes the current user infos, to fetch a new picture for instance
+   */
+  public async refreshCurrentUserInfo(): Promise<void> {
+    if (!isNil(this.currentUser) && !isNil(this.currentServer)) {
+      const api = useOneTimeAPI(
+        this.currentServer.PublicAddress,
+        this.getUserAccessToken(this.currentUser)
+      );
```
ferferga, line 209: If there's an user logged in, you don't need the `OneTimeApi`, you can use the global api instance. In fact if there's no global api defined, we can assume there's no current user/server. Check [this](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/plugins/remote/sdk/index.ts#L17-L51)

ferferga (changes requested):
This made me realize that we could probably simplify our checks in a lot of places, but that's for another time lol

frontend/src/plugins/remote/auth/index.ts
ferferga: Fair enough, this reminds me that I need to add a way to debug circular dependencies in Vite.

ferferga (approved):

PR #1939, opened 2023-04-10
fix: reloading after client changes

ferferga:
This should work out of the box for clients that have been running already, but the service worker will be permanent to them. all data associated with the site must be cleared for it to be completeley gone.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -88,7 +88,6 @@
     "unplugin-vue-router": "0.5.4",
     "vite": "4.2.1",
     "vite-plugin-pages": "0.29.0",
-    "vite-plugin-pwa": "0.14.7",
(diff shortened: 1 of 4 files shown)


frontend/index.html
```diff
@@ -6,6 +6,8 @@
     <meta
       name="viewport"
       content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover">
+	<meta http-equiv="Cache-Control" content="max-age=0, no-cache, no-store, must-revalidate">
+	<meta http-equiv="Expires" content="0">
```
ThibaultNocchi, line 10: Those aren't needed, I just tried  :)

ThibaultNocchi (approved):
LGTM

PR #1943, opened 2023-04-10
fix: remaining type errors

ferferga:
Locales are fully typed now. Keep track of https://github.com/intlify/vue-i18n-next/issues/1116 for better TS support for them

The rest of them are waiting for PRs at Vuetify's side, an SDK upgrade and a further investigation of our typed routes setup
Deep type instantiation is also fixed


frontend/src/components/Item/Metadata/ImageEditor.vue
```diff
@@ -7,8 +7,7 @@
       xl="1"
       lg="3"
       md="4"
-      sm="6"
-      xs="12">
+      sm="6">
```
ThibaultNocchi, line 10:
```suggestion
      sm="6"
      cols="12">
```

This is what the xs prop is really named

frontend/src/components/Item/Metadata/ImageEditor.vue
```diff
@@ -42,8 +41,7 @@
...
```
ThibaultNocchi, line 44: Same as above

frontend/src/components/Layout/AppBar/Buttons/UserButton.vue
```diff
@@ -3,7 +3,7 @@
     <template #icon>
       <user-image :user="auth.currentUser" :size="40" rounded />
       <v-menu location="bottom">
-        <v-list class="min-list-width" dense>
+        <v-list class="min-list-width">
```
ThibaultNocchi, line 6: Maybe we should use the `density` prop?

frontend/src/components/Playback/UpNext.vue
```diff
@@ -13,13 +13,12 @@
...
-                  {{ $tc('units.time.seconds', currentItemTimeLeft) }}
-                </span>
-              </template>
-            </i18n>
+            <span>
+              {{ $t('dialog.upNext.nextItemPlayingIn') }}
+              <span class="text-primary darken-2">
+                {{ $tc('units.time.seconds', currentItemTimeLeft) }}
+              </span>
+            </span>
```
ferferga, line 21: This is not strictly the original code, but as this component is still not plugged-in to the video player, it will probably need a refactor anyway.

The time should be passed as a prop as well, it looks strange to me to pass this as an slot.

ThibaultNocchi (approved):

PR #1947, opened 2023-04-12
ui: more library cards on big screens + slightly bigger font on home carousel

### LINKED ISSUE #1924: Scaling issue (i.e. 1440p)
### Description of the bug

There are scaling issue with some components on screens resolutions other than 1920p x 1080p. While it is not a critical issue, some UI components look a bit weird on different screen resolutions. As an example I provided a couple of screenshots in the screenshot section of this issue.

### Steps to reproduce

Use a 1440p screen resolution. This can easily be emulated in browsers.

### Expected behavior

Nice and 🤌 looking scaling on other screen resolutions than 1920p x 1080p.

### Logs

```text
N/A
```

## 1440p show thumbnails
afbeelding

## 1080p show thumnails
afbeelding

## 1440p home screen
afbeelding

## 1080p home screen
afbeelding

### Platform

Windows

### Browser

Firefox

### Jellyfin server version

10.8.9

### Additional context

```text
N/A
```

ThibaultNocchi:
fix #1924

* Display 12 cards on `xxl` screens (>1200px width) on the library page
* Slightly bigger font on home carousel titles. I haven't increased much due to text already be in rem and those handled by the browser.

### DIFF
--- a/frontend/src/components/Item/ItemGrid.vue
+++ b/frontend/src/components/Item/ItemGrid.vue
@@ -107,4 +107,12 @@ withDefaults(
 .large-grid.xl .card-grid-container.xl {
   grid-template-columns: repeat(5, minmax(calc(100% / 5), 1fr));
 }
+
+.card-grid-container.xxl {
+  grid-template-columns: repeat(12, minmax(calc(100% / 12), 1fr));
+}
+
+.large-grid.xxl .card-grid-container.xxl {
+  grid-template-columns: repeat(6, minmax(calc(100% / 6), 1fr));
+}
 </style>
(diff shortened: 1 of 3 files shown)


ferferga (approved):
@ThibaultNocchi LGTM, although honestly I think this is not the best solution.

Ideally, we get the grid working in such a way that we set the maximum height & width of a single card with a relative unit (no hardcoded pixels at all), so a single card has good sizing across all display sizes. 

Then, we let the grid fill all the available space with as many cards possible.

This would be fully display-size agnostic, setting the amount of cards to show for size ranges works but it's "hacky" imo. We aren't taking into account DPI either with this, which my proposal probably would.

This is not blocking at all (thus why I'm approving), but I'm leaving this so you can try and tackle it in this PR, or get this merged and open an issue to track this (as you prefer).

PR #1953, opened 2023-04-14
feat: load attached fonts properly

noaione:
Now attached fonts should load and not falling back to Liberation Sans

**Preview**
image

### DIFF
--- a/frontend/src/store/playerElement.ts
+++ b/frontend/src/store/playerElement.ts
@@ -85,7 +85,7 @@ class PlayerElementStore {
     }
   };
 
-  private _setSsaTrack = (trackSrc: string): void => {
+  private _setSsaTrack = (trackSrc: string, loadedFonts?: string[]): void => {
     if (
       !jassub &&
       mediaElementRef.value &&
@@ -94,6 +94,7 @@ class PlayerElementStore {
       jassub = new JASSUB({
         video: mediaElementRef.value,
         subUrl: trackSrc,
+        fonts: loadedFonts,
         workerUrl: jassubWorker,
         availableFonts: { 'liberation sans': jassubDefaultFont },
         // Both parameters needed for subs to work on iOS
@@ -101,6 +102,12 @@ class PlayerElementStore {
         onDemandRender: false
       });
     } else if (jassub) {
+      if (Array.isArray(loadedFonts)) {
+        for (const font of loadedFonts) {
+          jassub.addFont(font);
+        }
+      }
+
       jassub.setTrackByUrl(trackSrc);
     }
   };
@@ -113,6 +120,17 @@ class PlayerElementStore {
 
       jassub = undefined;
     }
+  };
+
+  private _isSupportedFonts = (mimeType: string): boolean => {
+    // ttf,otf,woff,woff2
+
+    return (
+      mimeType.startsWith('font/') &&
+      (mimeType.includes('ttf') ||
+        mimeType.includes('otf') ||
+        mimeType.includes('woff'))
+    );
   };
 
   /**
@@ -140,6 +158,14 @@ class PlayerElementStore {
         }
       }
 
+      const remote = useRemote();
+
+      let serverAddress = '';
+
+      if (remote.sdk.api) {
+        serverAddress = remote.sdk.api.basePath as string;
+      }
+
       playerElement.freeSsaTrack();
 
       /**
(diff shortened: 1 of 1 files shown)


frontend/src/store/playerElement.ts
```diff
@@ -115,6 +122,17 @@ class PlayerElementStore {
     }
   };
 
+  private _isSupportedFonts = (mimeType: string): boolean => {
+    // ttf,otf,woff,woff2
```
ferferga, line 126:
```suggestion
  private _isSupportedFont = (mimeType: string): boolean => {
```

We're checking a single font here
The comment is useless given how concise the function is, imo.

frontend/src/store/playerElement.ts
```diff
@@ -140,6 +158,14 @@ class PlayerElementStore {
         }
       }
 
+      const remote = useRemote();
```
ferferga, line 161: Move this to the top of the function please

frontend/src/store/playerElement.ts
```diff
@@ -140,6 +158,14 @@ class PlayerElementStore {
...
+
+      let serverAddress = '';
+
+      if (remote.sdk.api) {
+        serverAddress = remote.sdk.api.basePath as string;
+      }
```
ferferga, line 167:
```suggestion
      const serverAddress = remote.sdk.api?.basePath;
```

frontend/src/store/playerElement.ts
```diff
@@ -154,6 +180,19 @@ class PlayerElementStore {
         (sub) => sub.srcIndex === playbackManager.currentSubtitleStreamIndex
       );
 
+      const attachedFonts =
+        playbackManager.currentMediaSource?.MediaAttachments?.filter((attach) =>
+          this._isSupportedFonts(attach.MimeType || '')
```
ferferga, line 185:
```suggestion
        playbackManager.currentMediaSource?.MediaAttachments?.filter((a) =>
          a.MimeType ? this._isSupportedFonts(a.MimeType) : false
```

I think this one is less obscure for a filter function

frontend/src/store/playerElement.ts
```diff
@@ -154,6 +180,19 @@ class PlayerElementStore {
...
+        )
+          .map((attach): string | undefined => {
+            if (attach.DeliveryUrl && serverAddress) {
+              return `${serverAddress}${attach.DeliveryUrl}`;
```
ferferga, line 189:
```suggestion
          .map((a) => {
            if (a.DeliveryUrl && serverAddress) {
              return `${serverAddress}${a.DeliveryUrl}`;
```
To keep it consistent from above.
I also think the type annotations for the return are unnecessary in this case, but ignore that change if locally you get linter complaints.

frontend/src/store/playerElement.ts
```diff
@@ -154,6 +180,19 @@ class PlayerElementStore {
...
+            }
+          })
+          .filter((attach) => attach !== undefined && attach !== null) as
+          | string[]
+          | undefined;
```
ferferga, line 194:
```suggestion
          .filter((a): a is string => a !== undefined && a !== null) ?? [];
```

Use a type guard so types are correct after filtering

frontend/src/store/playerElement.ts
```diff
@@ -94,13 +94,20 @@ class PlayerElementStore {
...
         prescaleFactor: 0.8,
         onDemandRender: false
       });
     } else if (jassub) {
+      if (Array.isArray(loadedFonts)) {
+        for (const font of loadedFonts) {
+          jassub.addFont(font);
+        }
+      }
+
```
ferferga, line 110:
```suggestion
    } else if (jassub && Array.isArray(loadedFonts)) {
        for (const font of loadedFonts) {
          jassub.addFont(font);
        }
```

Also, are you sure this is ever going to be needed? I'm on mobile right now, but iirc we created/destroyed jassub every time the currently playing item changes? Or maybe just on unmount?

ferferga (changes requested):
Thank you very much for your contribution! 🙌

Just checked and I didn't see you in our Matrix room, so take this as an invitation to join. It's more easier to get in touch there for planning/dev questions.

This is pretty good overall, but some minor nitpicky stuff to (probably) keep the complexity level below the threshold!

noaione:
Applied suggestion, somehow sonarlint complexity check got worse though. Also joined the matrix channels.

ferferga (approved):

PR #1961, opened 2023-04-16
Fix the favorite button not updating its state

### LINKED ISSUE #1921: Like button doesn't fill after being clicked
### Description of the bug

When marking an item as liked, a loading spinner will appear and, afterwards, the heart icon will not be filled, showing that the item is liked.

### Steps to reproduce

1. Like an item

### Expected behavior

The heart icon should be filled

### Logs

_No response_

### Screenshots

_No response_

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.8.9

### Additional context

_No response_

The-Noah:
When clicking the favorite button, it properly updates the backend but not the frontend. This leads to two issues:
1. It does not visually change to show the new state
2. Clicking the button again does nothing (instead of flip-flopping like it should)

This PR fixes that.

Closes #1921 

> Note: this fix uses local state in the `LikeButton` component to fix the issue. If it is preferred, I can change it so it will refresh the state from the backend and not use local state.

### DIFF
--- a/frontend/src/components/Buttons/LikeButton.vue
+++ b/frontend/src/components/Buttons/LikeButton.vue
@@ -17,10 +17,13 @@ import { useRemote } from '@/composables';
 const props = defineProps<{ item: BaseItemDto }>();
 const remote = useRemote();
 const loading = ref(false);
+const isFavoriteOverride = ref<boolean | undefined>(undefined);
 
 const isFavorite = computed({
   get() {
-    return props.item.UserData?.IsFavorite ?? false;
+    return isFavoriteOverride.value === undefined
+      ? props.item.UserData?.IsFavorite ?? false
+      : isFavoriteOverride.value;
   },
   async set(newValue) {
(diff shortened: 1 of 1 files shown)


ferferga:
This is how it worked before, but it's not the real fix for this.

Items are always refreshed when updated in the store, so that's not the problem either (see the items store and how we listen to websocket). 

The fix is to ensure reactivity is not disconnected when the Axios interceptors are used (see axios.ts)

I think I know how to fix it but it's not trivial and it needs a refactor of the item store, so the interceptor uses getters instead of a new array.

The-Noah:
@ferferga thank you for your insight. I would still like to tackle this issue.

ferferga:
Closing this PR to reduce a bit of the noise. This PR is simple enough it isn't meaningful enough to keep open, even for future refactors.

Please create a new one if you come up with a solution that ticks all the boxes.

PR #1964, opened 2023-04-16
feat: add Delete Item to context menu

noaione:
Split from the following PR: #1951

### DIFF
--- a/frontend/locales/en-US.json
+++ b/frontend/locales/en-US.json
@@ -45,6 +45,9 @@
   "customRating": "Custom rating",
   "darkModeToggle": "Toggle dark mode",
   "dateAdded": "Date added",
+  "delete": "Delete",
+  "deleteItem": "Delete media",
+  "deleteItemDescription": "Deleting this item will delete it from both the file system and your media library. Are you sure you wish to continue?",
   "details": "Details",
   "dialog": {
     "upNext": {
@@ -79,6 +82,7 @@
   },
   "failedRetrievingDisplayPreferences": "Unable to get display preferences. Using last known settings.",
   "failedSettingDisplayPreferences": "Unable to update display preferences.",
+  "failedToDeleteItem": "Failed to delete item",
   "failedToRefreshItems": "Failed to refresh items",
   "favorite": "Favorite",
   "features": "Features",
--- a/frontend/locales/en-US.json
+++ b/frontend/locales/en-US.json
@@ -82,7 +82,6 @@
   },
   "failedRetrievingDisplayPreferences": "Unable to get display preferences. Using last known settings.",
   "failedSettingDisplayPreferences": "Unable to update display preferences.",
-  "failedToDeleteItem": "Failed to delete item",
   "failedToRefreshItems": "Failed to refresh items",
   "favorite": "Favorite",
   "features": "Features",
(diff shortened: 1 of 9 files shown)


frontend/locales/en-US.json
```diff
@@ -79,6 +82,7 @@
   },
   "failedRetrievingDisplayPreferences": "Unable to get display preferences. Using last known settings.",
   "failedSettingDisplayPreferences": "Unable to update display preferences.",
+  "failedToDeleteItem": "Failed to delete item",
```
ferferga, line 85: We want to remove a lot of useless error message. We still don't have that documented anywhere, but this is roughly what we want:

* When we don't know what caused the error: Use the generic `anErrorHappened` key.
* If we know what the error is: Use a custom message

In this case, we don't add any useful information of **why** the process failed. If we had something relevant to say (i.e lack of permissions, etc), then it would make sense a new error message. But if there's no known way this can fail, we're better off using the generic error message (the user knows anyway that he was deleting an item, no need to tell him again).

frontend/locales/en-US.json
```diff
@@ -559,4 +563,4 @@
   "year": "Year",
   "years": "Years",
   "youMayAlsoLike": "You may also like"
-}
+}
```
ferferga, line 566: Your editor should've added a newline here, but seems it didn't. Please do it, since I'm not sure if it messes with Weblate

frontend/src/components/Dialogs/ConfirmDialog.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <v-dialog
+    class="confirm-dialog"
+    content-class="confirm-dialog"
+    :model-value="dialog"
+    :fullscreen="$vuetify.display.mobile"
+    @update:model-value="close">
+    <v-card height="100%" class="d-flex width-90">
```
ferferga, line 8: I think you can pass title and subtitle here as props so the template is cleaner and uses fewer lines. Same with text really.

ferferga (changes requested):
Thank you very much! Some things for the time being, maybe after you integrate the composable I have further feedback

frontend/src/components/Item/ItemMenu.vue
```diff
@@ -284,6 +303,28 @@ function onActivatorClick(): void {
...
+    useSnackbar(t('errors.anErrorHappened'), 'error');
+
+    return;
+  }
+
+  try {
+    await remote.sdk.newUserApi(getLibraryApi).deleteItem({
+      itemId: menuProps.item.Id
+    });
+    router.replace('/');
```
ferferga, line 320: I guess this is only needed if we're in the page of an item? We can check if the route contains the item id before doing this I guess?

ferferga (changes requested):

ferferga:
@noaione Would you mind if I push to your branch some nitpicky changes in the confirm dialog? Mostly code organization stuff, I think it would be much faster than telling you exactly what I want and I don't want to be annoying with that kind of stuff when you have many useful PRs lying around that are waiting to get merged :P (will get to them as soon as we get delete and InstantMix in)

noaione:
Yep, it's fine

ferferga (approved):

PR #1965, opened 2023-04-16
feat: add Instant Mix to context menu

noaione:

### DIFF
--- a/frontend/locales/en-US.json
+++ b/frontend/locales/en-US.json
@@ -247,6 +247,7 @@
   "playback": {
     "addToQueue": "Add to queue",
     "clearQueue": "Clear queue and stop playback",
+    "instantMix": "Instant mix",
     "playAll": "Play all",
     "playNext": "Play next",
     "playbackSource": {
@@ -442,6 +443,8 @@
   "signIn": "Sign in",
   "snackbar": {
     "addedToQueue": "Added to queue",
+    "instantMixFailed": "Failed to add Instant mix to queue",
+    "instantMixQueued": "Instant mix added to queue",
     "playNext": "The selected item will be played after the current one",
     "routeValidationError": "The specified routeId in route params is not correct"
   },
--- a/frontend/locales/en-US.json
+++ b/frontend/locales/en-US.json
@@ -114,6 +114,7 @@
   "images": "Images",
   "incorrectUsernameOrPassword": "Incorrect username or password",
   "instantMixQueued": "Instant mix added to queue",
+  "instantMix": "Instant mix",
   "item": {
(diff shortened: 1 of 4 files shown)


frontend/src/utils/items.ts
```diff
@@ -81,6 +81,16 @@ export function isValidMD5(input: string): boolean {
...
 
+/**
+ * Checks if an item is a local item.
+ *
+ * @param item - The item to be checked.
+ * @returns - A boolean representing whether the item is a local item
+ */
+export function isLocalItem(item: BaseItemDto): boolean {
+  return item?.Id?.indexOf('local') === 0;
+}
```
ferferga, line 92: Why is this needed?

frontend/locales/en-US.json
```diff
@@ -442,6 +443,8 @@
   "signIn": "Sign in",
   "snackbar": {
     "addedToQueue": "Added to queue",
+    "instantMixFailed": "Failed to add Instant mix to queue",
```
ferferga, line 446: Same thing I mentioned in #1964 about the error keys

frontend/locales/en-US.json
```diff
@@ -559,4 +562,4 @@
   "year": "Year",
   "years": "Years",
   "youMayAlsoLike": "You may also like"
-}
+}
```
ferferga, line 565: Same as #1964 about possibly Weblate getting mad at this.

ferferga (changes requested):

frontend/src/utils/items.ts
noaione: Legacy code? It also used in jf-web so I'm basically just adding it here

ferferga: @noaione I think we can safely remove this, if it works without it. There's no point in adding obscure code (server and web are full of them since the fork) if it works without. If later on there's a related bug with it, we can always fix it.

I recall hearing about that property before, and I think it was related for offline usage (so cases when the client had downloaded content for offline playback). That never ended up getting implemented.

We can construct some stuff thinking forward and taking into account offline support, but assuming we can't do instant mixes because the items are cached offline *it's too much*

frontend/locales/en-US.json
ferferga: @noaione Sorry for forgetting about telling this before: Put all the new keys in all your PRs at the root level please. I'm trying to remove a lot of duplication around keys and I found out that most of that duplication is done blindly while using subkeys.

I have a branch ready locally to remove all the nesting after all your PRs are in :)

ferferga (approved):

PR #1966, opened 2023-04-16
feat: add Identify to context menu

noaione:


frontend/src/components/Item/ItemMenu.vue
```diff
@@ -36,6 +36,10 @@
     v-if="item.Id"
     v-model:dialog="metadataDialog"
     :item-id="item.Id" />
+  <identify-dialog
+    v-if="item.Id"
+    v-model:dialog="identifyItemDialog"
+    :item-id="item.Id" />
```
ferferga, line 42: Please match the callback style I used for the refreshmetadata dialog, to ensure only strictly necessary components are on screen all the time.

frontend/src/components/Item/Identify/IdentifyDialog.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <v-dialog
+    content-class="identify-dialog"
+    :model-value="dialog"
+    :fullscreen="$vuetify.display.mobile"
+    @update:model-value="close">
```
ferferga, line 6: Use ``after-leave`` and simplify the emits as I did in the refresh metadata.

frontend/locales/en-US.json
```diff
@@ -101,6 +101,13 @@
       "recentlyAdded": "Recently added"
     }
   },
+  "identify": {
+    "applyError": "Failed to apply selected search result",
+    "instructResult": "You can click on one of these cards to apply the result",
+    "replaceImages": "Replace all images",
+    "searchError": "Failed to search to remote providers",
+    "title": "Identify"
+  },
```
ferferga, line 110: No nesting, leave these at the top level

frontend/locales/en-US.json
```diff
@@ -318,6 +325,8 @@
   "saved": "Saved",
   "search": {
     "name": "Search",
+    "noResults": "No results found",
+    "results": "Results",
```
ferferga, line 329: Same as above

frontend/src/components/Item/Identify/IdentifyTab.vue
ferferga, line 1: I belive the content of this file could be merged to IdentifyDialog?

ferferga (commented):
been really busy with real life, but I tried to give a huge push to your PRs today. Thank you very much for your patience! 🙌

Some comments from a surface review, some others might come up later when I do a deep dive.

frontend/src/utils/items.ts
```diff
@@ -431,3 +470,154 @@ export function getMediaStreams(
...
+
+  if (remote.sdk.api === undefined) {
+    return;
+  }
+
+  if (itemId === undefined) {
+    return;
+  }
+
+  // Split the query to `search-item-` and the rest to provider IDs.
```
ferferga, line 497: Would be great if you can add more context on why this is needed

ferferga (changes requested):
Looks really good so far!

One general question though: Are you using the inspiration of the metadata editing dialogs, right?
Their styles are a bit convoluted and I think it's much better if we use the normal `v-window` vuetify stuff for tabs, instead of reinventing the wheel.

The metadata pages can be changed later to match that design as well.

ferferga:
@noaione Wdyt also about changing this to be an [`v-stepper`](https://v2.vuetifyjs.com/en/components/steppers/)?

They're not in Vuetify 3, but they might be a better suit here, so a `TODO` comment specifying that this could be migrated to that component might be suitable here :)

noaione:
Honestly, design is my weakest point, I never really use Vue so I didn't know the best approach to it lol.

Using v-stepper looks nice but I think it's Vue 2 only?

frontend/locales/en-US.json
```diff
@@ -101,6 +101,11 @@
       "recentlyAdded": "Recently added"
     }
   },
+  "identify": "Identify",
+  "identifyApplyError": "Failed to apply selected search result",
```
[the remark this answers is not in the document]
noaione: Would it be better to just use console and `errors.anErrorHappened` for this, same with the `identifySearchError`?

frontend/src/store/userLibraries.ts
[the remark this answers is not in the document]
noaione: I guess not specifying fields would actually return every fields?

frontend/src/components/Item/Identify/IdentifyDialog.vue
```diff
@@ -0,0 +1,329 @@
...
+
+const props = defineProps<{ item: BaseItemDto; mediaSourceIndex?: number }>();
+
+const emit = defineEmits<{
+  (e: 'close'): void;
+}>();
+
+const { t } = useI18n();
+const remote = useRemote();
+const tabName = ref<string>('searchMenu');
```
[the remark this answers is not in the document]
noaione: I'm using `ref` since that's what also used in metadata search.

frontend/src/components/Item/Identify/PosterBlock.vue
```diff
@@ -0,0 +1,79 @@
...
+  height: calc($maxW / $ar);
+}
+
+.thumb-m {
+  $ar: calc(16 / 9);
+
+  aspect-ratio: $ar;
+  max-height: calc($maxW / $ar);
+  height: calc($maxW / $ar);
+}
```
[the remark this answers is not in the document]
noaione: I think you meant `Card`, since I checked Blurhash and it didn't exist there. I'll try to resolve this

frontend/src/utils/items.ts
noaione: Updated the comment

noaione:
Tried to change the tab design and use something similar to `v-stepper` by using `v-linear-progress` to indicate progress.

https://user-images.githubusercontent.com/34302902/235566239-8cf22e10-0163-49c8-9636-99bebdc21f98.mp4

frontend/locales/en-US.json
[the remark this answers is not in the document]
ferferga: Yep, that was my point :)

If you don't have anything relevant about the source of the error, better to use the generic one.
However, if you think a common error source might be that the search box is empty, a message like `Can't perform search without an empty search query` gives the user more context about **why the error happened**.

frontend/src/components/Item/Identify/IdentifyResult.vue
```diff
@@ -0,0 +1,52 @@
...
+
+  $xSpace: 1rem;
+  margin-right: $xSpace;
+  margin-left: $xSpace;
+  margin-bottom: 1rem;
+
+  cursor: pointer;
+
+  /* 2/3 ratio */
+  max-width: 220px;
```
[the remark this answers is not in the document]
ferferga: @noaione Can't we use normal flex or CSS grids? That way it looks better in mobile too (in mobile every card has a different size depending on the inner text size)

frontend/src/components/Item/Identify/IdentifyResult.vue
```diff
@@ -0,0 +1,52 @@
...
+    variant="outlined"
+    @click="$emit('select', item)">
+    <poster-block :url="item.ImageUrl ?? undefined" :item-type="itemType" />
+    <div class="text-center text-subtitle-1 font-weight-bold mt-2 mx-2">
+      {{ item.Name }}
+    </div>
+    <div
+      class="text-center text-body-2 mt-1 mx-2"
+      :class="{
+        'mb-2': typeof item.ProductionYear !== 'number'
```
ferferga, line 13: I don't understand exactly why this style change is needed? 😅
Can you add some screenshots please?

frontend/src/components/Item/Identify/IdentifyDialog.vue
```diff
@@ -0,0 +1,363 @@
...
+    isLoading.value = false;
+    emit('close');
+    tabName.value = 'searchMenu';
+  }
+}
+
+/**
+ * Get a pre-filled search data object.
+ */
+function getFilledSearchData(
```
ferferga, line 238: Shouldn't we also prefill `name` and `year` fields?

ferferga (commented):
No worries about the lack of UI/UX experience, we don't have either!

I pushed some changes which I think are better for UI/UX. All of the minor details are included in the commit message itself, so you can take a look if you want.

frontend/src/components/Item/Identify/IdentifyResult.vue
noaione: Since ProductionYear can be missing, we don't want to add extra bottom margin between it. Although I guess it's not as readable as just using `my-2` in the ProductionYear <div/>

frontend/src/components/Item/Identify/IdentifyResult.vue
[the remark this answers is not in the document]
noaione: Used flexbox now

ferferga:
@noaione I had to bring back the explicit fields in the requests. Besides that, everything is intact from my force push
I added additional commits though:

* Everything is computed right now, which will help solving edge cases
* Placeholders for original values and clearable fields
* I discovered a better way to get provider IDs
* I used Vuetify's card styles solely. I think the design is much better now, though I admit it can still be improved. I'm not sure if we should keep the images with the `cover` prop or not. Please let me know.
* Added tooltips for long titles and confirm dialog

image
image

noaione:
@ferferga I would absolutely adjust the margin between the title and year in the results part.
And the new way to get Provider IDs absolutely remove some other provider IDs form input that can be used on said item, so I'll prefer the old ways, then prefill it.

**Jellyfin Web**
image


**Jellyfin Vue (Current)**
image

noaione:
Also the image box for music album is on portrait mode instead of square?

image

ferferga:
@noaione 
* The missing providers was an edge case I didn't consider. Fixed now
* Fixed card styles and add loading states and placeholders for broken images.
* 
How exactly do you want to adjust the margin? Can you commit it yourself please?

noaione:
> How exactly do you want to adjust the margin? Can you commit it yourself please?

The overall styling are already changed, so this doesn't apply anymore.
I think I don't have any more comment on this.

ferferga:
@noaione Great, I'll clean a bit the commit history tonight and merge. Thank you very much!

If have a proposal for you in case you want to take another Vue challenge and learn a lot along the way :): I think we can refactorize the Card component so it exposes `image`, `title` and `subtitle` slots, *that are used if no `item` prop is passed. **This way we can decouple the layout from the item rendering logic** The `image` slot child will have to contain the ``absolute-cover` class, so any custom element rendering in the card area will cover the entire space.

You can check the `VirtualGrid` component for some advanced slot usage to see how it works. Everything else is a handful of `v-if` to `$slots`.

Hopefully this will help us to create any Card variation, regardless if we have BaseItemDto or not. With type generics coming with Vue 3.3 (right around the corner, [see this](https://www.reddit.com/r/vuejs/comments/123cv9g/sfc_with_ts_generics_coming_in_33_evan_you/) and [this](https://github.com/vuejs/rfcs/discussions/436)) we would be able to also reuse the same grid we have for libraries, hence having a much more uniform styling.

ferferga (approved):

PR #1969, opened 2023-04-17
fix: unable to submit metadata edit

noaione:
Using computed Tagline instead of `metadata.Taglines` since that is a list.
Honestly, a weird choice to use list instead of just a normal string (which exist as `Tagline` but not used?).


ferferga (approved):
Our inherited API is full of weird quirks, so this doesn't surprise me at all. You'll keep amazing yourself for really long!

Thank you for noticing this! Sorry for the delay in your other PRs, I'm still on them, but life got me a bit busier now! 🙌

PR #1971, opened 2023-04-17
fix: volume keep resetting after stopping playback

### LINKED ISSUE #1956: Media volume resetting when stopping/playing back
### Description of the bug

Volume keep resetting to 100% every time you reset the playback manager.
The volume indicator is correct but the audio loudness are at 100%

### Steps to reproduce

1. Play any song/video
2. Set the audio volume low enough
3. Go out the playback/stop
4. Play it again
5. Audio is back at 100% while visual staying the same.

### Expected behavior

Volume kept at the last used one.

### Logs

_No response_

### Screenshots

https://user-images.githubusercontent.com/34302902/232099063-90a4c9a9-e253-439e-901b-14ed9b70bd73.mp4

### Platform

Windows

### Browser

Firefox

### Jellyfin server version

10.8.9

### Additional context

Using build from https://github.com/noaione/jellyfin-vue/tree/feat/more-context-menu

noaione:
Since the media element itself got removed after stopping, we basically need to reapply the current volume.
The solution is to just check if the `mediaElementRef` got set to `null` or `HTMLMediaElement` and then manually apply the current volume to the element directly.

Fix #1956

### DIFF
--- a/frontend/src/components/Layout/VolumeSlider.vue
+++ b/frontend/src/components/Layout/VolumeSlider.vue
@@ -8,13 +8,13 @@
       <v-icon :icon="icon" />
     </v-btn>
     <v-slider
-      v-model="sliderValue"
+      v-model="mediaVolume"
       class="volume-slider"
       hide-details
       thumb-label
       max="100">
       <template #thumb-label>
-        {{ Math.round(sliderValue) }}
+        {{ Math.round(mediaVolume) }}
       </template>
     </v-slider>
   </div>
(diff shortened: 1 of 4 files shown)


noaione:
Applied your suggestion, although I don't know if it's what you wanted

noaione:
Looks like I did something wrong since the volume is resetting again

ferferga:
@noaione I found out that this is a VueUse's bug. Opened PR there https://github.com/vueuse/vueuse/pull/2999, as the bug is an inconsistency between the DOM element's and composable's state.

Will keep this PR open until the fix is merged upstream and VueUse is updated by Renovate PRs to track this issue (and in the strange case we need to perform additional changes)

You can test my VueUse's changes by adding locally a copy of the composable and importing it instead from VueUse. The behaviour should match 1:1 your branch here. If not, please let me know so we can review it

ferferga:
Checked and fixed for me in #1970

Please let me know if there's an edge case not working as expected

PR #1980, opened 2023-04-28
fix: scanning task does not finish running

noaione:
Accompaniment for PR #1963 

For anything other than library, the task indicator would just infinitely run since their changes are published to `LibraryChanged` message.

Although, sometimes it will still run if there's no changes applied to it.


ferferga (changes requested):
This should live in the taskManager store socket's watcher to keep it within the same scope.

ferferga (approved):

PR #1985, opened 2023-05-01
chore: preparation for i18n cleanups

aweebs:
Pretty straight forward change to get the 918n extract commands working again. Running the remove script clears a lot out, but leaves empty blocks in the files :(
eg
```
{
  foo: {
     bar: 'baz'
  },
  ...
}
```
will become the follow if `foo.bar` is not used anywhere
```
{
  foo: {},
  ...
}
```
Happy to run remove and add the changes in too as part of this PR if that's desirable.

---

**EDIT: See [this](https://github.com/jellyfin/jellyfin-vue/pull/1985#pullrequestreview-1407675624) and [this](https://github.com/jellyfin/jellyfin-vue/pull/1985#issuecomment-1531201285) to have more context about the objective of this PR**


ferferga (changes requested):
I have started working on this locally. My idea was to have a PR fixing all the quirks first and later on another one running a cleanup. I don't mind discarding my local work if you want to tackle the changes we want yourself. I can also push the changes to this branch so you can continue where I left off. This is what I have done so far:

- [x] Ensure not only components are checked for strings, but also stores, etc
- [x] Add and configure eslint-plugin-vue-i18n
- [ ] Remove dynamic usage of translation keys (partly done). Keys that were passed as raw strings to `$t` to the template, for instance, can't be detected by vue-i18n-extract
- [ ] Flatten nested translation keys (partly done). We had a lot of duplication with nested keys, so we will use one level going forward and call it a day.
- [ ] Flatten the same keys in the rest of locales (not a must, but a good one to have to be fair)
- [ ] Apply the new keys to the code
- [ ] (FOR ANOTHER PR) Run vue-i18n-extract

Let me know what you want to do. Right now exams are ahead and I'm busy reviewing the context menu changes and possible architecture changes to better suit changes geared towards Tauri or external plugins (like the intro skipper plugin), without bloating the core application along the way,

aweebs:
> I have started working on this locally. My idea was to have a PR fixing all the quirks first and later on another one running a cleanup. I don't mind discarding my local work if you want to tackle the changes we want yourself. I can also push the changes to this branch so you can continue where I left off. This is what I have done so far:
>
> * [ ]  Remove dynamic usage of translation keys (partly done). Keys that were passed as raw strings to `$t` to the template, for instance, can't be detected by vue-i18n-extract

Are you sure there's no way to get this going? But I never liked the global `$` template variables anyway, but the fact it's not parsable at all is surprising.

> * [ ]  Flatten nested translation keys (partly done). We had a lot of duplication with nested keys, so we will use one level going forward and call it a day.

I think there's value in having separate translations for each component and route, and de-duplicating common translations. Having everything as a sorted list of translations strikes me as even messier.

> Let me know what you want to do. Right now exams are ahead and I'm busy reviewing the context menu changes and possible architecture changes to better suit changes geared towards Tauri or external plugins (like the intro skipper plugin), without bloating the core application along the way,

@ferferga I figured getting the scripts at least executing with their configuration would be a decent start and this could at least be merged but feel free to close if you have a PR in the works.
I'm also fairly strapped for time lately and won't be able to work on a bigger project like this for a while, but is your work in progress in a branch somewhere that I can look at it?

ferferga:
@aweebs I had the commits locally, I pushed them directly to this branch so you can pick them over if you wish. I deleted the one that started flattening the translations keys, since there were so many conflicts introduced by the `ItemMenu` PRs that it was easier to start afresh than resolving conflicts (which potentially could remove keys).

> Are you sure there's no way to get this going? But I never liked the global $ template variables anyway, but the fact it's not parsable at all is surprising.

The issue it's not in `t` vs `$t`. Check my commit, you will probably understand better. The issue was when we passed a variable instead of a raw string to them, in reality.

The way to work around it is to refactor the logic to always pass the raw string to the function directly, so `vue-i18n-extract` can get it's context. What I basically did for this was to run a cleanup of unused translations and check one by one the ones removed in `en-US`

> I think there's value in having separate translations for each component and route, and de-duplicating common translations. Having everything as a sorted list of translations strikes me as even messier.

Sadly this is a hard no. We've been how to do localisation issues since the first days of Jellyfin Vue. Never got to an agreement, so everybody did what they wanted (understandable given there are infinite options and approaches to localisation, and in general it's a common problem in software development). Localisation has always been the elephant in the room nobody wants to look at or treat properly. And, at the end of the day, it's a problem that us, the maintainers, will have to carry on with.

I know it might seem it makes more sense at first, but for us it doesn't work: A lot of times new contributors (and even ourselves) simply added keys without searching. Other times, we added them by concern, but without realising they existed beforehand somewhere else. Then, you need to move that translation key to another place where it makes more sense. Doing that kills the translation in other languages. Hence translators needed to redo their work multiple times. And this happened **a lot**. And we're not even talking about the duplication, which, if you want to tackle 

**TLDR: There are a lot of maintenability issues with translations and a lot of opinions about the matter, so sadly I need to take this hardly and push forward the ways the current leaders of this project (@ThibaultNocchi and me) like the most/want to work with.**

In fact, in my opinion what we should do is to move to the "key is the value" approach. Something [like this](https://github.com/helixsoftco/vuelix/tree/main/locales), even with all the drawbacks you'll find online because:

* Master language becomes the source of trust no matter what. en-US can be locked at the Weblate level, avoiding unintended changes (since that's another problem, we don't have any kind of review/reviewers approving translations users do at Weblate) or vandalism, since those changes would require both a PR to be fixed in the code as well, not only the locale files
* You get real autocompletion features in IDE, without any guesswork
* By reading the code, even on GitHub, you'll always know what text will be displayed.
* If there are different contexts that require different grammar features or pluralization, we're forced to do new translations for those, instead of poor concatenization of single terms (which might not be applicable to other languages).

But almost nobody in the team liked this approach, and @ThibaultNocchi didn't like it either, so we found this middle-ground of removing all flattening to simplify translation's usage and remove duplicated terms (and, in case of future key changes, this will make things easier) but this is still an ongoing discussion (which probably can be moved outside Matrix and moved to GitHub, like a lot of other discussions we have, but we're trying to improve in that regard and we want to be less lazy 😁. Formalizing things is really boring and consuming to be honest, but its a necessary evil).

> I figured getting the scripts at least executing with their configuration would be a decent start and this could at least be merged but feel free to close if you have a PR in the works.
I'm also fairly strapped for time lately and won't be able to work on a bigger project like this for a while, but is your work in progress in a branch somewhere that I can look at it?

No problem in leaving them open here, a PR can be merged from commits of multiple people and that's the power of git's collaboration 😁. So any spare time you find to do some of the remaining work will be welcome as always ^^.

ferferga:
Bundle size went from 3,031.03 kB to 2,375.00 kB after merging this

PR #2001, opened 2023-05-04
feat(player): Add media key control support

endrl:
* Add support for media control keys
* Add volumeUp, volumeDown to playbackManager


ferferga (changes requested):
I saw your comment about media keys in #1885, however I wanted to give you a response to the other stuff you brought up about TVs as well. I don't like people waiting too long for my feedback, but I'm super busy with exams so I'm addressing the smallest and oldest stuff first. But here's a sum up of our stance: We want to include 

> The question is should be "keyboard navigation" with auto focus and highlight a default feature which is always enabled? Or a build flag? Webos, tizen and any other client(*older) that might want to embed jellyfin-vue needs compatibility hacks anyway

If there is demand for this, and it's easy to give backwards compat, sure we want this stuff in! But we want to keep everything in `@jellyfin-vue/frontend` isolated from plugins (like the *Skip intro* feature @gibahjoe worked on #1887). Ideally we're looking forward to a multi-project (no build flags like in your PR either) with npm workspaces or microfrontend setup. I'm not knowledgeable enough in how to properly architecture all of that and specially microfronteds, so I'm first trying to educate myself. Once I feel confident enough, I might open a discussion to gather some feedback or a PR. But still it's much appreciated (and will allow to be shipped faster) if we get outside contributors' help in the form of RFCs or PRs. The TLDR of the ideal thing would be:

* Vite projects that expand/merges with the "core" configuration (the one we have now) for build changes targeted at specific stuff. Right now 2 comes to my mind:
       * A one targeting Tauri for bundling overrides for Tauri-specific logic, like the download one at #1962 
       * The one with the small set of polyfills needed for TVs

* Microfrontend architecture for features/specific components: That way, the "skip intro" button could be hooked by mounting it inside the video player's DOM and having access to the all the modules exposed by `@jellyfin-vue/frontend` (the main core)

*All the building process should be orchestrated by an interactive CLI configurator (but also allowing headless for CI), which will configure everything for the developer*

With the given examples, these are the configurations that come to my mind:
* Core Vite project + Core Vue app
* Core Vite project + Core Vue app + Skip intro (microfrontend)
* Tauri Vite project & Core Vite project+ Core Vue app
* Tauri Vite project & Core Vite project+ Core Vue app + Skip intro (microfrontend)
* Polyfilled Vite project & Core Vite project + Core Vue app
* Polyfilled Vite project & Core Vite project + Core Vue app + Skip intro (microfrontend)

Hopefully this clear enough for a vision of what we want to achieve

---

As for the contents of this PR itself: I don't understand why we need this? Since we already have [media shortcuts](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/pages/playback/video/index.vue#L192-L199) and [MediaSession support](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/store/playbackManager.ts#L490) (I can switch tracks in all of my devices without issue. What's the purpose of those special keys then?

endrl:
That's huge! But a good write up. I overlooked that implementation, but again it's chrome >=73 so may be moved as a fallback `if (navigator.mediaSession) else USE_KEYS`. But if i understand the idea this should be moved to a sub project (depends on how hard the line is, no one will touch this ever again. But these snippets blow up the code size).

Good luck!

ferferga:
But first of all: what are those keys for? Do you have any documentation for them (like in MDN?)

Because no harm in adding them alongside the others, they don't seem incompatible with MediaSession (?)

endrl:
[Multimedia Keys](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#multimedia_keys)
I think they shouldn't be used together with media Session.

ferferga:
I guess you have tested them in a desktop browser, right? If they work right, then I guess we have no issue.

Would you mind extracting them to a composable, so we can call it in both playback pages (music and video) and in the AudioControls?

We can go with the settings page later. I don't see much usefulness in an enable/disable toggle though.

endrl:
Added https://github.com/jellyfin/jellyfin-vue/discussions/2039

Volume changes should be shown by an "action overlay", but that's out of scope of this PR.

frontend/src/composables/use-playerkeys.ts
```diff
@@ -0,0 +1,55 @@
...
+  whenever(keys.mediastop, playbackManager.stop);
+  whenever(keys.mediatracknext, playbackManager.setNextTrack);
+  whenever(keys.mediatrackprevious, playbackManager.setPreviousTrack);
+  whenever(keys.audiovolumemute, playbackManager.toggleMute);
+
+  whenever(keys.space, playbackManager.playPause);
+  whenever(keys.k, playbackManager.playPause);
+  whenever(keys.m, playbackManager.toggleMute);
+
+  const callHandler = (): void => {
```
ferferga, line 30: Why this is here? This should be in the fullscreen player page, it's not something that we need to share with the rest of the players.

frontend/src/composables/use-playerkeys.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (osdHandler) {
+      osdHandler();
+    }
+  };
+
+  const throttledForwardFn = useThrottleFn(() => {
```
ferferga, line 36: Why these are throttled?

ferferga (changes requested):
* Fix SonarCloud's bugs please
* Remove the changes to the type files, for some reason it looks like the linter run of those? Even though they're ignored.

frontend/src/composables/use-playerkeys.ts
endrl: I replaced it with another attempt. The blanket function is just here to prevent the `undefined` check. Would you prefer the latter?

ferferga: @endrl But why are you moving this out from the fullscreen video playback page? Unless I'm missing something, there's no reason to.

endrl: Can you please clarify about what  code lines we are talking? The current behaviour is that `video/index || audio/index || AudioControls` is mounted (but never at the same time). They all get the same key bindings and the video player has a callback fn to show the osd with a timeout when you press forward/backward.

ferferga:
@endrl Please test with my changes and let me know how it goes. I spun up a quick codespace session to test and still found no reason to have a throttle function there, it's not how Youtube or Netflix player (whose I'm using as reference) works. I also added my concerns about fully factorising the logic between the non-fullscreen and fullscreen players and possible conflicts

ferferga:
@endrl Please test with my changes and let me know how it goes. I spun up a quick codespace session to test and still found no reason to have a throttle function there, it's not how Youtube or Netflix player (whose I'm using as reference) works. I also added my concerns about fully factorising the logic between the non-fullscreen and fullscreen players

frontend/src/composables/use-playerkeys.ts
```diff
@@ -0,0 +1,51 @@
...
+  whenever(keys.AudioVolumeDown, playbackManager.volumeDown);
+
+  whenever(keys.f, playerElement.toggleFullscreenPlayer);
+  whenever(keys.space, playbackManager.playPause);
+  whenever(keys.k, playbackManager.playPause);
+  whenever(keys.m, playbackManager.toggleMute);
+  whenever(keys.j, backwardFn);
+  whenever(keys.l, forwardFn);
+
+  if (!isTizen() && !isWebOS() && fullscreen) {
```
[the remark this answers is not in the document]
endrl: While thinking about it. up,down,left,right are always used for navigation except:
* Use left, right to seek while player is in fullscreen
  * video: the osd will become visible and should focus the slider. Now you can further seek or navigate down to the control elements and use left/right without trigger seek)
  * audio: same aus video. except that the osd is always visible (visualizer mode with (probably) hidden osd is around the corner which should behave the same)

Did if forgot something? So from code view:

Concept
- Currently `v-overlay` unmounts the video osd so we can't do -> mount fullscreen video + osd and focus slider
  - Attempt: listen just for left/right when video is fullscreen and osd is unmounted (to get the osd rendered, now the osd can capture focus for slider)
  - or refactor `v-overlay` to a div with `v-show` to force focus and prevent the above "temporary listener hack"

Conclusion:
Depending on the code solution the guards are required as is or modified.

Out of topic: You are immediately at the key navigation . But you don't need a TV for that. Imagine you have just a keyboard with left/top/right/down. If we get that good implemented support for all kind of TV views is ready

frontend/src/composables/use-playerkeys.ts
```diff
@@ -0,0 +1,51 @@
...
+    whenever(keys.ArrowUp, playbackManager.volumeUp);
```
endrl, line 46: You missed the point of `onKeyStroke`, you keep the btn pressed, which triggers endless events that needs to be throttled. Adds to feature of "keep the button pressed to seek through audio/video"

endrl:
Sounds interesting. So `osdHandler` would focus something and after timeout "unfocus".

Actually i am not sure if we should keep `v-slider`. From a future perspective it might be too limited when it comes to plugins and more features

- mouse hover won't show the new position (you need to click on that (pressed))
  - Jellyscrub to provide preview images just while you hover
- The html player element provides a buffer state. We can't show it.
- chapter marker?
- hook other plugins to show "whatever" on the slider

This is a lot to mess around, in worst case you overwrite private api and styling which we don't want (stability)

ferferga:
@endrl See this: https://github.com/jellyfin/jellyfin-vue/projects/4#card-89536907

This situation is exactly another motivation to start looking into the implementation details of Quasar as early as possible. Probably, once it gets into the client, the sliders are the first component that will get ported over (exactly for the limitation that's blocking this PR and your reasons).

See https://quasar.dev/vue-components/slider too

ferferga:
@endrl You'll be very happy soon then :)

If you have prior experience with it, I want to invite you to contribute here! Though it might take some time until I can review them properly and get merged, since there are other PRs and stuff waiting first (more details in the project board I linked before).

The most self-contained and small the PRs are, the easiest that it is to review them and move along faster :)

ferferga:
@endrl I added all the refactorization done in this PR to extract the common logic to a composable in #2264, as I was refactoring playerElement and I needed to apply the watchers to the page components themselves instead of the store. I added you as a co-author in the relevant commit.

Given this PR also included the extra logic for focusing the slider (and the issue it's not fixed upstream), I think this PR it's still relevant, but can you bring it up to date with current master so later on it can be picked up more easily? Do you have any interest in helping in the Vuetify removal of the client and create our new slider component?

PR #2013, opened 2023-05-17
feat: Reworked video keyboard keys and fix osd visual

endrl:
* OSD: refactor: end time position uses now flex instead of `<br/>`
* OSD: fixed: chars like `g` or `q` where cut at the bottom
* OSD: fixed: "mouse move" overwrites no longer staticOverlay by starting the timer
* ~~OSD: feat: keyboard events show the osd~~
* ~~VideoPlayer: continous events for left/right key down~~


endrl:
Ups, messed up the commit...

ferferga:
@endrl Why this isn't part of #2001?

endrl:
Good question. The topic there was also compatibility with webos. For a better feedback i want to fork `jellyfin-webos` with  patched `jellyfin-vue` and continue with final testing results.

This PR exists just because i am working on a new feature for `jellyfin` and better player control was required for testing, while i stumbled upon other things.

Out of curiosity. Am i the only one who can't use the mouse with the time slider? I found that `<v-slider @start=""` never emits and therefore never decouple from the player progress.

ferferga:
@endrl Media keys are not exclusive to webOS, so imo the key bindings from this PR could be merged to that one and leave the style fixes for this one.

Otherwise we scatter the context of the changes across multiple PRs + you'll need to fix conflicts more frequently.

PR #2033, opened 2023-06-06
Better Subtitles

bleak183:
Cross-browser responsive subtitles in javascript which are easily styleable through the PlayerElement.vue file.


ferferga (changes requested):
Additionally, it seems that you didn't run the linter. Run it using `npm run lint` and check if there are issues in the newly added code.

You must also do `npm run typecheck` fyi

ferferga:
It's still unclear to me what this PR wants to accomplish really, and from what I remember after I could find the original poster in Matrix chats, my changes doesn't apply to track elements and they must be divs (which means they will not be WCAG compliant).

There's also seemingly no interest whatsoever in the original poster, so I'm closing this.

PR #2047, opened 2023-06-12
perf: replace lodash's cloneDeep with klona

ferferga:
klona is miles faster and it's the most popular object cloning library: https://www.npmjs.com/package/klona

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -38,6 +38,7 @@
     "dompurify": "3.0.3",
     "hls.js": "1.4.5",
     "jassub": "1.7.1",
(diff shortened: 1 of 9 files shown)


ndom91:
@ferferga `structuredClone` ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#browser_compatibility)) is now supported in all major browsers. Maybe we don't need an npm package for this anymore?

ferferga:
@ndom91 Thank you very much for the hint! Wasn't aware of `structuredClone`.

Added in #2051

Issue #2009, opened 2023-05-13
Tauri app won't accept HTTP server url


prayag17:
### Description of the bug

The tauri client won't take a http:// server.

### Steps to reproduce

Try adding a server with http protocol 

### Expected behavior

The client should behave as normal

### Platform

Windows

### Browser

Chrome

### Jellyfin server version

10.8.10

### Additional context

```text
Tauri blocks XHR requests when in production but allows them in dev build
```

endrl:
Perfect, thank you for reaching out.

ferferga:
@prayag17 PR or indications in how to fix it are appreciated though

PR #2049, opened 2023-06-14
fix(tauri): app load

endrl:
Document and fix app load.

PR #2050, opened 2023-06-15
fix(tauri): http jellyfin server address

endrl:
Disable https context.

Closes: #2009


prayag17 (commented):
You don't need to use setup to create a new window, you can still use the older window created in tauri.conf.json, also you should disable the localhost server when in dev mode since vite already does that

endrl:
Updated, also added save/restore of window state which is probably more convenient

endrl:
Tauri merged a new option to enable http for windows. This supersedes the localhost plugin. https://github.com/tauri-apps/tauri/pull/7645

ferferga:
@endrl That's great news! As soon as I see a Tauri update from renovate I'll merge so you can update this PR accordingly :)

ferferga:
@endrl Can you try the artifacts of #2096 and let me know if merging that PR fixes this?

prayag17:
I will have to check if their fix works on Linux before merging it.

endrl:
Oh, already released. That's fast.

I installed the deb and i noticed:
- App Icon is not available
- Loading screen loops endless. It might be the CSP (console is not enabled!)

So we are back at: https://github.com/jellyfin/jellyfin-vue/pull/2049

ferferga:
@endrl But it's the original issue that ti's PR was addressing solved?

endrl:
Oh, there is missing flag. Now it's working. You can merge this and update to tauri 1.5

ferferga (approved):

tauri/tauri.conf.json
```diff
@@ -52,7 +52,8 @@
       }
     ],
     "security": {
-      "csp": "default-src blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'"
+      "csp": "default-src blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'",
+      "dangerousUseHttpScheme": true
     }
   }
-}
+}
```
ferferga, line 59:
```suggestion
}

```

ferferga:
@endrl The checos are still failing for some reason

ferferga:
@endrl Oops, good catch! Let's rerun checks after the renovate PR.

PR #2073, opened 2023-08-16
fix: music visualizer volume change

jakobkukla:
Output was doubled since the visualizer connects to the speakers by default. See https://github.com/hvianna/audioMotion-analyzer#connectspeakers-boolean for reference.


ferferga (approved):
Thank you very much for the investigation work! Shipping it! 🙌🚀

PR #2080, opened 2023-08-17
fix: player quirks and regressions

ThibaultNocchi:
* Fixed an infinite loop when changing sub and audio
* Fixed the subtitle pickers which considered disabled subs as -1 when the API expects undefined
* Fixed the source audio handler which triggered a refresh even when not changing
* Added a variable to remember and set the time when changing source due to an audio/burnt-in subs change.


frontend/src/components/Item/MediaStreamSelector.vue
```diff
@@ -151,7 +134,7 @@ if (
 watch(
   trackIndex,
   (newValue) => {
-    emits('input', newValue);
+    emits('input', newValue ?? undefined);
```
ferferga, line 137:
```suggestion
    emits('input', newValue);
```

Isn't this unnecessary?

frontend/src/pages/item/_itemId/index.vue
```diff
@@ -316,7 +316,7 @@ const item = (
 const currentSource = ref<MediaSourceInfo>({});
 const currentVideoTrack = ref<number>();
 const currentAudioTrack = ref<number>();
-const currentSubtitleTrack = ref<number>();
+const currentSubtitleTrack = ref<number | undefined>();
```
ferferga, line 319:
```suggestion
const currentSubtitleTrack = ref<number>();
```

Same as above, I think this is already resolved by TS as ``number | undefined`` without the explicit undefined.

ferferga (commented):
Does Vuetify VSelect work fine with the `-1 -> undefined` change, right?

frontend/src/components/Item/MediaStreamSelector.vue
ThibaultNocchi: As newValue can be null and not undefined, and that we need an undefined value in upper layers, yes I need it ^^

ferferga (approved):

PR #2103, opened 2023-10-05
Implemented most user settings

rdmchr:
I had some free time during my holidays so I implemented some user settings in the admin panel. The scope was too large to implement all settings during my holidays, but I hope that most admins will find all required settings implemented.

Also, you might find that on some pages there is way too much margin between items (example below), I'm not sure how to fix this. I mainly work with Tailwind in my day job and have no experience with fully-fledged component frameworks like Vuetify. I would love to hear how this can be solved, but it is fair to say that my experience with Vuetify was madding at times.
image

Nonetheless, if you don't dislike my code too much I would love to contribute some more in my next holidays. :)


frontend/src/pages/settings/users/index.vue
```diff
@@ -0,0 +1,106 @@
...
+              :key="user.Id ?? undefined">
+              <td
+                v-for="{ value } in headers"
+                :key="value">
+                {{
+                  value !== 'LastActivityDate'
+                    ? user[value]
+                    : useDateFns(
+                      formatRelative,
+                      parseJSON(user[value] ?? "2023-02-02T02:02:02Z"), //FIXME: using 'unknown' doesnt work here? its used somewhere else in the code as well, does that even work there?
```
rdmchr, line 50: Not sure what's going on here, but when tried using `"unknown"` instead of the hard-coded date I got this error: 
```
Uncaught (in promise) RangeError: Invalid time value
```
This happens for users who haven't logged in yet. I copied this code from somewhere else in this codebase, so if this gets fixed here that fix should also be tested there.

ferferga: How are you able to access this page if you're not logged in yet?

rdmchr: I don't understand your question. This is an admin page and should only be accessible by logged-in administrators, right? I didn't implement permission checks since those are on the server side. Do you want me to implement some code that redirects unauthorized/unauthenticated users back to the index/login page?

.docker/docker-entrypoint.sh
ferferga, line 1: Remove the commits that modify this please

frontend/locales/en-US.json
```diff
@@ -429,7 +432,46 @@
...
+          "maxAllowedRating": "Maximum allowed parental rating",
+          "maxAllowedRatingSubtitle": "Content with a higher rating will be hidden from this user.",
+          "parentalControl": "Parental Control",
+          "tagName": "Tag name",
+          "unblockTag": "Unblock tag"
+        },
+        "password": {
+          "confirmPassword": "Confirm new password",
+          "currentPassword": "Current password",
+          "newPassword": "New password",
```
ferferga, line 468: Fix this lint

frontend/src/pages/settings/users/_userId/index.vue
```diff
@@ -0,0 +1,532 @@
+<template>
+  <SettingsPage page-title="settings.users.users">
+    <template #actions>
+      <VBtn
+        variant="elevated"
+        href="https://jellyfin.org/docs/general/server/devices.html"
```
ferferga, line 6: This needs to be updated!
```suggestion
        href="https://jellyfin.org/docs/general/server/users/"
```

frontend/src/pages/settings/users/_userId/index.vue
```diff
@@ -0,0 +1,532 @@
...
+const router = useRouter();
+const remote = useRemote();
+
+const loading = ref<boolean>(false);
+const addTagDialogOpen = ref<boolean>(false);
+const newTagValue = ref<string>('');
+const user = ref<UserDto>({});
+const libraries = ref<BaseItemDtoQueryResult>();
+const parentalCategories = ref<{ label: string, id: number | undefined }[]>([]);
+const initializedUser = ref<{ Name: string, CurrentPassword: string, Password: string, ConfirmPassword: string, CanAccessAllLibraries: boolean, Folders: string[], maxParentalRating: number | undefined, BlockUnratedItems: UnratedItem[], BlockedTags: string[] }>({Name: '', CurrentPassword: '', Password: '', ConfirmPassword: '', CanAccessAllLibraries: false, Folders: [], maxParentalRating: undefined, BlockUnratedItems: [], BlockedTags: []});
```
ferferga, line 379: Remove this and run everything inside the onMounted hook at the top level, so navigation is blocked until all the data is populated and you don't need this.

frontend/src/pages/settings/users/index.vue
ferferga: You forgot to add [this](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/pages/settings/apikeys.vue#L101C1-L104C9)

As you said above, permissions are enforced server-side, but what's the point of letting normal users access stuff that they can't change?

frontend/src/pages/settings/users/_userId/index.vue
```diff
@@ -0,0 +1,532 @@
...
+                        density="compact" />
+                    </VCol>
+                  </VRow>
+                  <VRow dense>
+                    <VCol>
+                      <VCheckbox
+                        v-model="initializedUser.BlockUnratedItems"
+                        :label="t('shows')"
+                        value="Series"
+                        density="compact" />
```
ferferga, line 205: All of this can be ``v-for``'ed with a computed to deduplicate the code.

frontend/src/pages/settings/users/index.vue
ferferga, line 1: I'm not a fan of how this page looks at all. I have images of all possible sizes. Why not something more standard in material design like this? https://vuetifyjs.com/en/components/lists/#three-line (not necessarily with 3 lines)

You already have the UserImage component that's pretty and nice.

frontend/src/pages/settings/users/new.vue
ferferga, line 1: No padding at all, the button gets cropped
image

frontend/src/pages/settings/users/_userId/index.vue
ferferga, line 1: The position of the "Delete user" button is completely arbitrary:
image

"Add blocked tags" doesn't seem a button:
image

And here you're swapping the position of the two buttons, which might confuse users:

image

ferferga (changes requested):

frontend/src/pages/settings/users/index.vue
rdmchr: Wait, but I do have that in line 76, no?

frontend/src/pages/settings/users/_userId/index.vue
rdmchr: Not quite sure what you mean here. The logic should already be inside the onMounted hook. I made `initializeUser()` a separate function since I need that in two different places.

frontend/src/pages/settings/users/index.vue
rdmchr: Yeah I agree, a table is not the right element for this, but wouldn't this work better? https://vuetifyjs.com/en/components/cards/#custom-actions
(ignore the action, I'm just talking about the style here)

ferferga: @rdmchr I don't think so, since all the person images are portrayed as circular ones

frontend/src/pages/settings/users/_userId/index.vue
ferferga: @rdmchr You don't need ``onMounted`` for anything here, just extract it's contents to another function, like this:

```
async load() {
...
}

await load()
```

Read about [Vue's 3 Suspense](https://vuejs.org/guide/built-ins/suspense.html#async-setup)

frontend/src/pages/settings/users/index.vue
ferferga: @rdmchr My bad for everything: I thought you meant that you were able to access the route when no user was logged into the client (by inputting the exact url). Now I understand you mean **the users that are being managed**.

Any reason I'm missing why you don't do this like [web](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/dashboard/users/UserCardBox.tsx#L26)? Fired up a quick Codespace to check and it works fine with the same functions used and I'm not seeing any shortcomings with it.

frontend/src/pages/settings/users/_userId/index.vue
rdmchr: Wow,, I didn't know it was that easy. Thanks for the tip 😄

frontend/src/pages/settings/users/index.vue
rdmchr: No, I made a mistake here. I copied that code from https://github.com/rdmchr/jellyfin-vue/blob/9a1f88c5d61b0f3f53268b0a604f5be7bfff8f6c/frontend/src/pages/settings/apikeys.vue#L44-L50 and in that case, the `'unknown'` will "never" be the outcome, since every key has a date created, but not every user has a date when they were last seen (ex. newly created keys have a created date, freshly created users were never seen). 
So in the original place that code should work.

I think we should create some utility function for this use case which handles it correctly for both cases.

ferferga: @rdmchr And do you see any shortcoming to the Web's implementation? I think it's better to use browser's native functions when possible, they should be less prone to errors.

frontend/src/pages/settings/users/index.vue
rdmchr: Yeah, I see what you mean. This layout doesn't work with round images... 😄 

image

frontend/src/pages/settings/users/index.vue
rdmchr: @ferferga I implemented a basic method. Currently, you have to use it like this:
```Vue
          <VCardSubtitle>
            {{ $t('settings.users.lastActivityDate', {
              value: user.LastActivityDate ? getRelativeTime(new Date(user.LastActivityDate)) : t('unknown')})
            }}
          </VCardSubtitle>
````
To get this result:
image

Might be worth adding all that additional stuff to the method as well since I don't see any use case where you would want just the relative time.

File: frontend/src/pages/settings/users/index.vue
Comment (ferferga) [reply frontend/src/pages/settings/users/index.vue:1]: @rdmchr You have a lot of UI resources and examples at https://m3.material.io/, which can help you make up better designs in your head.

File: frontend/src/pages/settings/users/index.vue (hunk 1 above)
Comment (ferferga) [reply frontend/src/pages/settings/users/index.vue:50]: Two things about the text:
1. "Last seen unknown ago" sounds somewhat good in English, but with it we're overriding the date-fns translations and adding extra complexity. In Spanish, your sentence would be "Última vez hace desconocido" which is completely wrong. This solely makes me reluctant to add this in this state.
2. "unknown" should be lowercased.
3. You're assuming we don't know the date of the last login, while in fact the missing property means "this user has no logged-in yet". I don't know if there are other cases where the server can return ``undefined | null`` for this, but I can't imagine another. When a user hasn't logged-in, that's **a very precise information you know**, not *unknown*.

I'm divided about this, I think the safest approach would be to do it like web (skip displaying the text altogether) but the most concise one would be to display the "not logged-in text", even though it might have some edge cases we're not aware of.

Also, I'm not a fan of having an exclusive "shortcut" function for something that's not repeated many times in the codebase. Though I was unsure of what you mean, I thought you would add additional processing to the function, but in this case I think you're better off leaving it as is (or extracting the function from the template and putting it inside ``script setup``).

File: frontend/src/pages/settings/users/index.vue
Comment (rdmchr) [reply frontend/src/pages/settings/users/index.vue:1]: @ferferga Had a quick look, but I'm still stuck on the card design. Maybe something like this could work with the rounded profile pictures. In our case, the right image would have to be removed. It would also look more like the component you linked at the top, but on desktop, I think a grid would be better.
image

Comment (rdmchr) [reply frontend/src/pages/settings/users/index.vue:1]: @ferferga Alright, I think this one might be it.

image

File: frontend/src/pages/settings/users/index.vue (hunk 1 above)
Comment (rdmchr) [reply frontend/src/pages/settings/users/index.vue]: Yeah, you're right, I have removed the function and updated the page according to your recommendations.

File: frontend/locales/en-US.json [hunk 6]
```diff
@@ -429,7 +432,46 @@
...
+          "password": "Password",
+          "resetPassword": "Reset password"
+        }
+      },
+      "users": "Users"
+    }
```
ferferga, line 474: Can you put all those keys at the root level? We're killing off all the nested subkeys. More [here](https://github.com/jellyfin/jellyfin-vue/pull/1985)

ferferga (changes requested):
Did another round. It'd be great if you can take care of this one more thing.

There are some stuff I think that could be simplified in the index page, but I need to test them directly and I think it's far quicker if I commit the changes myself.

frontend/locales/en-US.json
m-radmacher: Can do. Should I just remove the categories and put all the keys on the root level or is there any syntax I should follow? Putton them all on the root level might cause conflicts if I don't rename some of them.

ferferga: @rdmchr Yeah, exactly that.

Are there any conflicts whose value differs from your original text? If not, just reuse the original key!

frontend/locales/en-US.json
```diff
@@ -429,7 +432,46 @@
...
+          "name": "Name",
+          "profile": "Profile"
+        },
+        "access": {
+          "access": "Access",
+          "allLibraries": "Enable access to all libraries"
+        },
+        "parentalControl": {
+          "addBlockedTag": "Add blocked tag",
+          "blockUnratedItems": "Block items with no or unregocnized rating information:",
```
oxixes, line 457: typo: `unregocnized `

frontend/locales/en-US.json
rdmchr: Yeah, some keys could be reused and I didn't encounter any difficult conflicts.

ferferga (approved):
Thank you very much, wanted to do some minor style chores before merging (I think it was quicker if I did it myself), but life caught in

Thank you very much for your contribution and your patience to get this through!

PR #2106, opened 2023-10-08
Fix direct play stream url regression

Willy-JL:
Stream URLs are at /Videos/ not /Video/.

5e33e8e0fbe465971c92074a366a5eefa2feb9cb introduced this regression, by replacing `'Videos'` with `this.currentlyPlayingMediaType` which contains value `'Video'`.

This regression means that playing videos that have direct play available results in a 404, and the player hangs.

What has me quite confused is... this regression was introduced back in June, almost 4 months ago... how did nobody notice that playing *any* direct play movie/video was broken?

### DIFF
--- a/frontend/src/store/playbackManager.ts
+++ b/frontend/src/store/playbackManager.ts
@@ -975,7 +975,9 @@ class PlaybackManagerStore {
 
       const parameters = new URLSearchParams(directOptions).toString();
 
-      if (mediaType === 'Video') mediaType = 'Videos';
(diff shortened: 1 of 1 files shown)


ferferga:
@Willy-JL Probably nobody willing to investigate. Saw the 404 error in Vue matrix chatroom but with current life I didn't have any time to investigate. I also use this mainly for music: video playback is still quite broken because of the profile detection we use (copied from web, which has improved a lot since then).

Thank you very much!

PR #2108, opened 2023-10-11
Fixes #2085 - Allow installation in a sub-path

### LINKED ISSUE #2085: Allow installation in a sub-path
### Description of the bug

For now it seems that jellyfin-vue is only installable on a domain root (localhost:8080/, jf-vue.pages.dev) and can't be installed in a subpath (mydomain.tld/jellyfin-vue/). `/config.json` is accessed via absolute path.

See https://github.com/jellyfin/jellyfin-vue/blob/0a01f80b44072248cf771ff1f0cebba8ddf741ea/frontend/src/utils/external-config.ts#L51

According to https://vitejs.dev/guide/build.html#public-base-path, I suppose you should use `import.meta.env.BASE_URL`.

### Steps to reproduce

```
npm install
npm run build -- --base=/jellyfin-vue/
```

Configure nginx with a subdirectory
Enjoy an empty page.
See in the requests panel of Firefox: `GET | https://mydomain.tld/config.json` while all other resources are accessed via `https://mydomain.tld/jellyfin-vue/<resource path>`.

### Expected behavior

config should be retrieved via `GET | https://mydomain.tld/jellyfin-vue/config.json`

### Logs

_No response_

### Screenshots

_No response_

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.8.10

### Additional context

_No response_

DeanWard:
By setting the [`base` option ](https://vitejs.dev/config/shared-options.html#base) to `./` we enable _relative_ loading of assets and replace the current _absolute_ path resolution. This, along with removing the leading `/` when loading the external config via fetch allows the client to be installed in a sub-path at any level of nesting.

This **does not** require any changes to the env file, or any other changes by the user.

### DIFF
--- a/frontend/src/utils/external-config.ts
+++ b/frontend/src/utils/external-config.ts
@@ -48,7 +48,7 @@ function validateJsonConfig(
 export async function getJSONConfig(): Promise<ExternalJSONConfig> {
   if (isNil(externalConfig)) {
     const loadedConfig: unknown = await (
-      await fetch('/config.json', { cache: 'no-store' })
+      await fetch('config.json', { cache: 'no-store' })
     ).json();
 
     validateJsonConfig(loadedConfig);
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -23,6 +23,7 @@ import { localeFilesFolder, srcRoot } from './scripts/paths';
 export default defineConfig(({ mode }): UserConfig => {
   const config: UserConfig = {
     appType: 'spa',
+    base: './',
     define: {
       __COMMIT_HASH__: JSON.stringify(process.env.COMMIT_HASH || '')
     },


ferferga:
@DeanWard Doing a quick test, as I suspected when we talked in Matrix, this has the same problem as the absolute approach: when using history mode, it doesn't know how to locate assets. I tested using the config file of our Dockerfile. I think I didn't forget any other case and this is the only one that gave me issues (but will recheck them later)

The difference between this and the absolute approach is that the absolute approach with history mode doesn't crash when not using a subpath.

Need to go now, but can you check in the meantime if the rewrite rule I added to the original issue does something to solve this case? Do you have any other idea or improvements we can add to our webserver config in that regard? If not, maybe we should document that history mode can't be used with subpaths (or remove it altogether, which I don't care that much honestly, though I see the asthetics of the clean urls).

DeanWard:
That is very frustrating isn't it. I've just tested with history mode, and it crashes out. If we could just get vite to output 

```
<script type="module" crossorigin src="assets/index-eaf9754e.js"></script>
<link rel="stylesheet" href="assets/style-01e1b61f.css">
```
 instead of

```
<script type="module" crossorigin src="./assets/index-eaf9754e.js"></script>
<link rel="stylesheet" href="./assets/style-01e1b61f.css">
```

it would work in all cases. I'll do some testing with your suggested rewrite rules but i'd like to find a solution that doesn't require messing with server configuration so that we can avoid having to maintain docs for all the different types of servers people might try to use.

PR #2130, opened 2023-11-03
feat(ci): Update Tauri artifacts

endrl:
- Provide better tauri artifacts
- Tauri recommends to use an older ubuntu version as this is the min version where it can actually run

### DIFF
--- a/.github/workflows/quality_checks.yml
+++ b/.github/workflows/quality_checks.yml
@@ -147,7 +147,7 @@ jobs:
       matrix:
         platform:
           - "macos-latest"
-          - "ubuntu-latest"
+          - "ubuntu-20.04"
           - "windows-latest"
     defaults:
       run:
@@ -170,7 +170,7 @@ jobs:
         run: npm ci --no-audit
 
       - name: Install Linux dependencies 📦🐧
-        if: matrix.platform == 'ubuntu-latest'
+        if: matrix.platform == 'ubuntu-20.04'
         run: |
           sudo apt update -qq
           sudo apt install -y --no-install-recommends $(cat apt_packages)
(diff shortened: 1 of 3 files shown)


ferferga:
> Tauri recommends to use an older ubuntu version as this is the min version where it can actually run

Source for this? ubuntu-latest always points to the latest Ubuntu LTS version, so we can be sure they're always stable? [See here](https://github.com/actions/runner-images?tab=readme-ov-file#available-images)

ferferga:
@endrl I see what you mean now. I think in that case the best course of action would be to statically link our executable, so it doesn't have any hard system dependency. What do you think about this? That way, we're not restricted at all because of the CI environment (not providing older images) and we're sure the app is universally compatible. 

I wonder if something along the lines of [this](https://stackoverflow.com/questions/31770604/how-to-generate-statically-linked-executables) would work.

endrl:
I had a look at the AppImage, it packages `libwebkit2gtk-4.0.so.37` which is the latest. See also [tauri's list](https://tauri.app/v1/references/webview-versions/#linux). So it's just good, you get compatibility and bleeding edge webview at the same time.

ferferga:
@endrl Can you check this now with the latest artifact from master? See [this](https://github.com/tauri-apps/tauri/issues/1355#issuecomment-1710278028), it looks like it's fixed now in the default target? (so their documentation is now deprecated)

I've been also playing with the cross as a runner (as suggested in that thread) and some optimizations on [this branch](https://github.com/jellyfin/jellyfin-vue/tree/static-tauri) so if the above doesn't work maybe we have more luck with cross.

I've run a workflow in that branch with my latest commit so you can [check it out](https://github.com/jellyfin/jellyfin-vue/actions/runs/7982165902) too!

Thank you very much in advance and sorry for the delay.

endrl:
The outcome of the issue is basically can't fix? https://github.com/tauri-apps/tauri/issues/1355#issuecomment-1031064919
Cross 2.5 links to glibc 2.31 (which is ubuntu [20.04](https://launchpad.net/ubuntu/focal/+source/glibc)), so you can also accept the pull request and skip docker at this point :)

I think github will maintain the image until the kernel is out of [support](https://ubuntu.com/about/release-cycle), but that's just a guess. And at this point you can still fallback to cross 2.5

`rustup default stable` simply sets a beta/nightly toolchain back to stable, has nothing to do with libc.

ferferga:
Hello @endrl 

I'm sorry for letting this go waiting for so long to eventualy not even getting it merged. Although it looks like a minor thing, keeping track of *yet another dependency* (the ubuntu images) is an additional burden we need to handle just for the sake of people running Tauri (which I estimate are just a bunch of fulks, given it's provided in a *best effort* basis) + outdated systems.

Additionally, I'm used to the JS way to do things, Rust (and all around Tauri really) is still mostly black magic for me, which means that there's additional burden every time something happens with it (time I could dedicate first to advance in the frontend).

I really hope that in future our Tauri app could have much more love than it has now, so these kind of things are worth the effort. Even better, hopefully in future Tauri versions this is much easier (or even unnecessary to do)!

PR #2135, opened 2023-11-07
feat(seasons): Add links back to a series and season from an episode.

BrianCArnold:


BrianCArnold:
This currently only links back to the series, going to work on seasons this evening if I can.

frontend/src/components/Item/SeasonTabs.vue
ferferga, line 1: @BrianCArnold What's the reasoning behind the changes in this file?

ferferga (approved):
I'm merging leaving SeasonTabs file behind since my concerns have not been addressed

BrianCArnold:
Somehow managed not to see your review, will have to look back at the pre-fp to check what they were.

BrianCArnold:
Not at my computer at the moment, but at a glance the main difference is that it pulls the data from each season in parallel rather than requesting season 1, waiting for it to finish, then requesting season 2, waiting for that to finish, then [...]

Higher latency connections also exacerbate this delay.

If one's connection to the server is high latency and the server is slow to respond due to a very large library, a request cycle per season can take up to 0.5s. This means that a show with 30 seasons (The Simpsons, for example) could take up to 15 seconds to load. 

By requesting the seasons in parallel, the load time is reduced by a factor of the number of seasons for that show.

PR #2189, opened 2024-01-02
refactor: strict type only imports

ferferga:
This is the recommended way to import types in TypeScript and be side-effect safe.


tschwery:
The "strange spaces" in the French translations should be kept. They are "Non-Breakable-Spaces" that are mandatory before some punctuation like `?` or `:`.

Issue #2129, opened 2023-11-02
Played status not updating on episode


EffakT:
### Description of the bug

When watching a series, after finishing an episode and the next episode starts playing. When navigating back to the series, the completed episode is not marked as played.
The progress is reset to 0, and looks unplayed

I'm running a jellyfin/jellyfin-vue:unstable docker container, which I update everyday at 2am NZT.

Debugging the code, it seems that the reportPlaybackStopped function is running with a currentTime of 0, therefore the server is setting the progress to 0.
I'm guessing that the reportPlaybackStopped is running after the reset of the player/loading of the new video, resulting in the currentTime being 0.

### Steps to reproduce

Start watching a new series (not marked as played). Finish the episode (or fast-forward to the end) to trigger the next episode.  Exit this new video, and navigate to the episode you just watched, it is not marked as played.

### Expected behavior

Episode should be marked as played when complete.

### Logs
Note, I've replaced the host with a random IP.

```text
TypeError: MediaSession.setPositionState: Invalid position 2653.005333, it can't be negative or greater than duration
    Cte https://192.168.88.54/assets/index-1167ecfb.js:77
    go https://192.168.88.54/assets/index-1167ecfb.js:1
    Ki https://192.168.88.54/assets/index-1167ecfb.js:1
    c https://192.168.88.54/assets/index-1167ecfb.js:1
    run https://192.168.88.54/assets/index-1167ecfb.js:1
    g https://192.168.88.54/assets/index-1167ecfb.js:1
    xb https://192.168.88.54/assets/index-1167ecfb.js:1
    Y https://192.168.88.54/assets/index-1167ecfb.js:1
    Re https://192.168.88.54/assets/index-1167ecfb.js:1
    run https://192.168.88.54/assets/index-1167ecfb.js:1
    update https://192.168.88.54/assets/index-1167ecfb.js:1
    j https://192.168.88.54/assets/index-1167ecfb.js:1
    M https://192.168.88.54/assets/index-1167ecfb.js:1
    p https://192.168.88.54/assets/index-1167ecfb.js:1
    Re https://192.168.88.54/assets/index-1167ecfb.js:1
    run https://192.168.88.54/assets/index-1167ecfb.js:1
    update https://192.168.88.54/assets/index-1167ecfb.js:1
    go https://192.168.88.54/assets/index-1167ecfb.js:1
    pw https://192.168.88.54/assets/index-1167ecfb.js:1
[...]

Issue #2028, opened 2023-06-01
Memory leak


ferferga:
### Description of the bug

There's a memory leak with DOM elements not being detached correctly in master when switching between pages, so long-running sessions of the client might make the browser consume a lot of memory. 

I believe this is caused by a combination of `Suspense` (which is still beta) and `Transition`, as showcased in https://github.com/antfu/repro-vue-suspense-transition-memleak

I'm still trying to isolate all the dependencies as much as possible to triage where this is coming from exactly, since this might not be a Vue bug.

ferferga:
This comes with huge delay, but there have been a lot of advancements in this regard.

Though Vue still has [many Suspense](https://github.com/vuejs/core/issues?q=is%3Aissue+is%3Aopen+transition+suspense) issues when combined with transitions, I'm almost sure the exact source of our leak is on Vuetify's side, since the antfu's repo was GC'ed correctly and many other complex demos with Suspense + Transition did too.

While profiling in our client, all the references pointed me to ``VBtn``, which lead me to focus on Vuetify only. I posted this on Vuetify's Discord the 08/06/2023 (DD/MM/YYYY):

<details>
For context, I'm using Edge DevTools

1. Open this: https://play.vuetifyjs.com/#eNqlUL1uwjAQfpWrF1qJxAOdUEB06wN0wwwhuVAX/8k2qSqUd6/tBBTCgsR29919P3fbM/kwJm9PSJak8CiNKD2umQIo2qw0JpWpqbTyJVdoByiBe6+gzXizYsTrw0EgI7CGTxRCQ0HTfLq+qQSvjlcCrODlQl3DV4/dUmN36x6RPlxBR6FD6yrLjQeH/tRn59Jo6+EMFhvooLFawizcO4vDoOo8XIOEldemFA7fonAvFUTInPQqmSxN/uO0Cs86Rz4bBo6RJSQkYkE99ox8e2/cktKqVoFWo+CtzRV6qoykm7BG7Ul5LjGrtdws8kX+Tmvu/BjO0clsb/WvQxtEGJmPbDxv/h6wGjaTw6J3GKBMlHsXLe6kafBr0WYWVY0W7aMXTWjjqyaju8uie8dUF/4tuDq6yasrl968vYR86uQkFpV20bCbk5QzOCcC2f0DuX8JMw==

2. Open the performance monitor. See the DOM node count and note it.
3. Toggle the button a few times using the toggle button.
[...]

PR #2201, opened 2024-01-08
feat,fix,refactor&perf: offline navigation, full item reactivity, simplifications and bug fixes

### LINKED ISSUE #1803: Logos are sometimes ignored from the main page slide content
### Description of the bug

Some logos are ignored from the slide content thing on the main page

Sans titre 36

Sans titre 38

Both of these media contains a logo, only one displays it, it's also displayed on Jellyfin's media info page.
Logo size are identical.

### Steps to reproduce

Scrolls between media infos

### Expected behavior

Logos that exists should be displayed on the slider thing

### Logs

_No response_

### Screenshots

_No response_

### Platform

Windows

### Browser

Chrome

### Jellyfin server version

10.8

### Additional context

_No response_

### LINKED ISSUE #2194: Playback is not working
### Description of the bug

Playback is not working in current master. This is due to ongoing refactor efforts (that had to start somehow) with the items store and api fetching that I hope to finish soon.

ferferga:
I've been out a lot due to uni, but I'm not with empty hands either :):

Fixes #2194 #2129 #2028 (this time completely for real!) #1921, possibly fixes #1803 #1717 #801 #762

**Development improvements**

- Fully new data fetching. Created the new ``useApi`` and ``useBaseItem`` composables, deprecate ``userLibraries`` and ``itemsStore``, created ``apiStore``. Both composables are fully documented in the codebase, but here's a brief explanation:

   * Depending on the response type you must use one of another. TypeScript typings will help you use the correct one, the incorrect one will only return ``undefined``
   * Fully reactive, both in parameters and methods (which simplified a lot the library pages and allowed us to do some cool things, like in MarkPlayedButton).
   * The Api store has 2 reactive caches:
        * Map of the id of an item with the item itself
        * Map of a function of the api, the arguments used and the response. If the response is BaseItemDto based, it will do a lookup on the first map.
   * The composables will always return cached data first (if it exists) and then, in the background, refresh it. Given they return a ComputedRef, the data will be updated automatically.
   * If there's cached data for the request a page is performing, the page will load instantly and navigation will not be blocked.
   * All the item fields and images will always be fetched, for consistency
   
- Fix Suspense not really blocking navigation before resolve.
- Fix development build not working
- Fix the memory leak (root source was the items store, which we couldn't properly fix back at the time). The most memory and CPU intensive parts of the application are still the library pages, but even while on them, the GC fires with light scrolling, so it looks like that all our data structures are easily GCeable.

**User facing improvements**

- Navigation is much quicker now
     * Although we do a best-effort lazy load, full library pages are still slow to load due to the server constraints.
- More CPU-efficient splashscreen
- Reactive everywhere (we need further server support to get updates to other events, like editing metadata descriptions)
- Best-effort offline support: I added logic for caching pending responses and executing them afterwards, but still the client is not ready to be fully offline (in fact, it can only work offline if previously it has been online), so this is still WIP.
- Fix loading bar endlessly stuck.

**Some videos**:

https://github.com/jellyfin/jellyfin-vue/assets/10274099/d88dd62e-e11c-4d80-8873-f7431070d15b

https://github.com/jellyfin/jellyfin-vue/assets/10274099/a7f8c0d1-6e7c-45d0-8b81-3ee84250c6a7

https://github.com/jellyfin/jellyfin-vue/assets/10274099/6abc1c44-eab7-44de-892f-1cb5029c637b


Extarys:
The webapp feels a bit more responsive indeed, thanks for all your effort and I hope uni is treating you well. I'm still curious about an odd duck though.

Before/master(?):
image


With this branch:
image


I'm not sure where the big 25MB request is coming from. All those Items request looks the same to me :joy: I see a lot of smaller requests (50 items) which is great, but then _the_ chunk loads.

I cleared the network tab then hit refresh on my movie library and filtered by XHR requests on both tests.

```
https://domain/Items?userId=xxx&startIndex=50&recursive=true&sortOrder=Ascending&parentId=f137a2dd21bbc1b99aa5c0f6bf02a805&fields=AirTime&fields=CanDelete&fields=CanDownload&fields=ChannelInfo&fields=Chapters&fields=ChildCount&fields=CumulativeRunTimeTicks&fields=CustomRating&fields=DateCreated&fields=DateLastMediaAdded&fields=DisplayPreferencesId&fields=Etag&fields=ExternalUrls&fields=Genres&fields=HomePageUrl&fields=ItemCounts&fields=MediaSourceCount&fields=MediaSources&fields=OriginalTitle&fields=Overview&fields=ParentId&fields=Path&fields=People&fields=PlayAccess&fields=ProductionLocations&fields=ProviderIds&fields=PrimaryImageAspectRatio&fields=RecursiveItemCount&fields=Settings&fields=ScreenshotImageTags&fields=SeriesPrimaryImage&fields=SeriesStudio&fields=SortName&fields=SpecialEpisodeNumbers&fields=Studios&fields=BasicSyncInfo&fields=SyncInfo&fields=Taglines&fields=Tags&fields=RemoteTrailers&fields=MediaStreams&fields=SeasonUserData&fields=ServiceName&fields=ThemeSongIds&fields=ThemeVideoIds&fields=ExternalEtag&fields=PresentationUniqueKey&fields=InheritedParentalRatingValue&fields=ExternalSeriesId&fields=SeriesPresentationUniqueKey&fields=DateLastRefreshed&fields=DateLastSaved&fields=RefreshState&fields=ChannelImage&fields=EnableMediaSourceDisplay&fields=Width&fields=Height&fields=ExtraIds&fields=LocalTrailerCount&fields=IsHD&fields=SpecialFeatureCount&includeItemTypes=Movie&sortBy=SortName&enableImageTypes=Primary&enableImageTypes=Art&enableImageTypes=Backdrop&enableImageTypes=Banner&enableImageTypes=Logo&enableImageTypes=Thumb&enableImageTypes=Disc&enableImageTypes=Box&enableImageTypes=Screenshot&enableImageTypes=Menu&enableImageTypes=Chapter&enableImageTypes=BoxRear&enableImageTypes=Profile&enableTotalRecordCount=false&enableImages=true
```

ferferga:
@Extarys Yes, this is intended. I'll explain you how it works:

Before (Vue 3 migration), all pages loaded its own data on mount, meaning components (=pages in any context, everything is a component) could render empty and we had to add their own loading state.

Vue 3 introduced [Suspense](https://vuejs.org/guide/built-ins/suspense.html#suspense), which basically allows you to keep the old component while a new one is resolving. It was a matter of moving the request from it's position to the top level. However, in order to ease the migration, all the redundant logic for handling updates was left untouched and we were still caching data like in the old way, but only 2 components were using it, so it was a waste of memory. The composables merge all of that into a single statement.

This means that now, (besides a few metadata pages, which will be refactored later), like all pages, the library page also makes use of the cache and it will not render until the data is available (if no cache is available)

This is how it works for a single library (every library will have this process repeated for it's own):

- The user navigates to the library page and navigation is blocked while 2 requests happen: one for the library itself and another for the 50 first items of that library (the 50 numbers is hardcoded).
- Both requests are cached and the composable returns reactive references to the cache.
- On component mount (which remember, only happens after both 1st requests are resolved), we set the limit to undefined, so all the items (except the ones that we already have) are fetched. The composable picks the change of limit and automatically triggers a new request for all the data. The reactive cache still points to the first request
- The server answers and the reactive reference is updated to the new request, so all the items of that library appears now.

When the user navigates back to the same page:
- The request for the first 50 items resolves immediately since it's cached. The navigation is not blocked
- In the background, the first 50 items request is done again to refresh the cache.*
- The request for all the items also resolves immediately since it's cached. All the data is shown rightaway (without the *Showing 50 items. Loading more...* message)
- In the background, the request withoutr limits is done again to refresh the cache.

As you can see, even for the first 50 items the server is slow to answer. At the current state, there's no point in me implementing more logic to handle the scrolling position and query: the current method of doing the fetch twice is already worse for the server, since internally it makes the DB query for all, makes all the filtering in C# and then limits (is a little bit faster because the last steps of the filtering process is faster if it has to do less iterations of course) but the UX is much better because the user "feels" the page loads before.

The only solution for this at this current state would be for the server to provide an endpoint that just returns item ids, so we only need to update those that are in the screen. However, that won't happen, since that's an stopgap for the real problem.

Users expect to have all the content of their libraries, so of course I won't add artificial limits for this (current web is filled with those and people complained, with reason to in my opinion, I even dislike that I'm only able to play max 100 items at once!).

If data usages are a concern, use search.

EffakT:
Was going to test #2129, and noticed when you attempt a login with incorrect password it shows the error as expected, however does not reset the "loading" state. This works as expected on nightly (aside from the infinite loading bar).

Extarys:
@ferferga Thank you for taking the time to write all this up! Great explanation and it makes total sense. ;)

ferferga:
@EffakT Good catch! Try now please, both issues should be fixed

ferferga:
@EffakT Good catch! Try now please, both issues should be fixed

EffakT:
@ferferga Not sure if its related to the last change, it doesn't seem to be loading the video information (video/audio/subtitles) on first load of a episode page.
Triggering an ajax call on the page (clicking tick icon, or the heart icon) makes the fields appear. I can see the Items query runs again returning a single item.

Though, can confirm both the login and play status issues are sorted.

ferferga:
@EffakT I fixed it as well, not caused by this PR but no hassle to include it either. Shipping now!

PR #2205, opened 2024-01-10
fix: Add immediate tag to MediaSources watcher

EffakT:
When viewing an episode/item page, the MediaSources watcher does not trigger instantly, causing the Video, Audio, and Subtitles dropdowns to not show until the MediaSources are updated again.
Watchers by default are lazy, when in this instance it should be eager and trigger when it attaches.

Adding the immediate tag makes this watcher eager, resolving the issue.

### DIFF
--- a/frontend/src/pages/item/[itemId].vue
+++ b/frontend/src/pages/item/[itemId].vue
@@ -371,5 +371,5 @@ watch(() => item.value.MediaSources, () => {
   if (item.value.MediaSources && item.value.MediaSources.length > 0) {
     currentSource.value = item.value.MediaSources[0];
   }
-});
+}, { immediate: true });
 </script>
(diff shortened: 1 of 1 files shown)


ferferga:
@EffakT Sorry for pushing to the branch, but I think it was quicker doing it than entering into the review loophole.

With the migration to Vue 3 and composition API, I discovered that previously we abused watchers way too much for things that can be computed! Computed properties are awesome and paired with setters they're usually underlooked by beginners (like we were at the beginning).

* They're explicit: you know exactly when and how the property change
* They're [faster](https://github.com/vuejs/core/pull/5912) (and [this](https://twitter.com/johnsoncodehk/status/1740627870295437812/photo/1)). I know that there are a lot of quirks here and then, but as we go, I try to do as many microoptimizations as possible.
* You can even handle the previous value more easily than with a watch since Vue 3.4: ``computed((previous) =>``

There are a lot of places where we still have this kind of pattern, since we focused past in the day in getting Vuetify and Vue 3 up and running as soon as possible and a lot of refactors and simplifications that previously were impossible due to Options API are now possible.

If you don't want this to be your first and only PR, this is your chance :)
I'll approve but wait for your feedback so you can confirm me that everything works as with the immediate watcher.

ferferga (approved):

EffakT:
@ferferga  No worries, was just in a bit of a hurry this morning so quickly patched it so it was at least working.
Using computed is absolutely a better idea. Though in the past I've run into issues when I had to use watchers instead of computed due to data structure issues (guessing this is one of the quirks you are talking about).

Looks like this is working as expected now.

PR #2218, opened 2024-01-24
chore(readme): add :unstable to docker run

tippfehlr:
The current command, `docker run -d -p 8080:80 ghcr.io/jellyfin/jellyfin-vue`, simply prints

```
docker: Error response from daemon: manifest unknown.
```

As long as there’s no latest (stable) release the “default” branch should point to unstable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,9 @@ This is an alternative client for Jellyfin based on Vue.js. It might not be feat
 
 # Usage instructions for end users 👨‍👩‍👧‍👦
 
-*The following instructions will get you up and running quickly with builds
-from the current `master` branch, which are bleeding edge by default*
+*The following instructions will get you up and running quickly with
+bleeding-edge builds from the current `master` branch.
+There are no stable releases yet.*
 
 ## [Hosted instance 🌍](https://jf-vue.pages.dev/)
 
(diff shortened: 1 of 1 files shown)


ferferga (approved):

ferferga:
@tippfehlr I added a minor improvement to cover both CF deployment and Docker with a single sentence, please let me know what do you think.

I'll merge if you think it's fine.

tippfehlr:
@ferferga what do you think of this version?
Obviously remove this again if there is a stable release.

I also fixed a piece of grammar for the warning

ferferga:
@tippfehlr Great, thank you! (I'm not native so this kind of stuff sometimes goes unnoticed to me)


As for the stable releases, sure thing (although I'm not sure if instructions for using a pinned version should be in the deployment section of the wiki, since I see the readme as something that should be straight to the point).

tippfehlr:
@ferferga If pinned versions make sense I think they should be rather defined by the developer than the user (-> releases, latest tag). However, that probably doesn’t make sense for this project as of now, I can’t tell.

I agree that pinning versions doesn’t need be mentioned in the readme as it’s not specific to this project but rather basic docker knowledge (and docker containers don’t update automatically anyway). \
Can for sure be described in the wiki. I’m always happy to find generic useful information in wikis.

Thanks for the quick merge :heart:

PR #2219, opened 2024-01-25
fix: Report playback progress at regular interval

### LINKED ISSUE #2216: Playback progress is not reporting until close of playback view
### Description of the bug

Playback progress is not reporting back to the server until you close the playback view.
This means that when skipping ahead or behind, the server does not update.
If you don't close the playback view before closing the browser, the progress is not reported to the server, loosing the progress.

### Steps to reproduce

Play a video for a while, no request is send to the server to update progress.
Close the browser without closing the playback view, the server does not know the progress and shows as unwatched.

### Expected behavior

At regular interviews while playing (or on timeupdate, maybe throttled as to not spam), on pause, unpause, volumechange, the current progress should be reported to the server. 
This means you will never loose a large amount or all progress if the session is lost for whatever reason.

### Logs

_No response_

### Screenshots

_No response_

### Platform

macOS

### Browser

Firefox

### Jellyfin server version

10.8.12

### Additional context

_No response_

EffakT:
The UI was only reporting playback on close, meaning when a playback session was lost due to a crash, or server close. Progress would be lost.
The playback in the dashboard would also not see progress if the user was to skip ahead or behind.

This is resolved by adding a watcher to the "currentTime" which triggers the `reportPlaybackProgress` function. This already has a throttle to prevent constant updates (See `_progressReportInterval` variable).

This could be extended in order to support pause/unpause, However not sure the best way to handle this (do we need to send different events to tell the server the media is paused, or is this handled by the isPaused attribute).

Additionally added .history to gitignore due to vscode's local history.

Resolves #2216

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,7 @@ frontend/dist
 
 # IDE / Editor
 .idea
+.history
 
 # Service worker
 sw.js
--- a/frontend/src/store/playbackManager.ts
+++ b/frontend/src/store/playbackManager.ts
@@ -1270,6 +1270,18 @@ class PlaybackManagerStore {
     }, { flush: 'sync' });
 
     /**
+     * Report progress playback when current time changes
+     */
+    watch(
+      () => this.currentTime,
+      async () => {
+        if (this._pendingProgressReport) {
+          await this._reportPlaybackProgress();
+        }
+      }
+    );
+
+    /**
      * Report playback stop when closing the tab
      */
     useEventListener('beforeunload', async () => {


ferferga:
@EffakT I think the issue is the callback flush that I added here:

https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/store/playbackManager.ts#L1263

This is the only change that happened to playback reporting, for some reason this passed type checking.

Though I want to refactor and simplify all of this (still thinking how) as this is not the first time this happens and this has been a very fragile behaviour for a while. 

But I understand you want this fixed in the meantime, but please check if it works without the change in the flush timing.

EffakT:
@ferferga 
I've run a quick test, and it seems before the flush change, it will report progress on player state change (pause, play, fastforward), however not on on normal play.
This needs to trigger on normal play (or at a regular interval) in order for the server to update state. As if you loose the browser session, it won't report the playback.

ferferga:
Hello @EffakT

Apologize for taking too long to deliver. I opened #2232 to refactor the whole logic and do throttling in the function itself. I'm closing this PR in favor of the another one. Given you found the issue, a review from you would be highly appreciated as always! :P

I'm super sure this was working before without the watchEffect but there is also extra logic that changed (I tested an old deployment just to make sure indeed!)

PR #2232, opened 2024-02-22
refactor: playback reporting throttling

### LINKED ISSUE #2216: Playback progress is not reporting until close of playback view

ferferga:
Fixes #2216

### DIFF
--- a/frontend/src/store/playbackManager.ts
+++ b/frontend/src/store/playbackManager.ts
@@ -19,7 +19,7 @@ import { getItemsApi } from '@jellyfin/sdk/lib/utils/api/items-api';
 import { getMediaInfoApi } from '@jellyfin/sdk/lib/utils/api/media-info-api';
 import { getPlaystateApi } from '@jellyfin/sdk/lib/utils/api/playstate-api';
 import { getTvShowsApi } from '@jellyfin/sdk/lib/utils/api/tv-shows-api';
-import { useEventListener } from '@vueuse/core';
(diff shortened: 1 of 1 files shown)


EffakT:
Used it for about 2 hrs this morning,seems to be working as expected now.
One thing, it might be worth reporting to the server on "Pause" and "Unpause", as I believe the server uses this parameter to handle inactive streams.

ferferga:
@EffakT This is already implemented, though now it's throttled. See [here](https://github.com/jellyfin/jellyfin-vue/blob/8a4d841b8f12d156fbd9e83c33ec034384c1d437/frontend/src/store/playbackManager.ts#L486)

I'm fixing that in my next push and merging.

PR #2242, opened 2024-02-29
fix: correctly trigger showPassword

qwerzl:
This is a very, very simple edit. As the icon is added as `inner-icon`, the icon event should also be `click:append-inner`.


ferferga:
@qwerzl Thank you very much for your contribution! 💖

PR #2256, opened 2024-03-09
Fixed stretched actor images

PixyTM:
Fixed images that where stretched with object-fit:cover so that they keep the original image ratio

### DIFF
--- a/frontend/src/components/lib/JImg.vue
+++ b/frontend/src/components/lib/JImg.vue
@@ -74,6 +74,7 @@ watchImmediate(() => props.src, () => {
 .j-img {
   width: 100%;
   height: 100%;
+  object-fit: cover;
 }
 
 div.j-img {


ferferga:
Hello @PixyTM thank you very much for your contribution!

Can you tell me where those images are stretched? Or attach an screenshot?

Because this might be the expected behaviour of the component and previously (when we used VImg) it was working by chance.
Note how this component has a ``cover`` prop that uses a ``div`` instead of an ``img`` element and the div has a ``background-size: cover`` in it's styles, so maybe **it's more fitting to pass** the cover prop in the parent?

PixyTM:
Hi, yes, I completely forgot to attach the images, sorry about this.
It affects the actor images on a movie/show detail page and on the person detail page. I also noticed it has a tiny impact on some posters when they are in a list (for instance in the last two screenshots) (1 or 2 pixel difference)

before
after

before2
after2

ferferga (approved):

PR #2266, opened 2024-03-18
feat(playback): Add playback speed option

### LINKED ISSUE #2226: quality and speed selector disabled
### Description of the bug

I can’t select or view the quality (or speed) of media

### Steps to reproduce

1. play any video
2. click on playback settings

### Expected behavior

- Current quality and speed is displayed
- quality and speed can be selected

### Logs

_No response_

### Screenshots

image

### Platform

Linux

### Browser

Firefox

### Jellyfin server version

10.8.13

### Additional context

```text
also tested on Brave
```

JVT038:
This PR adds the option of selecting different speeds.
The options are 0.5, 0.75, 1, 1.25, 1.5 and 2 (just like in the jellyfin-web repo).

Also, it partially closes #2226


frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
```diff
@@ -62,9 +62,12 @@
...
             <VCol :cols="8">
               <VSelect
+                v-model="playbackManager.playbackSpeed"
                 density="comfortable"
-                disabled
-                hide-details />
+                hide-details
+                :items="playbackSpeed"
+                item-title="title"
+                item-value="speed" />
```
ferferga, line 70: What do you think about using a [Combobox](https://vuetifyjs.com/en/components/combobox/#multiple-combobox) instead? So we provide a good UX by giving a predefined selection of speeds but also freedom of choice in a "single package".

Instead of 1x, I'd add a "Normal" label to that option (just like YouTube does).

In that case, since you're going to use vue-i18n's t function for translating, you must wrap this array in a Vue's ``computed`` so language changes are reactive.

ferferga (commented):
Thank you very much for your first contribution! Just one minor thing I'd like to discuss with you first:

frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
JVT038: I think we should keep using VSelect instead of the ComboBox, because after trying out ComboBox, I've encountered some validation(?) issues and changing the speed is no longer possible for some reason.
If we just keep the predefined speeds, then we don't have to deal with validating the user input.
If you want to try it yourself, here's a patch I made:
```
diff --git a/frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue b/frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
index bcaada01..3ae26b26 100644
--- a/frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
+++ b/frontend/src/components/Buttons/Playback/PlaybackSettingsButton.vue
@@ -61,13 +61,14 @@
               <label>{{ $t('speed') }}</label>
             </VCol>
             <VCol :cols="8">
-              <VSelect
-                v-model="playbackManager.playbackSpeed"
+              <VCombobox
+                v-model.number="playbackManager.playbackSpeed"
                 density="comfortable"
                 hide-details
                 :items="playbackSpeed"
                 item-title="title"
-                item-value="speed" />
+                item-value="speed"
+                :rules="[validationRule]" />
             </VCol>
           </VRow>
           <VRow align="center">
@@ -90,34 +91,44 @@
 </template>

 <script setup lang="ts">
+import { computed } from 'vue';
 import { playbackManager } from '@/store/playback-manager';
 import { playerElement } from '@/store/player-element';
+import { isNumber } from '@/utils/validation';

 const menuModel = defineModel<boolean>();
-const playbackSpeed = [
-  {
-    title: '0.5x',
-    speed: 0.5
-  },
-  {
-    title: '0.75x',
-    speed: 0.75
-  },
-  {
-    title: '1x',
-    speed: 1
-  },
-  {
-    title: '1.25x',
-    speed: 1.25
-  },
-  {
-    title: '1.5x',
-    speed: 1.5
-  },
-  {
-    title: '2x',
-    speed: 2
-  }
+const playbackSpeed = computed(() => {
+  return [
+    {
+      title: '0.5x',
+      speed: 0.5
+    },
+    {
+      title: '0.75x',
+      speed: 0.75
+    },
+    {
+      title: 'Normal',
+      speed: 1
+    },
+    {
+      title: '1.25x',
+      speed: 1.25
+    },
+    {
+      title: '1.5x',
+      speed: 1.5
+    },
+    {
+      title: '2x',
+      speed: 2
+    }
+  ];
+});
+
+const validationRule = [
+  (value: string) => () : true | string => value.length > 0 || 'Required',
+  (value: string) => () : true | string => isNumber(value) || 'The input has to be a number',
+  (value: string) => () : true | string => Number.parseFloat(value) > 0 || 'The input has to be greater than zero.'
 ];
 </script>
``` 

I'd like to change '1x' to 'Normal', but I don't know how to add translation strings for all the languages.

ferferga:
@JVT038 Please test with my latest changes and tell me if you find something out of order.

JVT038:
I have tried your changes and noticed that I often see NaN appear while typing something. 
Additionally, I think there should be more feedback to the user if they've typed an invalid string. Something like 'The speed must be between 0 and 16'.
There's also this x in the bottom that doesn't really do anything when you click on it, so I think that should be removed.

JVT038:
Some issues I've encountered:

1. The speed input is 'Normal' by default. When I press on the backspace to type something else, NaN appears.
2. When I enter a value, I'd like to press 'Enter' and have the value added. Instead, the cursor stays focused on the speed input and the speed doesn't change until I un-focus manually from the input.
3. Because the speed value is 'Normal' by default (which is actually the integer 1), I can't type a speed with from 1 - 2. When I type 1, 'Normal' automatically autocompletes the input and I can't type something like 1.3 or 1.2. Also, values from 10 (such as 11, 12, 16, etc.) can't be typed either, because they begin with a 1.
4. When my cursor is focused on the input, an 'X' appears next to the collapse / expand arrow, but it doesn't do anything when I  click on it.

ferferga:
@JVT038 Please give it another go

JVT038:
It works fine for now.

Only thing I noticed is if I enter a high speed (such as 10x), my PC can't keep up with transcoding, so the movie no longer plays. The subtitles keep going, however. 

But I don't think this is an issue of jellyfin-vue right?

ferferga:
@JVT038

Playback speed is fine here, at least on Edge. Nothing we can change and it depends on the server. 

The x it's for denoting the multiplier, clearable fields in material design are like this:

https://vuetifyjs.com/en/components/text-fields/#clearable

Anyway, you think it should be removed?

JVT038:
I think the x should either be right next to the speed (so that it literally says 3x instead of 3                          x) or the x should be removed.

ferferga:
@JVT038 What about in the left side?

ferferga:
I believe that, at the left, is pretty clear what the intention is, so I'm merging this rightaway.

tippfehlr:
Hi, thanks for implementing this! One thing I noticed is that on mobile when clicking on the selector the keyboard pops up (because of the number input). For me that was somewhat unexpected and I just clicked it away instantly.
I love the freedom to choose any on pc, but would lock it to the options on devices without hardware keyboards. (If not detectable, maybe mobile, idk)

PR #2271, opened 2024-03-21
fix: defaults merging, verify token is really deauthed

### LINKED ISSUE #2269: Refreshing Jellyfin-Vue while logged in breaks in Firefox
### Description of the bug

Jellyfin-Vue breaks when you reload the page on Firefox while logged in.
The splash screen shows the Jellyfin logo and that's it. Nothing loads anymore.

### Steps to reproduce

1. Log in while using Firefox
2. Refresh the page (F5 or CTRL + R)
3. See error 

### Expected behavior

I expect Jellyfin-Vue to function properly after reloading the page 

### Logs

_No response_

### Screenshots

I'm stuck on this splash screen:
afbeelding

### Platform

Windows

### Browser

Firefox

### Jellyfin server version

10.8.13

### Additional context

```text
When I use Chrome instead, this error does not occur.

This also occurs in troubleshoot mode, which is a 'safe mode' in Firefox that disables all extensions. So it's not caused by extensions either.
```

ferferga:
Possibly fixes #2269

### DIFF
--- a/frontend/src/plugins/remote/auth/index.ts
+++ b/frontend/src/plugins/remote/auth/index.ts
@@ -55,10 +55,10 @@ class RemotePluginAuth {
   }
 
   public get currentUserToken(): string | undefined {
-    return this.getUserAccessToken(this.currentUser);
+    return this._getUserAccessToken(this.currentUser);
   }
 
-  public readonly getUserAccessToken = (
+  private readonly _getUserAccessToken = (
     user: UserDto | undefined
   ): string | undefined => {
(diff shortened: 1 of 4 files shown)


ferferga:
@JVT038 @EffakT Could reproduce this successfully, so I'm merging this rightaway to get this hotfix in ASAP as I'm sure this time it's fixed for real (sorry for the delay, had some uni exams).

The logic was incorrect when the browser didn't had support for the [Network Information API](https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API) which is the case for Firefox (it should have been an OR instead of an AND).

PR #2360, opened 2024-06-05
feat(subtitles): settings & custom element for external subtitles

seanmcbroom:
This pull request introduces new features and improvements related to subtitle settings and displaying subtitle tracks. The main changes include:

Subtitle Settings Page:
- Added a new settings page dedicated to subtitle customization.
- Users can now select their prefered subtitle font from a list of available fonts.
- Font size and position from the bottom of the screen can be adjusted.
- Options to enable or disable subtitle backdrop and stroke.

Custom Subtitle Track Element:
- Integrated a custom track element to handle external subtitle files, supporting both VTT and ASS/SSA formats.
- Added logic to parse and sanitize subtitle files, ensuring harmful tags are removed and specific tags are replaced with styled `<span>` tags for safe rendering


frontend/src/store/player-element.ts
```diff
@@ -105,73 +121,150 @@ class PlayerElementStore extends CommonStore<PlayerElementState> {
...
-   * It then find the potential index of the applied VTT sub
-   * Or the PlaybackExternalTrack object of the potential SSA sub
-   *
-   * If external and VTT, it shows the correct one
-   * If external and SSA, it loads it in SO
-   *
-   * If embedded, a new transcode is automatically fetched from the playbackManager watchers.
+   * Logic for applying custom subtitle track.
+   * 
+   * Returns flase if subtitle devliery method isn't external
```
ferferga, line 140: What's flase?

frontend/src/utils/subtitles.ts
```diff
@@ -0,0 +1,162 @@
...
+* Parses a VTT (WebVTT) file from a given URL
+* Extracts dialogue lines with start and end times, and text content.
+* 
+* Converts specific tags to styled <span> tags, 
+* and sanitizes the content by removing harmful tags.
+*/
+export async function parseVttFile(src: string) {
+    return fetch(src)
+      .then(response => response.text())
+      .then((vttText) => {
```
ferferga, line 47: * Again, no then callbacks.
* Why not use axios?

frontend/src/utils/subtitles.ts
```diff
@@ -0,0 +1,162 @@
...
+                .replace(/<\/em>/g, '</span>')
+                .replace(/<strong>/g, '<span style="font-weight: bold;">')
+                .replace(/<\/strong>/g, '</span>')
+                .replace(/<mark>/g, '<span style="background-color: yellow;">')
+                .replace(/<\/mark>/g, '</span>')
+                .replace(/<script.*?>.*?<\/script>/g, '')
+                .replace(/<iframe.*?>.*?<\/iframe>/g, '')
+                .replace(/<object.*?>.*?<\/object>/g, '')
+                .replace(/<embed.*?>.*?<\/embed>/g, '')
+                .replace(/<style.*?>.*?<\/style>/g, '');
```
ferferga, line 87: This is unnecessary since we have DOMPurify and ``JSafeHtml`` components.
I made a PR at https://github.com/jellyfin/jellyfin-vue/pull/2361 where I cherrypicked some local improvements I had for that component in my ongoing work into the MediaCapabilities migration (see [project](https://github.com/orgs/jellyfin/projects/49?pane=issue&itemId=49620582)) so you can have them already (the most relevant change is that, before, it rendered whatever was passed to it under a ``span``, now it renders the content in-place.

frontend/src/utils/subtitles.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+/**
+* Parses an ASS/SSA (SubStation Alpha) file from a given URL.
+* Extracts dialogue lines with start and end times, and text content.
+* 
+* Converts specific tags to styled <span> tags, 
+* and sanitizes the content by removing harmful tags.
+*/
+export async function parseSsaFile(src: string) {
```
ferferga, line 113: Same as above. This is so similar to above function, really you don't think the common parts can be refactored out and shared between both? See for instance the ``api.ts`` file from composables, where there's a lot of extracted parts out there for some insights!

frontend/src/store/player-element.ts
```diff
@@ -8,18 +8,26 @@ import JASSUB from 'jassub';
...
 import { computed, nextTick, shallowRef, watch } from 'vue';
-import { playbackManager } from './playback-manager';
+import { playbackManager, type PlaybackExternalTrack } from './playback-manager';
 import { isArray, isNil, sealed } from '@/utils/validation';
 import { mediaElementRef } from '@/store';
 import { CommonStore } from '@/store/super/common-store';
 import { router } from '@/plugins/router';
 import { remote } from '@/plugins/remote';
+import { SubtitleDeliveryMethod } from '@jellyfin/sdk/lib/generated-client/models/subtitle-delivery-method';
+import { isMobile } from '@/utils/browser-detection';
```
ferferga, line 18: I also plan to remove this in the migration to MediaCapabilities. Why is needed?
I've explained this in some other issues/PRs but can't find them right now, but basically the reasoning behind it it's that user agents are in its way to deprecation and the barriers of what a mobile device is really unambiguos nowadays? Is a VisionPro a mobile device? A Surface? An Android Auto/CarPlay unit?

We strictly need to check the relevant thing we need (for instance touch support) or size/scaling, nothing else. I aim for this to be as generic (but as precise) as possible in this regard.

frontend/src/store/client-settings.ts
```diff
@@ -52,6 +61,39 @@ class ClientSettingsStore extends SyncedStore<ClientSettingsState> {
...
+      marginBottom: `${this.subtitleAppearance.positionFromBottom}vh`,
+      backgroundColor: this.subtitleAppearance.backdrop ? 'rgba(0, 0, 0, 0.5)' : 'transparent',
+      padding: '10px',
+      color: 'white',
+      /**
+       * Unwrap stroke style if stroke is enabled
+       */
+      ...(this.subtitleAppearance.stroke && strokeStyle)
+    }
+  }
```
ferferga, line 95: Why not do something like this? https://vuejs.org/api/sfc-css-features.html#v-bind-in-css

ferferga (commented):
Thank you very much for your conntribution ❤️❤️❤️!

Beforehand, please bear with me because, as I point out in one of my comments, I'm not fully knowledgeable of all the subtitle flow we need to handle in Jellyfin and I might be naive in some aspects.

Regardless of the refactoring bits, I'd like to have some input from @erikbucik as well. Although current web client provides with font customization options, I don't think this is a good idea. While in the surface it might seem better to provide the same options as the current web client and a huge amount of customization, I don't think being feature complete === repeat the same mistakes/UX quirks of the current web client. This is my reasoning:

* Users expect consistency across an application. This means that, by default, subtitles should use the same typography that we use everywhere, right? Are we on the same page on this?
* Introducing these kind of customization options might appeal at first, because we're catering for more user tastes. But, regardless what we do, in this case it will be impossible to cater to all tastes and, at the end of the day, we're offering a closed typography selection. If I'd want another typography, I'd need to rebuild the client with the typography of my liking, figure out all this logic, add it to all this logic and profit! At this point, we're making more difficult changing it, which is something that I could also do right now (add font, add CSS, build and profit!). A counter-example of this would be the [playback speed selection](https://github.com/jellyfin/jellyfin-vue/pull/2266)): it's integrated in a way that it's flexible enough to cater all the endless likings of an user without compromises, it's not complex and there are good defaults for those who don't want the extra customization (the selection menu on click).

Hopefully I explained myself right, since I'm in the middle of some exams I'm reviewing this in the quick way, sorry!

Thank you very much again for your contribution!

frontend/src/components/Playback/SubtitleTrack.vue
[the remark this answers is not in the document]
seanmcbroom: The custom subtitle component is nessesary because styles on the default track element are very limited, this is allows for more customization. For example the stroke outline wouldn't be possible without a custom component since those styles are not supported on the default track element.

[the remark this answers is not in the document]
seanmcbroom: This should also make it possible for extensions to interact with the subtitle track (although I haven't tested it yet). For example it could work with a furigana extension to add furigana characters over kanji while watching with Japanese subtitles.

frontend/src/pages/settings/subtitles.vue
```diff
@@ -0,0 +1,119 @@
...
+const fontSize = ref(clientSettings.subtitleAppearance.fontSize);
+const positionFromBottom = ref(clientSettings.subtitleAppearance.positionFromBottom);
+
+// Functions to update settings on slider release
+const updateFontSize = () => {
+  clientSettings.subtitleAppearance.fontSize = fontSize.value;
+};
+const updatePositionFromBottom = () => {
+  clientSettings.subtitleAppearance.positionFromBottom = positionFromBottom.value;
+};
```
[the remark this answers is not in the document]
seanmcbroom: If linked directly to the store, they update the settings every tick. Which causes a ton of unnecessary updates to the client store.

frontend/src/store/player-element.ts
seanmcbroom: Typo meant to type false.

frontend/src/components/Playback/SubtitleTrack.vue
```diff
@@ -0,0 +1,88 @@
...
+};
+/**
+ * Function to update the current subtitle based on the current time of the media element.
+ */
+const updateSubtitle = () => {
+  const currentSubtitleTrackData = playerElement.currentExternalSubtitleTrack?.parsed;
+  const currentTime = mediaElementRef.value?.currentTime;
+
+  currentSubtitle.value = findCurrentSubtitle(currentSubtitleTrackData, currentTime);
+};
```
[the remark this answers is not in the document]
seanmcbroom: You're right this should be I'm quite new to vue so I made a few weird choices, anyways I switched it to a computed value.

frontend/src/store/player-element.ts
seanmcbroom: It's used to check how to apply subtitles, since displaying the custom element isn't always possible on some devices i.e. inside the iOS fullscreen media element (which is it's own custom element). So in cases like that you'd still apply a track to the video element.

However you're correct this is a very generic fix that won't be perfect in all scenarios, so I'll start thinking of another way to implement this.

frontend/locales/en.json
```diff
@@ -352,7 +354,10 @@
   "startNow": "Start now",
   "status": "Status",
   "stretch": "Stretch",
+  "stroke": "Stroke",
   "studios": "Studios",
+  "subtitleFont": "Subtitle Font",
+  "subtitlePreviewText": "This is a preview of subtitles \n on this device.",
```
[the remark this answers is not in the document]
seanmcbroom: I was debating adding customizable line height but it seemed pointless, it can be removed.

frontend/src/utils/subtitles.ts
seanmcbroom: I refactored the SSA/ASS parsing. I was more focused on writing out VTT first since it's the more popular format and much simpler, so most of the code was reused and not complete. Now it should take the specified format into account and add multi-line dialogue together.

seanmcbroom:
First of all, thank you for taking the time to review the pull request, I know you've been busy and I wish you good luck on your exams!

I've already responded to a few of the discussions. I couldn't get to all of them today but I'll try to get to them done within the next few days. 👍

frontend/src/components/Playback/SubtitleTrack.vue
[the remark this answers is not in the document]
seanmcbroom: Oh and to answer your question of the different subtitle formats, currently it only handles VTT/SubRip subtitles (.srt) and ASS/SSA (.ass) subtitles. Any other external format isn't yet supported (same as before).

frontend/src/store/client-settings.ts
seanmcbroom: This was done for reuse-ability since these same CSS properties are used in the settings page and in the track component.

ferferga: @seanmcbroom Can't we use the SubtitleTrack component to display in settings as well? So we keep CSS with CSS. That way we don't also need the webkit prefixes, since those are handled by lightningcss at bundle time.

A boolean prop ``settings`` or passing a custom slot could do the trick maybe?

frontend/src/components/Playback/SubtitleTrack.vue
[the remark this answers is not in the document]
ferferga: Hopefully you're enjoying it! :)

ferferga (commented):
A few extra comments, now I think I fully understand the scope of this much better, thank you!

Answered the non resolved conversations of my previous comments and I'm also adding some small things that I noticed in this second pass.

Re-tagging @erikbucik for his input regarding my first review comment and whether we should give typography choices or not.

Also, FYI, I might merge these days a few PRs of mine: they're part of some work I've been doing for my uni thesis (the reason why I'm so busy lately, since it needs to be delivered next week :P) and I had them locally for a while, and for my defense it's good to have them upstreamed already. Don't take it as I'm neglecting this PR please, definitely good stuff here!

Also, if you can, merge the latest changes locally and run ``npm ci`` so you have the linter working again.

frontend/src/store/player-element.ts
ferferga: I upstreamed to VueUse all the logic that makes fullscreen work in Safari, take a look: https://github.com/vueuse/vueuse/pull/2822

If I recall correctly, the limitations about custom elements inside video only applied when playing fullscreen. Maybe you could extract this check: https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/composables/use-playback.ts#L21

If fullscreeen works without element, then you can assume it's not iOS, if it doesn't, then it's iOS (add proper comment block for that of course for future reference).

erikbucik:
I think it's good to let the user choose a different font. If not purely because of personal preferences, then for better readability.
Our current UI font (Figtree) may not be the easiest to read quickly (smaller x-height, wider and geometric-like letterforms, etc.), therefore allowing a font change for subtitles is a good idea.
The subtitle font servers one purpose (aside from fully supporting the language the subs are in) => helping the watcher read & understand the subs really quickly so that they can look at the video for as long as possible.

The other customization options (position, stroke, shadow, etc.) are also nice to have.

ferferga:
@seanmcbroom After further chat with Erik, we came up with the following conclusions as well (though these are more technical related):

* The displayed fonts should be the system's ones. You can get the current installed fonts using this API: https://developer.mozilla.org/en-US/docs/Web/API/Window/queryLocalFonts
* By using the system fonts, we're sure the text can be displayed and we don't need to increase the bundle with a lot of fonts.
* That API is currently just supported by Chromium browsers. It entered [draft](https://wicg.github.io/local-font-access/) at W3C just 2 weeks ago (7 Jun 2024), so I expect it will take a while to reach the other browsers still, but it should come at some point.
* Given this is a niche feature and Chromium is the most [widely used browser](https://gs.statcounter.com/browser-market-share/), it's not critical to think of a workaround and the selector could be disabled.
* That API requires permission. Use https://vueuse.org/core/usePermission/#usepermission to reactively track if permission is given and disable/enable the selector with that.
* Use https://vueuse.org/core/useSupported/#usage to reactively check if the API is supported
* Given the API it's not widely supported, we need to display informational messages:

**Custom fonts are currently not supported.**

And the following buttons/links:

"Learn more": When not supported, linking to the CanIUse page
"Allow permission": When the API is supported but permission is denied

* The font selector must be in its own component, so it could be reused for changing Books typography or even the whole UI.
* The font selector should display each font in the specific font:

image

* I'd say that the VSelect structure should be something like this:

```
<VListItem>*Current font* (Default)</VListItem>
<VDivider />
...Rest of fonts (also wrapped in VListItem)
```

* *Current Font* must be replaced by the current's font name of course. It would be great though if, instead of hardcoding it, we could query the fon'ts name from CSS and computedStyle or using Font API. That way, if we change the font later, somehow CSS stylesheet is overriden by the user (because he modified the html file, for example) or allow for font customization in appearance settings for the whole client, the real selected font is displayed.

What do you think of these points?

frontend/src/utils/subtitles.ts
```diff
@@ -0,0 +1,214 @@
...
+    return subtitles;
+  } catch (err) {
+    console.error("Error parsing VTT subtitles", err);
+  }
+}
+
+/**
+ * Parses dialogue line from SSA file.
+ */
+const parseSsaDialogue = (line: string, formatFields: string[]) => {
```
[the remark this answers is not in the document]
seanmcbroom: Yes, SSA files are standardized and should always follow the same format. I don't know what you would define as "simple" subtitles (probably subtitles without styles), but there appears to be a style definition under the V4 Styles headers that is named "Default" when it's default styles. (To be honest I don't have many SSA tracks to check from either so this is something I'll look into)

https://wiki.videolan.org/SubStation_Alpha

seanmcbroom:
I agree with all your points and proposed changes to the font selection. I wasn't even aware of the queryLocalFonts method and the timing for it being passed into chromium is almost perfect, so I guess it was meant to be haha.

frontend/src/components/Playback/SubtitleTrack.vue
```diff
@@ -0,0 +1,110 @@
...
+<script setup>
+import { computed, defineProps } from 'vue';
+import { clientSettings } from '@/store/client-settings';
+import { mediaControls } from '@/store';
+import { playerElement } from '@/store/player-element';
+import { isNil } from '@/utils/validation';
+
+defineProps({
+  previewText: String
+});
```
ferferga, line 27: See how we define the props in other places. This is using Vue built-in types, but if we use TypeScript interfaces, we could make it optional much easily:

```suggestion
defineProps<{
  previewText?: string;
}>();
```

Which is the equivalent with Vue types to something like this (might not work since I'm typing this by heart and it's been a while since I don't use this types):

```suggestion
defineProps({
  previewText: {
     type: String,
     default: undefined
});
```

Yeah, the Vue types should be deprecated altogether in a future Vue 4 because it's all super confusing :P

frontend/src/components/Playback/SubtitleTrack.vue
```diff
@@ -0,0 +1,110 @@
...
+</script>
+
+<style scoped>
+.subtitle-track {
+	position: absolute;
+	text-align: center;
+	width: 100%;
+  bottom: 0;
+  left: 0;
+}
```
ferferga, line 104: Can you also use our unocss classes? By typing ``uno-`` in the class attribute (and with the extension installed), it will autocomplete for you. This way we reduce bundle size by sharing global classes.

```suggestion
.subtitle-track {
  position: absolute;
  text-align: center;
  width: 100%;
  bottom: 0;
  left: 0;
}
```

You have interactive docs [here](https://unocss.dev/interactive/)

frontend/src/utils/subtitles.ts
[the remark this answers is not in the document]
seanmcbroom: It's already run asynchronously, so it shouldn't be blocking the main loop anyway. What benefits does WebWorker provide besides being run asynchronously?

frontend/src/utils/subtitles.ts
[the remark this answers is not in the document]
ferferga: @seanmcbroom Yes, sorry for not being more clear. With simple I meant subtitles without fancy styles.

frontend/src/pages/settings/subtitles.vue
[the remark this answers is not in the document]
ferferga: @seanmcbroom Don't forget this!

frontend/src/utils/subtitles.ts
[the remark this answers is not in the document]
ferferga: @seanmcbroom Being async doesn't mean it doesn't block the main loop. async/await is just syntatic sugar over Promises. Promises themselves don't block the main thread while they're being resolved, the event loop can defer them in the background... But if you fire wrap in a promise a while loop that resolves at the 1 millionth iteration, you'll see how, as soon as the event loop starts processing the promise, the whole thread will be blocked because the loop itself is blocking.

asynchronous code still runs in the main thread, although in a deferred way.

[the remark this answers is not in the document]
seanmcbroom: I see, sorry about my lack of knowledge regarding how javascript is run on a hardware level. I'll look into it and update you.

ferferga:
The history of this PR got mangled somehow, I'll try tomorrow to clean it up, was almost halfway through it in my last review.
Updates to the branch must be made using rebases, not merge commits (but this is a FYI for now, don't do anything to fix this, I'll take care of it).

ferferga:
@seanmcbroom I've rebased your branch and properly fixed all the conflicts.

The way to update your branch locally is to add the main repo as a remote:

``git remote add upstream https://github.com/jellyfin/jellyfin-vue``

Fetch changes:

``git fetch --all``

And rebase (the ``-i`` is optional):

``git rebase -i upstream/master``

With the -i, you can squash, reword and edit commits (that's what I did to fix up the names that were wrong and didn't pass commit linting). GitLens extension (recommended by the workspace) can really help as well in those tasks by showing better how the git-rebase-todo file will perform the operations. There are plenty of tutorials online about the different rebase operations if you want to know more, but you can also ask here for an specific one.

With rebases, conflicts are fixed in a commit by commit basis, that's why there are no merge commits involved and the history of changes is fully linear.

To fix the divergence between your local branch and the remote one, the easiest thing is that, located in the master branch (the one of this PR) you switch to a new branch:

``git checkout -b blah``

Delete the old master:

``git branch -D master``

Fetch changes (this is done automatically by git and VSCode, but just in case):

`git fetch --all`

And change to the branch and delete the temporal one:

`git checkout master && git branch -D blah`

Hopefully these instructions are enough to get you back on track again.

frontend/src/components/Selectors/FontSelector.vue
```diff
@@ -0,0 +1,74 @@
...
+      class="uno-font-bold"
+      href="https://caniuse.com/mdn-api_window_querylocalfonts"
+      target="_blank"
+      rel="noopener">
+      {{ $t('learnMore') }}
+    </a>
+  </VAlert>
+
+  <VAlert
+    v-if="isQueryLocalFontsSupported && !fontAccess"
```
ferferga, line 20:
```suggestion
    v-else-if="!fontAccess"
```

frontend/src/components/Playback/SubtitleTrack.vue
```diff
@@ -0,0 +1,95 @@
...
+  for (const [i, subtitle] of data.dialogue.entries()) {
+    if (subtitle.start < currentTime && subtitle.end > currentTime) {
+      lastIndex = i; // Update the last found index
+
+      return subtitle;
+    } else if (subtitle.start > currentTime) {
+      break;
+    }
+  }
+};
```
ferferga, line 60: I believe a more performant approach would be to have a Map (computed on the ``parseSsaFile`` of course):

* key is the timestamp rounded to only one decimal (I believe this should cover all cases keeping a good precision?)
* Value is the content of the subtitle

The computed property just needs to also do a ``Math.round`` to the current time and return the value of the map, if it's undefined, the [previous one](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/components/lib/JVirtual/JVirtual.vue#L200) is returned.

frontend/src/pages/settings/subtitles.vue
```diff
@@ -0,0 +1,52 @@
...
+
+<script setup lang="ts">
+import { useI18n } from 'vue-i18n';
+import { useRoute } from 'vue-router/auto';
+import { clientSettings } from '@/store/client-settings';
+
+const { t } = useI18n();
+const route = useRoute();
+
+route.meta.title = t('subtitles');
```
ferferga, line 51: Update to use the new ``usePageTitle`` from ``@/composables/``. There are many examples across the codebase in case you need them.

frontend/src/store/client-settings.ts
ferferga, line 1: *Optional, I can do this myself later, I don't want to further annoy you with a thing that I should have noted since the beginning*

Can you extract all the subtitle-related functionality to a new store? Create the folder ``@/store/client-settings`` and move this store to it and rename it to ``index.ts`` (so we keep the current import paths and keep LoC to a minimum, at least for this PR). The new store should be named ``subtitles.ts`` inside the new client-settings folder.

ferferga (changes requested):
Last blockers (except my last comment), I promise!

Additionally, please take a look at SonarCloud's reported issues and fix them accordingly. They're pointed out in the automatic comment their bot creates or you can [jump directly from here](https://sonarcloud.io/project/issues?id=jellyfin_jellyfin-vue&pullRequest=2360&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true).

I believe local linter should've pointed them out as well, but it seems not, I'll check why.

frontend/src/components/Playback/SubtitleTrack.vue
seanmcbroom: I don't quite understand what you mean by previous one. If you're referring to the previous subtitle, I believe it would be disregarding the expected behavior of subtitles, since there can be intentional gaps between dialogue (Usually when there's no speech). Other than that I agree with refactoring this code to use a map since it would probably be more performant and easier to read.

ferferga (dismissed):
Hello again @seanmcbroom 
The wait is finally over and this looks good to merge for me! I have a few minor questions though, but besides that everything's great and that's why I approve.

I pushed some changes, I added descriptions to each commit if you want to look further, but basically I finished the FontSelector implementation so it's ready to be implemented in the future in an "Appearance" settings page.

Here's a picture of the options when the app's typography is not the default one:
image

Here's a picture when both app typography and selected typography are the same:
image

I also optimized the algorithm for finding the current subtitle and offloaded the parsing to a WebWorker.

These changes deprecate the `useFont` composable, since now we can just rely on the CSS variable or the value of the clientSettings store instead (sorry for not seeing it better in hindsight before actually diving into the code! 🙄

Please test everything and let me know how it goes.

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -15,8 +15,11 @@
         :loop="playbackManager.isRepeatingOnce"
         :class="{ 'uno-object-fill': playerElement.isStretched.value }"
         @loadeddata="onLoadedData">
+        <SubtitleTrack
+          v-if="subtitleSettings.state.enabled && playerElement.currentExternalSubtitleTrack?.parsed !== undefined" />
```
ferferga, line 19: Previously we were showing the browser's tracks alongisde our custom ones. I guess this is not the wanted behaviour, right? I also added a new toggle in the settings to toggle the subtitle customization (that's what configures the `subtitleSettings.state.enabled` variable

frontend/src/store/player-element.ts
```diff
@@ -132,51 +189,107 @@ class PlayerElementStore extends CommonStore<PlayerElementState> {
...
       return;
     }
 
-    await nextTick();
-
     /**
-     * Disabling VTT and SSA subs at first
+     * Clear VTT and SSA subs first
      */
     for (const textTrack of mediaElementRef.value.textTracks) {
```
ferferga, line 294: Seems this isn't changed in this PR, but typescript reports that textTracks is not iterable (though it might be a false positive because the environment is still not well configured). Can you verify it please?

frontend/src/components/Playback/PlayerElement.vue
ferferga, line 19: Previously we were showing the browser's tracks alongisde our custom ones. I guess this is not the wanted behaviour, right? I also added a new toggle in the settings to toggle the subtitle customization (that's what configures the `subtitleSettings.state.enabled` variable

ferferga (approved):
Hello again @seanmcbroom 
The wait is finally over and this looks good to merge for me! I have a few minor questions though, but besides that everything's great and that's why I approve.

I pushed some changes, I added descriptions to each commit if you want to look further, but basically I finished the FontSelector implementation so it's ready to be implemented in the future in an "Appearance" settings page.

Here's a picture of the options when the app's typography is not the default one:
image

Here's a picture when both app typography and selected typography are the same:
image

I also optimized the algorithm for finding the current subtitle and offloaded the parsing to a WebWorker.

These changes deprecate the `useFont` composable, since now we can just rely on the CSS variable or the value of the clientSettings store instead (sorry for not seeing it better in hindsight before actually diving into the code! 🙄

Please test everything and let me know how it goes.

ferferga:
Also, I'm wondering if we should sync these settings? The typography available in one device might not be available in others...

seanmcbroom:
> Also, I'm wondering if we should sync these settings? The typography available in one device might not be available in others...
> 
> Don't worry about the merge conflict btw

Yeah these settings should probably be client-independent, since there are different default font faces on most operating systems.

ferferga:
@seanmcbroom I forgot that back in the day I added support in the synced store for syncing specific keys only. I synced all of them but the font family, does that sound good to you? As soon as you let me know about this, confirm that everything is alright with the new subtitle finding algorithm and [answer this question](https://github.com/jellyfin/jellyfin-vue/pull/2360#discussion_r1749213226) we're shipping this! 🚀

seanmcbroom:
Sorry, I wrote the review changes yesterday but didn't realize they didn't submit until now. Anyways, those should be the final changes. Once those minor bugs are worked out I think everything should be ready!

frontend/src/components/Playback/SubtitleTrack.vue
```diff
@@ -0,0 +1,100 @@
...
+    const isNext = hasPrevious && predicate(dialogue.value[previous.index + 1]);
+    const isCurrent = hasPrevious && predicate(dialogue.value[previous.index]);
+
+    if (isCurrent) {
+      return previous;
+    } else {
+      const newIndex = isNext ? previous.index + 1 : findSubtitle(dialogue.value);
+
+      return { index: newIndex, sub: dialogue.value[newIndex] };
+    }
```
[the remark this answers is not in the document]
ferferga: @seanmcbroom This is the day of things that we think we did but we didn't 🤦‍♂️.

 If you check `findSubtitle`, it returns `undefined` instead of -1 when no element is found, so your condition as is will never be truthy (instead we should check for nil, good catch!).

The question I have now is whether what do you think is more correct: return undefined or return the previous good known value?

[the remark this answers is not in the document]
seanmcbroom: If no subtitle is found it should return undefined, since there can be intentional gaps without subtitles.

frontend/src/components/Playback/SubtitleTrack.vue
```diff
@@ -0,0 +1,100 @@
...
+      fontFamily: `${fontFamily.value} !important`
+    })
+  };
+});
+</script>
+
+<style scoped>
+.stroked {
+  --webkit-text-stroke-color: black;
+  --webkit-text-stroke-width: 7px;
```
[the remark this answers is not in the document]
ferferga: Good catch mate!

frontend/src/components/Playback/SubtitleTrack.vue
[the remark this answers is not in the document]
ferferga: I also found out thanks to your comment that I forgot to pass the last found index to `findSubtitle`!

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -15,8 +15,11 @@
...
         <track
           v-for="sub in playbackManager.currentItemVttParsedSubtitleTracks"
+          v-else
           :key="`${playbackManager.currentSourceUrl}-${sub.srcIndex}`"
```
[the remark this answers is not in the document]
ferferga: @seanmcbroom This is our only blocker now then 🙄. I wonder if the issue is caused by some issue in the transferable (main <-> WebWorker "serialization" of objects). I added a `console.log` at the end of the `parseSsa` while playing the only movie I know of from my library that has SSA (it's late here so I will need to find more media tomorrow).

Can you help me debug if you're getting results correctly in the main thread? Perhaps something else was borked? The logic here is super straightforward, it's really strange that such a simple `if` broke this.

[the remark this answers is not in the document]
seanmcbroom: Seems like its a problem between the weird behavior of the media element and the TextTrack api. I'd say to just load all text tracks onto the media element and set the mode to "disabled". This will stop the client from loading the cues and processing events (https://developer.mozilla.org/en-US/docs/Web/API/TextTrack/mode#disabled). 

As for the custom track element, I believe it just cannot be displayed inside of the video element sadly.

I'm going to push a commit that will hopefully address these issues and you can review it.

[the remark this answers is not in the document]
seanmcbroom: I just realized the track element is set to disabled by default, so I've removed that portion of the code

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -15,17 +15,17 @@
         :loop="playbackManager.isRepeatingOnce"
         :class="{ 'uno-object-fill': playerElement.isStretched.value }"
         @loadeddata="onLoadedData">
-        <SubtitleTrack
-          v-if="subtitleSettings.state.enabled && playerElement.currentExternalSubtitleTrack?.parsed !== undefined" />
         <track
           v-for="sub in playbackManager.currentItemVttParsedSubtitleTracks"
-          v-else
```
ferferga: This is still unclear to me, both must be rendered at the same time? To me, it makes sense for it to be one of the other, not both at the same time. Perhaps what was wrong before was the order of the conditional (so it should be track first with `!subtitleSettings.state.enabled` and SubtitleTrack afterwards with `v-else-if`?

seanmcbroom: Because of the weird way the track element works for subtitles, loading tracks onto the media element doesn't render them, they're disabled by default actually. So they're not visible, don't load in any data, or handle any events. Until they're set to "showing" in the code.

We have to load all the tracks to be able to "show" them in the `player-element`, when available.

I don't think there's a way to use the if-else statements here because of the way tracks work.

ferferga: Alright!

ferferga:
@seanmcbroom Okey, let's ship this!
Thank you very much for all your knowledge about subtitles and patience while getting this in! ❤️❤️🚀🚀🚀

PR #2383, opened 2024-06-22
fix(PlayPauseButton): conflict with spacebar in fullscreen video

ferferga:


frontend/src/components/Buttons/Playback/PlayPauseButton.vue
```diff
@@ -3,14 +3,21 @@
     icon
     :size="size"
     :loading="playbackManager.isBuffering"
-    @click="playbackManager.playPause">
+    @click="(e: MouseEvent) => {
+      if ('pointerType' in e && e.pointerType) {
```
manicExpressive, line 7: `e.keyCode != 32` might be a better fix - this solution breaks touch input on the button (I wouldn’t have noticed but happened to be testing via an iPad!).

More on this situation which I am only just learning about: http://www.quirksmode.org/dom/events/click.html

manicExpressive (changes requested):

frontend/src/components/Buttons/Playback/PlayPauseButton.vue
ferferga: I also ran the tests in Microsoft Edge on a Surface, where that property's value was "touch". So I assume in Chromium that property changes between "mouse", "touch" and "" when it's not pointer based, but in Safari that property doesn't exist whatsoever? Funky differences between browsers strikes again

I don't recall seeing keyCode in the payload of that event (in Chromium) either, but I need to check again. If it doesn't exist, I think that the only cross browser solution would be to focus the button as I mentioned in the issue.

But we can't do that now because the underlying ref is not exposed by Vuetify (and using querySelector is not a good solution). Would you like contributing in creating a JBtn component that has the same implementation as Vuetify but in SFC?

ferferga: @manicExpressive I can confirm you now that keyCode doesn't exist in Chromium-based browsers.

manicExpressive: Got it, I think I'm tracking with you now that I've poked around on my (Ubuntu / Chrome / Firefox) desktop a bit.

I don't have capacity in the near term to contribute much, but I'll circle back and check-in when my schedule opens up a bit.

PR #2399, opened 2024-07-07
fix(routing): fix history routerMode

Beat-YT:
### Fixes

- **serverAdd**: Block server addition and selection when disallowed by config.
- **router**: Correct history routerMode to ensure proper asset pathing.

### Description

1. Updated the server addition and selection logic to respect the `allowServerSelection` configuration setting.
2. Fixed an issue with the history routerMode where the base path was incorrectly set to `./`, causing static assets to be pathed incorrectly (e.g., `/server/assets/example.png` instead of `/assets/example.png` when on `/server/login`).

### Testing

- Verified that server addition and selection are blocked when `allowServerSelection` is false.
- Tested history routerMode functionality to ensure static assets are correctly pathed.

### DIFF
--- a/frontend/public/config.json
+++ b/frontend/public/config.json
@@ -1,5 +1,5 @@
 {
   "defaultServerURLs": [],
   "allowServerSelection": true,
-  "routerMode": "hash"
+  "routerMode": "history"
 }
--- a/frontend/src/plugins/router/middlewares/login.ts
+++ b/frontend/src/plugins/router/middlewares/login.ts
@@ -24,10 +24,6 @@ export async function loginGuard(
 
   if (!isNil(remote.auth.currentServer) && !isNil(remote.auth.currentUser) && !isNil(remote.auth.currentUserToken) && routes.has(to.path)) {
     destinationRoute = { path: '/', replace: true };
-  } else if (to.path === serverAddUrl && remote.auth.servers.length > 0 || to.path === serverSelectUrl) {
-    if (!jsonConfig.allowServerSelection) {
-      destinationRoute = { path: serverLoginUrl, replace: true };
-    }
   }
 
   if (remote.auth.servers.length <= 0 && jsonConfig.defaultServerURLs.length <= 0) {
--- a/frontend/src/plugins/router/middlewares/login.ts
+++ b/frontend/src/plugins/router/middlewares/login.ts
@@ -24,6 +24,10 @@ export async function loginGuard(
 
   if (!isNil(remote.auth.currentServer) && !isNil(remote.auth.currentUser) && !isNil(remote.auth.currentUserToken) && routes.has(to.path)) {
     destinationRoute = { path: '/', replace: true };
+  } else if (to.path === serverAddUrl && remote.auth.servers.length > 0 || to.path === serverSelectUrl) {
+    if (!jsonConfig.allowServerSelection) {
+      destinationRoute = { path: serverLoginUrl, replace: true };
+    }
   }
 
   if (remote.auth.servers.length <= 0 && jsonConfig.defaultServerURLs.length <= 0) {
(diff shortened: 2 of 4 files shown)


ferferga:
@Beat-YT Thank you very much for your PR! Have you checked if the revert to absolute paths doesn't cause a regression in the situations described in https://github.com/jellyfin/jellyfin-vue/pull/2108 and https://github.com/jellyfin/jellyfin-vue/issues/2085?

Beat-YT:
@ferferga  Thank you for the reply
I understand,
if someone wants to use in a sub-path, we need relative url.
but to get a domain deployment we need to use absolute.

I think we can solve this by adding baseUrl to the config file.

Beat-YT:
@ferferga I believe this overcomes the situations described in https://github.com/jellyfin/jellyfin-vue/pull/2108 and https://github.com/jellyfin/jellyfin-vue/issues/2085?

Beat-YT:
@ferferga an alternative solution would be to race promise between the absolute and relative config file.

ferferga:
@Beat-YT I've tested this and this makes a regression to both issues. This is a problem that we've been dealing with for some time and I researched it extensively: the issue when using HTML5 history mode is that the browser can't distinguish between what's a route inside the application and what's the server path.
For example if I'm in ``/settings``, what does settings mean, it's inside the app or the webserver? That's the distinction that the shebang hints at the web browser. I thought at first that ``import.meta.env`` was something I missed back in the day where the browser provides info about the environment, but that just prints whatever ``base`` property it's set at build time, so this is still impossible to achieve at runtime.

We need the frontend to be as compatible as possible out of the box and currently it is, so it can be hosted out of the box when:

* Served from the root path
* Served from subpath

Sadly that means we need the hash history but, believe me, I'd also like to get rid of it. It would be awesome for everybody if we could get rid of any hardcoded ``base`` option and have it calculated at runtime. But currently it's impossible.
For someone who want clean urls, the solution is:

* Handle it themselves at the webserver level
* Build the client with the desired base.

You can quickly reproduce all of this by spinning a Codespace in your branch, replace ``location / {`` from [here](https://github.com/jellyfin/jellyfin-vue/blob/master/packaging/docker/contents/nginx.conf#L4) to ``location /subpath1/subpath2 {`` and [this one](https://github.com/jellyfin/jellyfin-vue/blob/master/packaging/docker/Dockerfile#L24) to ``COPY --from=build /app/frontend/dist/ /usr/share/nginx/html/subpath1/subpath2`` and [build](https://github.com/jellyfin/jellyfin-vue/blob/master/packaging/docker/README.md) the Docker image.

As for the other commit, it seems fine to me, so that specific one can be cherrypicked. I'll wait for your feedback first, in case I misunderstood something of your implementation and then we can follow with that commit.

Beat-YT:
@ferferga Thanks for your reply!

Perhaps using a plugin such as [vite-plugin-dynamic-base](https://github.com/chenxch/vite-plugin-dynamic-base/tree/main) can do the trick?

I can PoC the implementation if you are fine with adding said plugin to the project.

Otherwise I agree with you that it’s better to keep using the current hash router.

Beat-YT:
I removed the block server addition part to open a new pull request about it.

ferferga:
@Beat-YT The plugin won't fix anything either (at least unless I'm missing something you saw but I didn't).

What Vite does is hardcode whatever string is given to the ``base`` option so it's used everywhere. What the plugin does is to replace that hardocoded string to a JS variable, so you can change that variable at runtime using JS. But the plugin doesn't "magically" solves the problematic of identifying which part is which from the URL.

ferferga:
Hello @Beat-YT, I'm going to close this in the meantime for tidying up tha backlog a little.
If you ever are interested in tackling this again, feel free to open a new one.

PR #2404, opened 2024-07-11
Add support for native PGS subtitle rendering without transcoding

Arcus92:
**Changes**
This PR allows the client to request and render native PGS subtitle streams without transcoding. PGS is a graphical subtitle format. The subtitles are rendered locally by the client on-top of the video element.

This adds a new dependency [libpgs-js](https://github.com/Arcus92/libpgs-js). I just created this library for the Jellyfin project. If desired, you can fork it into the https://github.com/jellyfin domain.

**Related Pull Request (backend):** https://github.com/jellyfin/jellyfin/pull/12056
**Related Pull Request (web):** https://github.com/jellyfin/jellyfin-web/pull/5688

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -36,6 +36,7 @@
     "fast-equals": "5.0.1",
     "hls.js": "1.5.14",
     "jassub": "1.7.16",
+    "libpgs": "0.4.1",
     "lodash-es": "4.17.21",
     "marked": "14.0.0",
     "sortablejs": "1.15.2",
--- a/frontend/src/store/playback-manager.ts
+++ b/frontend/src/store/playback-manager.ts
@@ -291,7 +291,7 @@ class PlaybackManagerStore extends CommonStore<PlaybackManagerState> {
           srcLang: sub.Language ?? undefined,
           type: sub.DeliveryMethod ?? SubtitleDeliveryMethod.Drop,
           srcIndex: sub.srcIndex,
-          codec: sub.Codec === null ? undefined : sub.Codec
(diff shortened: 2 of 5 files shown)


ferferga (commented):
Code LGTM, not tested with server yet but I suppose you did. Any reason why this is still listed as draft? Can this go in full review mode?

Arcus92:
@ferferga Yes, it is working just fine. 
Like the default web interface I would like to add a opt-in option in the subtitle settings. However, I had no time to look into the setting menu in vue yet. Maybe this weekend.

Arcus92:
@ferferga Just noticed there aren't any subtitle settings at the moment. Is it okay to just leave it enabled for now?

Btw. here is it working in vue:
grafik

ferferga:
@Arcus92 The subtitle settings are being worked on in https://github.com/jellyfin/jellyfin-vue/pull/2360
After your perf improvements in your library, is there any good reason for keeping the opt-in setting for rendering? (I'd say it should be opt-out in Vue as well, given this client targets evergreen browsers and modern platforms, unlike web).

Arcus92:
It should be fine performance wise. I haven't tested different server architectures yet. Ripping .sup files from the source is an intensive task. Worst case ffmpeg could run out of memory or cache disk space and block playback or crash the server. Currently this is just my unfounded fear. For **web**, I just wanted to be extra safe because this a brand new feature.
Since **vue** is already experimental we could just enable it - with your permission - and wait for some feedback. Best case no one will notice the difference.

ferferga:
@Arcus92 It's completely alright for me. I'll fix the conflicts and ship it later today!

PR #2416, opened 2024-07-29
fix(serverAdd): block server addition and selection when disallowed by config

Beat-YT:
### Fix

Block server addition and selection when disallowed by config.

### Description

Updated the server addition and selection logic to respect the `allowServerSelection` configuration setting.

### Testing

Verified that server addition and selection are blocked when `allowServerSelection` is false.


ferferga:
FYI Im out on holiday so it will take a couple of weeks before this (and your other PR) are reviewed and merged!

Beat-YT:
Hey @ferferga what does `return false` do ?

PR #2419, opened 2024-07-30
Update README to reflect actual status

joshuaboniface:
I want to make it very clear in the README what the Vue client is and what it's goals are vis-a-vis the project as a whole, to avoid any mistaken impression about where it's at.

For technical reasons, mostly Vue.js dropping support for "old" browsers, Jellyfin Vue can never reasonably replace the main web client. So, it will always be at most an alternative client.

This PR ensures that this is now spelled out clearly at the top of the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@
 
 Jellyfin Vue is an experimental, alternative browser-based web client for Jellyfin written using Vue.js.
 
-Note: Jellyfin Vue is not currently planned or targeted to replace [the main Jellyfin Web client](https://github.com/jellyfin/jellyfin-web), and is not feature-complete.
+Note: Jellyfin Vue is not planned or targeted to replace [the main Jellyfin Web client](https://github.com/jellyfin/jellyfin-web), and is not feature-complete.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,7 +72,9 @@
 
 ---
 
-This is an alternative client for Jellyfin based on Vue.js. It might not be feature complete and it's constantly evolving.
+Jellyfin Vue is an experimental, alternative browser-based web client for Jellyfin written using Vue.js.
+
+Note: Jellyfin Vue is not planned or targeted to replace [the main Jellyfin Web client](https://github.com/jellyfin/jellyfin-web), and is not feature-complete.
```
ferferga, line 77:
```suggestion
> [!NOTE]
> Jellyfin Vue is not planned or targeted to replace [the main Jellyfin Web client](https://github.com/jellyfin/jellyfin-web), and is not feature-complete.
```

ferferga (commented):
What do you think of using the notification syntax so GitHub highlights it better?

Issue #2403, opened 2024-07-11
Back button doesn't always work


ferferga:
### Description of the bug

We replace the back implementation so it's impossible to go outside the scope of the application, but Vue Router's doesn't report properly the length of the history.

### Steps to reproduce

Go back more than once using the left-side button

### Expected behavior

You could traverse all your navigation history to go back to the first page that you visited

### Platform

Linux

### Browser

Edge

### Jellyfin server version

10.9.7

PR #2464, opened 2024-09-17
fix(router): don't use replace on every route when going back

ToastKiste21:
Related Issue: [#2403](https://github.com/jellyfin/jellyfin-vue/issues/2403)

**Summary**
This merge request resolves a navigation issue where users were unable to traverse the entire navigation history using the back button in the application's sidebar. The problem stemmed from overriding the router.back method with router.replace, which disrupted the history stack and prevented Vue Router from accurately tracking the history length.

### DIFF
--- a/frontend/src/plugins/router/index.ts
+++ b/frontend/src/plugins/router/index.ts
@@ -45,11 +45,10 @@ const backTransition = 'slide-x';
 router.back = () => {
   const route = router.currentRoute;
 
-    /**
+  /**
    * Play the default page transition but reversed, to play a different effect when going
    * to the previous page.
    */
-
   route.value.meta.layout.transition = {
     enter: 'slide-x-reverse',
     leave: route.value.meta.layout.transition.leave ?? backTransition
--- a/frontend/src/plugins/router/index.ts
+++ b/frontend/src/plugins/router/index.ts
@@ -50,7 +50,13 @@ router.back = () => {
     leave: route.value.meta.layout.transition.leave ?? backTransition
   };
 
-  window.history.back();
+  const historyState = router.options.history.state;
+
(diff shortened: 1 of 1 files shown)


ferferga (changes requested):
Thank you very much for your willingness to fix this! However, this doesn't fix the issue:

* This reintroduces the situation described in the comment: if you click on the link of the readme of this repo to load the hosted instance and then click in the back button, you will be redirected back to GitHub, while we want to hack all the back navigation inside the application (of course not in the browser's buttons, just in the app's).
* There's a reason why Vue Router has its own back function and we should not be more "clever". I believe this approach kills history routing and resolve hooks from Vue Router.

In either case, an appropiate fix would be to do nothing or perform the back on the logic truthiness, and only replace with index on falsyness. Can you try that and test if it fits all the conditions stated above?

ferferga (commented):
This works perfectly now! Can you please reintroduce the deleted comment about the transition though?

frontend/src/plugins/router/index.ts
```diff
@@ -42,23 +42,26 @@ router.beforeEach(metaGuard);
...
-router.back = (): ReturnType<typeof router.back> => {
+router.back = () => {
   const route = router.currentRoute;
 
-  /**
+    /**
    * Play the default page transition but reversed, to play a different effect when going
    * to the previous page.
    */
+
```
ferferga, line 52:
```suggestion
  /**
   * Play the default page transition but reversed, to play a different effect when going
   * to the previous page.
   */
```

ferferga (approved):

ferferga:
@ToastKiste21 Thank you very much for finding that replace was the issue!

PR #2480, opened 2024-10-25
chore: update to node 22 lts & add size reporting

ferferga:


ferferga:
Blocked until Node.js 22 is promoted to LTS (29/10/2024).
The size report requires Node.js 22 to work.

PR #2500, opened 2024-11-20
feat(subtitles): switch to assjs subtitle renderer

brys0:
Advanced SSA or ASS subs should now be rendered in DOM using assjs. This should greatly improve the performance of subtitle rendering.

Still have to "figure out" the "Precise Rendering" option to enable transcoding of subtitles on the server. Would appreciate help on that. Button and property "usePreciseSubtitles" already exists and is ready to be adapted to *actually* work.


brys0:
No problem. Sorry will get to it later today!

frontend/src/store/player-element.ts
```diff
@@ -118,63 +125,28 @@ class PlayerElementStore extends CommonStore<PlayerElementState> {
     }
   };
 
-  private readonly _setSsaTrack = (trackSrc: string, attachedFonts?: string[]): void => {
+  private _fetchSSATrack = async (trackSrc: string) => {
+    fetch(trackSrc).then((body) => body.text()).then((text) => {
+      this._asssub = new ASSSUB(text, mediaElementRef.value as (HTMLVideoElement), { resampling: 'video_width' });
+    });
+  };
```
ferferga, line 132: Use our current axios instance and ``await`` syntax please.
Also, mind adding a comment explaining why the default resampling (video_height) method from the library can't be used but video_width?

ferferga (commented):
I fixed the checks so I can test the deployment in the meantime. Would you mind fixing all the lint problems introduced by your changes as well as SonarCloud issues?

hawken93:
Probably something that was a problem before as well, but thought I would mention it:

The subtitle selector acts weird. Disabling doesn't disable, switching to a different subtitle track seems to work but subtitles are now stuck at the last one showing after subtitle track has been selected.. But it might be viewed as a different issue to solve.

Rendering seems fine and if left in default conditions I think it works :)

brys0:
> Probably something that was a problem before as well, but thought I would mention it:
> 
> The subtitle selector acts weird. Disabling doesn't disable, switching to a different subtitle track seems to work but subtitles are now stuck at the last one showing after subtitle track has been selected.. But it might be viewed as a different issue to solve.
> 
> Rendering seems fine and if left in default conditions I think it works :)

Hmmm... That's weird! In my testing disabling subs worked fine, so did switching.. everything should work okay as when switching the subtitle track we manually destroy the current assjs instance if it exists. I'll check later today and see if I can replicate.

frontend/src/store/player-element.ts
```diff
@@ -355,12 +314,21 @@ class PlayerElementStore extends CommonStore<PlayerElementState> {
...
+
+
+    });
+
     /**
      * We need to destroy JASSUB so the canvas can be recreated in the other view
      */
     watch(videoContainerRef, () => {
       if (!videoContainerRef.value) {
-        this._freeSsaTrack();
```
seanmcbroom, line 363: This function has to be refactored for ass.js. It prevents multiple canvases from rendering at the same time when switching through tracks.

frontend/src/store/player-element.ts
brys0: Understood, I can do that, my apologies, it was a quick fix! As per why I'm not using video_height instead on my silly screen which is an ultrawide monitor video_width actually works better most of the time, but I should've considered other screen sizes.

frontend/src/store/player-element.ts
brys0: Hey, believe this is working in latest commit.

frontend/src/store/player-element.ts
brys0: Alright, I believe this is working in the latest commit.

brys0:
I feel like the TODO is helpful in this context.. Should it still be removed?

ferferga:
@brys0 @seanmcbroom @hawken93 I've simplified the logic (still all the subtitle handling logic is really convoluted and needs further refactoring for sure) and I only could reproduce concurrency errors caused by the delay the worker introduces once, but all the race conditions with the fetch are fixed. After the removal of the worker processing, everything should be good to go in that regard as well.

@brys0 I rebased the branch to properly remove all the unrelated changes introduced in the commits and keep it up to date. We use rebases to keep branches up to date instead of merge commits.

Please give it everyone another go and let me know if this is ready to be merged.

ferferga (approved):

brys0:
Can a maintainer merge this PR? All conflicts have been resolved.

ThaUnknown:
hello, for some reason it was decided that this was a good idea? I made a nice comparison table as to why it's not: https://jassub.pages.dev

ferferga:
@ThaUnknown If I recall correctly, I think we discussed this in Matrix a little, and we did it because the library was simpler with no heavy reliance on WASM (less bundle size) and honestly assjs did a best effort. I had a lot of movies that didn't had subtitles at all with your lib because some characters [were missing from the attached font family from the media](https://github.com/ThaUnknown/jassub/issues/46) (and no solution was possible with the current architecture, at least that's what we understand from your last comment there), which assjs does a "best effort". Neither lib is perfect honestly (which is understandable since ASS itself it's extremely complex to get right), but assjs, at least for our use case, seemed better since it handled the edge cases a bit better, which at the end is better than nothing. Not all is performance, here we are trading it off with better compatibility (at least in our experience).

ThaUnknown:
my 2 cents:

> and no solution was possible with the current architecture

oh there are plenty of solutions, but none of them are the responsibility of the library itself, it's something the end developer needs to handle, for which the library provides plenty of options

you could compile your own font with those characters, you could detect track language or detect the track language by it's content by matching unicode ranges and change the default font to a public font that includes those characters etc

from my experience assJS is unusable outside of basic text, at which point you might as well just convert the text to VTT and render that instead of bothering with an ASS renderer, it doesn't implement most features, and the ones it does it doesn't parse correctly

> heavy reliance on WASM 

2MB is not really all that much, especially considering it's non-blocking and streamed-in only when required, this is like 2 images worth of data, which jf-vue displays plenty of so there's not much strain on the user, but yeah your library, you do what you want, I just made that original comment since it seemed like the decision was made without knowing how horrificly limited ASSjs is

brys0:
You bring up valid points. I created this PR mainly out of boredom and because I thought assjs was a really cool library, it obviously has some limitations compared to jassub, and perhaps it would be best to let the user pick and lazy load the resulting library. I meant no offense by the resulting PR of my curiosity.

PR #2503, opened 2024-11-21
feat(tauri): implement fullscreen toggle functionality

seanmcbroom:
Added fullscreen toggle feature using Tauri's window API. Permissions for fullscreen have also been updated in the tauri.conf.json file.


ferferga:
I'm labeling this as blocked because of: https://github.com/jellyfin/jellyfin-vue/discussions/2002

Now that I'm dividing the project more and more in a monorepo (see packages folder), my idea was at some point to have a ``tauri-runtime`` package that contains all tauri-related specific logic, so we don't "pollute" the main codebase with code that doesn't belong to the standard DOM environment (improving maintenance and reducing bundle size along the way).

For this, I think it would be a matter of creating that `tauri-runtime` package in ``packages`` folder and then make the ``main.ts`` and ``index.html`` of ``@jellyfin-vue/frontend`` exportable in their package.json, so you can import them in the entrypoint of the ``tauri-runtime`` package. Then, in that entrypoint, set an event handler that listens for window's fullscreen events and has Tauri's fullscreen method as its callback.

This is the overall high-level idea of how I think it would work, do you wanna try to tackle it?

seanmcbroom:
I understand. It sounds like a significant amount of work for the small change I’m proposing. Perhaps it makes sense to open a new PR to lay the groundwork for adding functionality like this in the future..

On another note, I noticed there hasn’t been an official Tauri build released in over a year. Do you think this is something that will see more support in the future? Personally, I find it loads much faster than the web version since most assets are requested locally. Plus, as this is an alternative client, I imagine most users would prefer running it locally rather than hosting it.

ferferga:
Yes, it should be a different PR of course. Are you interested to tackle it nevertheless?
Check this project for more details about the current plans for Tauri: https://github.com/orgs/jellyfin/projects/49

ferferga:
@seanmcbroom Check https://github.com/jellyfin/jellyfin-vue/pull/2539

It contains a new `@jellyfin-vue/tauri-runtime` package, which will be the one shipped by Tauri.

It works by injecting modules before the frontend's ones, so you can create your own polyfills that mocks the standard DOM API with the Tauri implementation.

PR #2504, opened 2024-11-21
fix: /server/login with disabled server selection

hawken93:
It's been a while since I contributed, so I'm kinda new again here and I don't know exactly what I'm doing.

However as far as my testing goes, this seemed to work. However I didn't verify if subtitles still work. But I imagine it should.

the div class=uno-my-auto div caused the media element to not realize it had more room to fill. Removing it let the parent elements css classes become effective. It still seems to work as expected in the mini version of the player.

This could also cause the stretch option to do something, however I'm not sure why anyone would even want this functionality (non-aspect-preserving stretch to fit - if it works). I would probably recommend removing it for simplicity... That is not part of this PR however :)

### DIFF
--- a/frontend/src/plugins/router/middlewares/login.ts
+++ b/frontend/src/plugins/router/middlewares/login.ts
@@ -41,7 +41,11 @@ export const loginGuard = async (
 ): Promise<Exclude<NavigationGuardReturn, Error>> => {
   const toServerPages = serverPages.has(to.name);
 
-  if (!jsonConfig.allowServerSelection && toServerPages) {
+  /**
+   * Do not allow the server selection pages if allowServerSelection=false,
+   * but do allow the login page.
+   */
+  if (!jsonConfig.allowServerSelection && toServerPages && to.name != '/server/login') {
     return false;
   }
 
--- a/frontend/src/plugins/router/middlewares/login.ts
+++ b/frontend/src/plugins/router/middlewares/login.ts
@@ -42,10 +42,10 @@ export const loginGuard = async (
   const toServerPages = serverPages.has(to.name);
 
   /**
-   * Do not allow the server selection pages if allowServerSelection=false,
+   * Do not allow the server selection pages if `allowServerSelection` is false in config.json,
    * but do allow the login page.
    */
-  if (!jsonConfig.allowServerSelection && toServerPages && to.name != '/server/login') {
+  if (!jsonConfig.allowServerSelection && (toServerPages && to.name !== serverLoginUrl)) {
     return false;
   }
 


hawken93:
#569

seanmcbroom:
The video element must be wrapped in a div element for jassub to render on. Not sure if this is necessary with the switch to ass.js though.

hawken93:
I tried to find a way to style the div element so it would work, but I was unsuccessful. Do you guys know of any way to do this that also works with the mini version?

frontend/src/components/Playback/PlayerElement.vue
```diff
@@ -4,29 +4,27 @@
...
+      <Component
+        :is="mediaElementType"
+        v-show="mediaElementType === 'video' && videoContainerRef"
+        ref="mediaElementRef"
+        :poster="String(posterUrl)"
+        autoplay
+        crossorigin
+        playsinline
+        :loop="playbackManager.isRepeatingOnce"
+        :class="{ 'uno-object-fill': playerElement.isStretched.value, 'uno-max-h-100vh': true}"
```
seanmcbroom, line 16: Could you test the following style? I don't have a ton of media to test with myself. I believe this would fix the problem with video sizing and would allow us to keep the parent div element.

`:class="{ 'uno-object-cover': playerElement.isStretched.value, 'uno-w-screen': playerElement.isStretched.value, 'uno-h-full': true }"`

hawken93: I tested with uno-object-cover uno-w-screen uno-max-h-100vh. For fullscreen it kinda worked, however it does not limit height. This causes a narrow window to cut video in the top and bottom.

Without the object-fit it seems a lot better. I wonder if the video container should always try to be wide.

```
:class="{ 'uno-w-screen': true, 'uno-h-full': true }"
```

I still don't get why the stretch option is wanted. I don't know anyone that asked for the video to not be scaled to fit the screen

hawken93: As far as I can find in forums, i.e. https://forum.jellyfin.org/t-stretch-as-deafult-please,
stretch seems to have been a sort of inbetween option between not scaling up, and scaling up with some compromise. If this fix works I don't know that the stretch option will be needed anymore

ferferga:
@hawken93 The PR that switches from JASSub to assjs that @seanmcbroom mentioned is https://github.com/jellyfin/jellyfin-vue/pull/2500

Perhaps you can test over that branch and see if subtitles still play nice without the wrapping div?

hawken93:
In preparation to test with assjs

hawken93:
Test with #2500.

In case of merge, please consider #2500 a blocker for this one now and let me clean this one up before merge :)

hawken93:
Alright, I have tested a little more now and I still prefer removing the div.

For the object-fit method to work, the video element seems to use the size of the parent element. However updating the size on the parent div didn't work for some reason. So to really fix this, the div should somehow inherit the size of its parent, but width 100% and height 100% stylings did not work. What's needed is probably some css inheritance magic so that the div becomes "transparent" to the calculations.

To work around this, the video element was styled to take up as much space as possible, and then object-fit helped decide what to do with that space. But then the mini video player did grew to 100% height because the 100%-styling of the div.. But it's necessary for the object-fit stylings to work.

By removing the div, the video element fills the parent element and it seems to work very nicely, and it removes the complexity of having to figure this out.

I can also verify that ASS subs still work with the div removed.

However, there are a couple more things I'd like to look at:
* The ASS subs aren't centered, they are left aligned and that causes an unintended effect whenever the video has a black bar to its left.
* I want to make sure that the video doesn't get paused when fullscreening from the mini video player. It's unintuitive/nonstandard/cumbersome and it doesn't pause when minimizing.

ferferga:
Sorry for not getting my review in this one until now, but I'm still trying to caught up in the assjs PR, aside from some real life stuff :). I'm also of the opinion that extra elements, if they don't serve a purpose, are unnecessary so I'm on board with it as soon as everything keeps working.

I think the focus of the library is to use modern web APIs as much as possible in a performant way, so things like improving API and fixing minor quirks (like the need for the extra div) it's lower in the priority list (ref https://github.com/weizhenye/ASS/issues/53#issuecomment-2466273246). We can try to contribute upstream an improvement for it regardless.

As for the fix for the pausing when fullscreening, if you find a fix, please do it in another PR so it's isolated and can be better tracked in a different commit. I believe it's cauised by a race condition.

ferferga:
@hawken93 We can come with your PR now. Rebases are highly welcome in this repo, so feel free to drop the other PR's comments and keep your commits only now that all  the required work is upstream.

hawken93:
Alright! I'm dropping the video scaling work now as it seems that a fix has been merged for it!
The only thing remaining is the oneliner where I let the login page work when the user is not permitted to add their own servers.

frontend/src/plugins/router/middlewares/login.ts
```diff
@@ -41,7 +41,11 @@ export const loginGuard = async (
 ): Promise<Exclude<NavigationGuardReturn, Error>> => {
   const toServerPages = serverPages.has(to.name);
 
-  if (!jsonConfig.allowServerSelection && toServerPages) {
+  /**
+   * Do not allow the server selection pages if allowServerSelection=false,
+   * but do allow the login page.
+   */
+  if (!jsonConfig.allowServerSelection && toServerPages && to.name != '/server/login') {
```
ferferga, line 48:
```suggestion
  /**
   * Do not allow the server selection pages if `allowServerSelection` is false in config.json,
   * but do allow the login page.
   */
  if (!jsonConfig.allowServerSelection && (toServerPages && to.name !== serverLoginUrl)) {
```

Please run ESLint before pushing to verify new code does not introduce new problems. This was being flagged by the ``eqeqeq`` rule. Also:
* The login url was defined as a constant in the top of the file and we were repeating it here unnecessarily
* Added parenthesis to clarify the different sections of the conditional for readability.
* If we're adding comments here (the logic is simple enough to be understandable and I don't believe it required comments in the first place) but if we're doing them anyway, let's them be top notch! 👌. I believe explaining that `allowServerSelection` comes from the config file (although it's implicit if someone tracks the source of the `jsonConfig` variable where further comments are located).

ferferga (changes requested):
The changes are simple enough and It's a matter of comitting my suggestion only LOL, but I wanted to know beforehand if you agree with my proposal or have any further input 👍

frontend/src/plugins/router/middlewares/login.ts
hawken93: I agree with all of this! It's been a while so I expected some styling edits :)

ferferga (approved):

PR #2505, opened 2024-11-25
chore(ecmascript): switch to ES2024

ferferga:
The ECMAScript upgrade is needed for `Object.groupBy`, which we use across the codebase

This can't be merged until TypeScript 5.7 has ubiquituous support by ``VSCode``, ``ESLint`` and `vue-tsc`


ferferga:
Something I've been forgetting to update is the browserslist query. However, there's no possibility to query by ECMAScript version: https://github.com/browserslist/browserslist/issues/795

TODO: Investigate with CanIUse or browserslist themselves in a way to support querying for specific ecma versions, since it would be much easier and maintenable to do it that way.

PR #2515, opened 2024-11-29
fix: navigation out of video player - VOverlay was hijacking navigation

hawken93:
not too much more to say.

closeOnBack according to vuetify documentation:

> Closes the overlay content when the browser's back button is pressed or `$router.back()` is called, cancelling the original navigation. `persistent` overlays will cancel navigation and animate as if they were clicked outside instead of closing.

### DIFF
--- a/frontend/src/components/Playback/MiniVideoPlayer.vue
+++ b/frontend/src/components/Playback/MiniVideoPlayer.vue
@@ -7,6 +7,7 @@
       :scrim="false"
       scroll-strategy="none"
       content-class="minimized-overlay"
+      :close-on-back="false"
       :width="$vuetify.display.mobile ? '60vw' : '25vw'">
       <div
         ref="videoContainerRef"
@@ -14,6 +15,7 @@
       <VOverlay
         :model-value="isHovering"
         contained
+        :close-on-back="false"
         height="100%"
         width="100%">
         <div class="d-flex flex-column">
--- a/frontend/src/pages/playback/video.vue
+++ b/frontend/src/pages/playback/video.vue
@@ -10,6 +10,7 @@
         v-model="overlay"
         contained
         scrim="transparent"
+        :close-on-back="false"
         width="100%"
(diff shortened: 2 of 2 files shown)


frontend/src/components/Playback/MiniVideoPlayer.vue
```diff
@@ -7,13 +7,15 @@
       :scrim="false"
       scroll-strategy="none"
       content-class="minimized-overlay"
+      :close-on-back="false"
```
ferferga, line 10:
```suggestion
      close-on-back="false"
```

The same thing applies to all the other occurrences. Doesn't it work this way for you?

ferferga (commented):
This PR is good enough as is, but aren't you interested in creating our own implementation in `JOverlay` so we have this component already stripped from Vuetify?

VOverlay has also been affected by memory leaks since Vuetify 3.x. They've been fixing them but I'm not in doubt that there are probably others since the amount of escape-hatches outside Vue they use it's astonishing (and sadly that component will still be used for VMenus and some other components, so we wouldn't be getting rid of it 100%). But hey, one component less, and in fact one of my biggest priorities to get out and an easy one to do (at least on paper).

frontend/src/components/Playback/MiniVideoPlayer.vue
hawken93: As for creating JOverlay I am interested but I have time commitment issues 😓  I imagine that our need for functionality from it will be limited and hopefully not a very big deal. I'm just going to prioritize that mini->fullscreen pausing bug for now and then I can circle around.
But I hope this PR can still stand on its own. It fixes a really annoying issue with some tiny code changes :)

ferferga (approved):
@hawken93 Thank you very much for your troubleshooting and good to know it sounds good to you to make the `JOverlay` component. Thank you for your interest :).

I'm merging this in the meantime🚀

PR #2521, opened 2024-12-03
feat: create JOverlay component

ferferga:
Replace VOverlay with it

Checklist:

- [ ] Show darker color on hover as optional prop
- [ ] Make it work properly in the video mini player
- [ ] Use it for GenericItemCard

PR #2526, opened 2024-12-04
feat: add box-shadow to GenericItemCard

Stereo8:
adds inner shadow to bottom edge of card, so action icons are more visible

Before: 
image

After: 
image

### DIFF
--- a/frontend/src/components/Item/Card/GenericItemCard.vue
+++ b/frontend/src/components/Item/Card/GenericItemCard.vue
@@ -163,6 +163,7 @@ const hasClick = computed(() => !!attrs.onClick);
 @media (hover: hover) and (pointer: fine) {
   .card-box:hover .card-overlay-hover {
     background: rgba(var(--j-color-background), 0.5);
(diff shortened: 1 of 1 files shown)


ferferga:
@Stereo8 Hello, first of all, thank you very much for your contribution

I've already opened a PR for creating the `JOverlay` component that replaces the `VOverlay` component from Vuetify, and also uniformizes the different "overlay" usages across the app at #2521. I'm currently super busy with exams, so I haven't had time to implement it properly for `GenericItemCard`. Given you already opened a PR for this, I pushed my local changes in a `wip` commit so you can take a look at it and maybe continue this work upon it? You just need to target this PR against the `j-overlay` branch instead of master.

There are probably some other style fixes needed for fully tackling the issue.

Besides, your modification here only works for dark mode, since it always adds a dark overlay. The `JOverlay` already makes use of the `j-color` variable that is updated depending on the user's color.

ferferga:
Superseded by #2537

PR #2531, opened 2024-12-08
fix: parallel seasons fetching

hawken93:
I noticed that seasons loaded sequentially upon loading the show page.

This code change introduces a new async function to handle the fetching of a single season, and this function is run in parallel and waited with Promise.all.

I've done a functional test and I also couldn't get the linter to stop OOMing.. So I cross my fingers that the code looks good!

### DIFF
--- a/frontend/src/components/Item/SeasonTabs.vue
+++ b/frontend/src/components/Item/SeasonTabs.vue
@@ -79,6 +79,23 @@ const seasons = ref<BaseItemDto[] | null | undefined>([]);
 const seasonEpisodes = ref<TvShowItem['seasonEpisodes']>({});
 
 /**
+ * Fetch episodes in one season
+ */
+async function fetch_season(season: BaseItemDto): Promise<void> {
+  const episodes = (
+    await remote.sdk.newUserApi(getItemsApi).getItems({
+      userId: remote.auth.currentUserId,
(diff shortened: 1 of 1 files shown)


ferferga (changes requested):
Hello and sorry for taking a bit to review.

I completely forgot that there were still places where we don't make use of the [api composables I made a while back](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/composables/apis.ts). These composables cache all the requests and make them reactive (if the item gets an update from server through the WebSocket, or it's edited through metadata editor), also if you change a parameter of the request, the request is re-performed automatically. You can see the full docs in the same file and in other components where they're used.

Can you switch to these composables here as well? Be aware that `useApi` is for requests that don't return `BaseItem`-related queries and `useBaseItem` is for the rest of them.

Try also the following approach (pseudocode), because I believe it's cleaner and doesn't unnecessarily allocates a new array:

```typescript
const all_seasons = await Promise.all(seasons.value.map(s => useBaseItem(...)(...)));
```

You can then create 2 additional ComputedRef, 1 for the season results and another for the loading state (in case the loading state is needed by some property, if not, just the one for the season results is enough) that maps the results as appropiate.

ferferga:
Superseded by #2547

PR #2553, opened 2025-01-08
fix(wizard): fix stepper error on wizard page

Eeymoo:
I have modified the stepper in the wizard according to the syntax of Vuetify 3.


Eeymoo:
@ferferga I'm sorry, I couldn't find the reason why the Actions did not pass. I'm not sure if it's due to historical reasons or my changes. I only modified the `frontend/src/pages/wizard.vue` file and the automatically generated `frontend/types/global/components.d.ts` file.

```
frontend/src/pages/wizard.vue
frontend/types/global/components.d.ts
```

ferferga (approved):
@Eeymoo No worries, we've been slowly fixing sthose type errors since the migration to Vue 3. Most of them are caused by Vuetify.

Thank you very much for your PR! If you're interested in further contributions, it would be great if you can port the rest of the components that were left behind (only skeletron cards are left, if I recall correctly. The boilerplate ones that appeared while content was loading is no longer useful, since the pages are completely mounted now when accessed, but the static ones are for displaying as placeholders when no items are returned by the API).

PR #2568, opened 2025-01-25
fix: failing Docker images from CI

ferferga:
After the merge of #2539, the Docker builds from CI stopped working properly. However, local and Codespace builds (even locally built Docker images) still are working fine, so no idea what's going on. This PR is a playground to test stuff since it can only be reproduced in CI-produced artifacts. When the fix is found, this same PR is going to contain the fixes.


ferferga:
The issue is only reproducible when using buildx

PR #2572, opened 2025-01-25
feat(tauri): implement fullscreen toggle functionality

ferferga:
Implements full window's fullscreen functionality inside @jellyfin-vue/tauri-runtime package

Supersedes #2503


ferferga:
@seanmcbroom The current code does correctly replace the native functions (you can check by running `npm run prod` inside the ``tauri-runtime`` package). However, Tauri itself doesn't seem to work with the native calls either (at least on Windows), even if I use `setInterval` to toggle between the states only (making sure this way that our "monkeypatch" it's not the issue).

Do you have any idea why? I believe this meant that the original PR didn't work either, at least on Windows.

Issue #2605, opened 2025-03-16
audio transcoding not working


BotBlake:
### Description of the bug

I have a few `.wma` files that cannot be played by jellyfin-vue.
I assume that this is caused by it having to transcode that but the server failing to do so.

Trying to start playback on finamp results in either "playback starting" but a non moving progress bar, or a never ending loading circle on the progress bar.
It always tells the server we would `Direct Play` and doesnt tell it that Playback has not actually started yet.

In Jellyfin-Vue it looks like so:

![Image](https://github.com/user-attachments/assets/74f40f66-3ce9-47a3-a36e-f26314bbebba)

Whilst in the jf-server Admin UI, it looks like its just playing:

![Image](https://github.com/user-attachments/assets/c8019242-27ca-41e0-a1bb-215156535856)

From looking at that, it seems like the Client is trying to DirectPlay.
Note that playing back the same file on jf-web and jf-androidtv (even on the same device) results in a Transcode + Transcode log.

This could be caused by two things:
1. Audio transcoding is simply not (yet) supported, and therefore the Client just Direct Plays and doesnt know how to handle unsupported Codecs.
2. Audio transcoding was working previously but broke due to a Server side change in 10.10.x

In case it it option two, you will find more detailed information about this HERE:
https://github.com/jellyfin/jellyfin/issues/12926

Main change is that the "AudioCodec" Parameter must now be send.

### Steps to reproduce

1. Get a file that will result in transcoding - e.g. `.wma`
For me a ``WMAv2 asf` - File did the trick.

2. Try to start Playback in the Jellyfin-Vue Client

### Expected behavior

I would have expected playback to work - Or at the very minimum get an Error message explaining how transcoding is not working.
Generaly having a handler for "unsupported file".

### Logs

```text
⚠ An AudioContext was blocked from automatic playback. It must be created or resumed after user interaction with the page.

⚠ Swiper Loop Warning: The number of slides is not enough for loop mode, it will be disabled and not function properly. You need to add more slides (or make duplicates) or lower the values of slidesPerView and slidesPerGroup parameters.

⚠ HTTP-Content-Type "video/x-ms-asf" is not supported. Loading media resource https://<domain>/Audio/<itemid1>/stream.asf?Static=true&mediaSourceId=<itemid1>&DeviceId=<sessionid1>&api_key=<sessionid4>&Tag=<sessionid5>&liveStreamId= failed.

[...]

PR #2607, opened 2025-03-18
fix(playback): audio transcoding

ferferga:
Ref: https://github.com/jellyfin/jellyfin/issues/12926


ferferga:
This was a workaround rather than a fix. The proper way of fixing this [is explained here](https://github.com/jellyfin/jellyfin-vue/issues/2605#issuecomment-2834526124) and, in the meantime, another PR fixing this is welcome.

PR #2646, opened 2025-05-12
feat: added Server Settings Page

PeachesMLG:
Added a page to modify server settings
image


frontend/src/pages/settings/server.vue
```diff
@@ -0,0 +1,255 @@
...
+          Paths
+        </h3>
+
+        <VTextField
+          v-model="serverSettings.state.value.cachePath"
+          label="Cache Path" />
+
+        <VTextField
+          v-model="serverSettings.state.value.metadataPath"
+          label="Metadata Path" />
```
PeachesMLG, line 44: I wanted to make a new component which would autocomplete the paths with a file explorer type thing, but thought Id do that in a different PR

frontend/src/pages/settings/server.vue
```diff
@@ -0,0 +1,255 @@
...
+
+    serverSettings.value.state.value.loginDisclaimer = branding.LoginDisclaimer;
+    serverSettings.value.state.value.customCSS = branding.CustomCss;
+    serverSettings.value.state.value.enableSplash = branding.SplashscreenEnabled;
+  } catch (error) {
+    console.error('Error loading settings:', error);
+  } finally {
+    loading.value = false;
+  }
+});
```
PeachesMLG, line 254: Honestly not sure how I feel about this code, was considering changing it to a hook or abstracting it away somewhere else, thought id get opinions first though

frontend/src/pages/settings/server.vue
```diff
@@ -0,0 +1,255 @@
...
+              Cancel
+            </VBtn>
+            <VBtn
+              color="primary"
+              @click="cssDialog = false">
+              OK
+            </VBtn>
+          </VCardActions>
+        </VCard>
+      </VDialog>
```
PeachesMLG, line 115: This is a popup window to add the css onto.

image

PeachesMLG:
Looks like some of the github actions are failing, but they look like nothing todo with me and are appearing on other PR's so I will ignore them

frontend/src/pages/settings/server.vue
ferferga: It already exists - check [the API composables I've made to simplify cases like this](https://github.com/jellyfin/jellyfin-vue/blob/6c42d848c4a77aaf50f11c3db40b0457c8c78bb2/frontend/src/composables/apis.ts). These composables:
* Track the changes to the parameters, reexecuting the request when they change. 
* Cache the requests, so consequent access to this page will be immediate. The request will be made anyway on mount, but the data will be updated as soon as the cache is refreshed.
* As the cache is reactive, if the updates are received from the websocket (though currently these don't apply to server settings), the data will be updated accordingly. Example: editing an item's metadata while a music track is playing means that the footer updates with the modal. Favoriting/mark as watched is also applied to all the current rendered components in the application (though I've found some edge cases for this I have in my TODO).

They are also helpful for updating server data, you can reduce all of this to a single call of the composable and changing the parameters accordingly. [See how it's done when liking an item](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/components/Buttons/LikeButton.vue). **Just initialize it with the method for getting the server data and change it to the methods that update the data whenever the first change to any of the form elements occur (make sure to set `skipCache` when updating)**

You can check more examples by seeing other components and in the documentation of the functions (I know the file itself it's really large and I want to refactor it to reduce the complexity a bit, but at least is fully documented). At first it might seem daunting, but I think it's really straightforward.

Also, we are using [`Suspense`](https://vuejs.org/guide/built-ins/suspense.html#suspense), so all the async/await/promise stuff must be at the top level `<script setup>`, not in `onMounted` - this way, the navigation is blocked until all the promises have been settled.

ferferga (changes requested):
> Looks like some of the github actions are failing, but they look like nothing todo with me and are appearing on other PR's so I will ignore them

Yeah, most of them are caused by `VButton`, as I've discussed with you on Matrix.

> Replaced because of conventional commits

About this comment in the other PR - Learn a bit about git rebase, you will thank me later :). Extensions like GitLens also simplifies a lot the rebasing by showing a nice graph of how the commits are affected while editing/squashing or reordering them.

frontend/src/pages/settings/server.vue
```diff
@@ -0,0 +1,256 @@
...
+  <SettingsPage>
+    <template #title>
+      Server Settings
+    </template>
+
+    <template #content>
+      <VCol
+        md="6"
+        class="uno-pb-4 uno-pt-0">
+        <h3 class="mb-2 text-lg font-bold">
```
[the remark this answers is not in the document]
PeachesMLG: I believe these are all supported by UnoCSS? I made a playground [here](https://unocss.dev/play/#html=DwCwzABAxgNghgZwQXgEQFsBGBaATBAFwFMAPA7GAcwgDMB7AO3MzpgBNUA%2BAFSIQOAB6cJwBQQA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAHU8aAMyUAvnCxQIIOAHIuuCBLp0DSJBgAekWOmwBDLgBt4mHPiIlgpABTIqFBu9ABccADaVKiRBhJcdDB6ALQhrhgG5IhwUq7QEQa0aAZwGgC6spVUdAAW0DAJMHQRQaiGCUmpdQ1NBoUwtjAprqRwgzbD0E64pBhw9QBuGFBhE8ODTq5ZVBqyIgotUTFyokoqqv4AlLLC8owsbJw8fPzXt6cKktKBJ6h0Ei2GAiAEYAHQAJg%2BqAkaFwhVqrDALQA9Cj6CAwXUUTt2mUblRqhorkggA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4XyA) and it parsed it fine?

Unless im missing something?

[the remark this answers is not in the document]
ferferga: @PeachesMLG Those work because we also currently have [Vuetify's helper classes](https://vuetifyjs.com/en/styles/spacing/#usage). Those match Tailwind's sometimes, but sometimes they do follow their own convention (which means its more difficult for newcomers, since you need to learn a new naming schema).

Sadly, they need to coexist so current components work and there's no way to properly opt out of those ([it's a single import](https://github.com/jellyfin/jellyfin-vue/blob/master/frontend/src/plugins/vuetify.ts#L6)). That's why we have the `uno-` prefix, so they don't clash with each other and we are sure where the new unocss classes are being used or not. Here either:

* The styles are doing nothing and can be removed
* Vuetify's syntax matches with Tailwind's, so you are using Vuetify's classes instead.

It's then a matter of checking if those classes are actually doing something and prefix them.
[I've made a gist with the generated Vuetify's stylesheet for reference](https://gist.github.com/ferferga/0cfaf818f0a219640ac8e5e23f2b239f), attaching it in case it's useful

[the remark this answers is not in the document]
PeachesMLG: Ok, im with you, makes sense :+1:

ferferga (approved):

PR #2662, opened 2025-06-16
Allow dupeplicate imports

Alexk2309:
I encountered an issue during development where the system was treating the 'vue-router/auto-routes' imports as if they were part of the main 'vue-router' library. As a result, every time I saved the file, it would incorrectly try to merge those imports together, which was clearly not the intended behavior.

These changes should prevent the issues.

### DIFF
--- a/frontend/eslint.config.ts
+++ b/frontend/eslint.config.ts
@@ -14,5 +14,10 @@ export default defineConfig([
       'types/global/routes.d.ts',
       'types/global/components.d.ts'
     ]
+  },
+  {
+    rules: {
+      'import/no-duplicates': 'error'
+    }
   }
 ]);


ferferga:
I'm not a fan of this - it's a quick workaround, yes, but why not report this issue upstream? Are you interested in making a quick playground on Stackblitz and opening [the issue](https://github.com/un-ts/eslint-plugin-import-x/issues)?

**EDIT**: It seems [it exists already](https://github.com/un-ts/eslint-plugin-import-x/issues/308)
**EDIT 2**: [It seems this one is even a better fit](https://github.com/un-ts/eslint-plugin-import-x/issues/372). I don't understand why maintainers thing is an expected outcome though, package.json from vue-router is worth investigating I guess.

ferferga:
Not needed after upgrading to vue router v5.y.z, since the auto routes are now part of the main package.

PR #2663, opened 2025-06-16
Add fix for routers

Alexk2309:
These are general router-related fixes. I was running into a few issues—specifically, when reloading the page while on a nested route, it would always redirect to the root (`/`) route, which isn't correct. I resolved this by adding `await router.isReady();` to ensure the router is fully initialized before proceeding.

Additionally, I can't see why the circular dependency issues were happening in the first place. I believe it would be cleaner and more maintainable to consolidate all router-related logic into the router file itself, as it's a better design practice.

Of course, if you have any concerns or strong reasons against this approach, I'm happy to revisit and revise the PR. Thanks!

This has all been tested and works fine.

### DIFF
--- a/frontend/src/components/Layout/MainView.vue
+++ b/frontend/src/components/Layout/MainView.vue
@@ -74,7 +74,8 @@ const { comp, route } = defineProps<{
 }>();
 
 const id = useId();
-const root = isNil(inject(JView_isRouting));
+const injectedRouting = inject(JView_isRouting, null);
+const root = isNil(injectedRouting);
 
 const resolved = computed({
   get() {
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -5,7 +5,6 @@
  */
 import '@jellyfin-vue/i18n';
 import { createApp } from 'vue';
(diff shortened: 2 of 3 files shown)


Alexk2309:
image

https://github.com/user-attachments/assets/d9e97507-6cc6-4dea-90c8-de55942c4876



@ferferga

ferferga:
@Alexk2309 The issue is only during development, this clears out some things - it looks like to be a mismatch between production during HMR. I guess the screenshot is just for showing better the same console logs that are shown in the video?
Ideally this should be investigated so we can report this upstream though.

There are 2 solutions:
* Wrapping the call in `if (import.meta.env.DEV)` so it's ripped out for production
* Pass `undefined` as the flush option during development

I prefer 1 since 2 changes the behaviour and creates another bigger mismatch between dev and prod. Add a comment before the if, with a TODO, so we know in the future we need to investigate it further to report upstream.

ferferga:
This doesn't happen anymore with the current dependency set as of today (this was probably fixed by Vite 8, since the dev and prod environments are no longer mismatched, due to both using rolldown under the hood)

PR #2664, opened 2025-06-16
added fix to show progress meter

Alexk2309:
The progress meter wasn't showing, and this fix should make it work.

### DIFF
--- a/packages/frontend/src/components/Item/Card/GenericItemCard.vue
+++ b/packages/frontend/src/components/Item/Card/GenericItemCard.vue
@@ -41,7 +41,8 @@
           "
           :model-value="progress"
           absolute
-          location="bottom" />
+          location="bottom"
+          class="uno-z-50" />
       </div>
     </Component>
     <div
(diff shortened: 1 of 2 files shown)


frontend/src/components/Item/Card/ItemCard.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <GenericItemCard
-    :progress="progress"
+    :progress="item.Type !== BaseItemKind.Series ? progress : undefined"
```
Alexk2309, line 3: It doesn't really make sense for a series to have a progress value. It seems like it's actually reflecting the progress of the most recent episode rather than the series as a whole. I understand if this approach isn't ideal, and I'm open to changing it if there's a better way to handle it.

ferferga:
Please use conventional commits from now on (no need to reword/ammend the current ones, because I will squash this PR) but just FYI from now on :)

ferferga:
@Alexk2309 Hello, I'm sorry this hasn't been merged before, I don't even remember why, perhaps because of the series thing, which is not ideal, but it can be solved in another PR.

Thank you very much for your contribution.

PR #2779, opened 2026-04-10
feat: account  page

poytiis:
Created an Account page that allows users to edit their password and profile image. The page is accessible via Profile Image → Settings → Account.

Screenshot_account


ferferga (approved):
Excellent work!! Thank you very much!

(I apologize for taking a bit to review 😅)

Issue #2603, opened 2025-03-15
Missing subtitle selection


alvitali:
### Description of the bug

Newly installed jellyfin-vue, but subtitle selector is greyed out, and I cannot add subtitles (which Jellyfin sees). Neither .srt, PGS or DVDSUB.

Judging by other issues on here, subtitles should be a feature that's available

### Steps to reproduce

New install of jellyfin-vue, using DEFAULT_SERVER and DISABLE_SERVER_SELECTION both true

### Expected behavior

I can click the subtitles icon in the video player and choose subtitles to be displayed

### Logs

```text
n.a.
```

### Platform

Linux

### Browser

Safari

### Jellyfin server version

10.10.6

### Additional context

```text

```

ferferga:
From the expected behaviour, I guess you mean the button from the video player? Did you wait for the video to be fully loaded before trying to toggle subtitles? Can you maybe attach a video?

alvitali:
hey ferferga

yes exactly, the subtitle button from the video player. what do you mean with fully loaded? the video is playing, but not buffered in its entirety i assume (cannot tell how far buffer is progressed). even after playing for a few minutes, this is what the subtitle selector looks like:

<img width="453" alt="Image" src="https://github.com/user-attachments/assets/47ce1594-1bd2-4f37-b1a7-6c086304d928" />

The file in question has both embedded PGSSUB as well as external .srt subtitles (and embeded ASS subtitles, but these might be tricky anyways)

ferferga:
The ones in the cog icon doesn't work either? I'm in the middle of recovering my media after a HDD failure, so I I'm just guessing and right now I'm not sure if there was a logic difference between the button inside the submenun and the main one.

alvitali:
Sorry to hear, good luck with the recovery :/

I don't have a subtitle button in the settings menu inside the video player:

<img width="454" alt="Image" src="https://github.com/user-attachments/assets/4877b5c6-8346-42f3-9c25-90b365a39542" />

I've noticed everytime I open the settings menu inside the player for the first time on a given playback, the playback stutters / goes black, and resumes after 2ish seconds

alvitali:
Which hamburger menu? I'm on commit #69716c0 right now

Also, when I open the settings page, I cannot access the options for Account, Home Screen, Playback or Media Players, as these are all grayed out as well:

[...]

PR #2814, opened 2026-05-17
fix(subtitle): missing, cannot select or disable, and default selection subtitle

ghostvar:
based on #2603 issue, missing and instability when load subtitle, some part directly manipulate live object which I'm little afraid of, and computedAsync property return undefined value when contain await in it.

I'm just tested in my jellyfin server version 10.11.8 and only tested several video with .srt and .ass subtitle format.


packages/frontend/src/pages/playback/video.vue
```diff
@@ -6,11 +6,7 @@
     @mousemove.passive="handleMouseMove"
     @touchend.passive="handleMouseMove">
     <JOverlay
-      class="uno-h-full uno-flex uno-flex-col uno-items-center uno-justify-between"
-      :class="{
-        'uno-opacity-100': overlay,
-        'uno-opacity-0': !overlay
-      }">
+      class="uno-h-full uno-flex uno-flex-col uno-items-center uno-justify-between" :hover="!overlay">
```
ferferga, line 9: Why?

packages/frontend/src/pages/playback/video.vue
```diff
@@ -159,6 +155,7 @@ const overlay = computed({
 });
 
 const timeout = useTimeoutFn(() => {
+  console.log('overlay', false)
```
ferferga, line 158: Unnecessary

packages/frontend/src/store/playback-manager.ts
```diff
@@ -142,7 +145,21 @@ class PlaybackManagerStore extends CommonStore<PlaybackManagerState> {
...
+    /**
+     * DeliveryMethod and DeliveryUrl props for MediaStream type Subtitle,
+     * this manual properties building for that currentItemParsedSubtitleTracks from player-elements required.
+     */
+    const ApiKey = RemotePluginAuthInstance.currentUserToken.value;
+    const mediaSource = this.currentItem.value?.MediaSources?.[this.currentMediaSourceIndex.value ?? 0];
+    !isNil(mediaSource) && Array.isArray(mediaSource.MediaStreams) && mediaSource.MediaStreams.map((Stream: MediaStream, Index: number) => {
+      return Stream.Type === MediaStreamType.Subtitle && !isNil(ApiKey) ? Object.assign(Stream, {
+        DeliveryMethod: SubtitleDeliveryMethod.External,
+        DeliveryUrl: `/Videos/${mediaSource?.Id}/${mediaSource?.Id}/Subtitles/${Index}/0/Stream.${Stream.Codec}?ApiKey=${ApiKey}`
```
ferferga, line 158: Isn't there any TS SDK method for this?

packages/frontend/src/store/player-element.ts
```diff
@@ -130,9 +130,10 @@ class PlayerElementStore extends CommonStore<PlayerElementState, 'isStretched' |
     ) as SubtitleExternalTrack;
 
     if (this._useCustomSubtitleTrack.value) {
-      const data = await runGenericWorkerFunc('parseVttFile')(el.src);
-
-      el.parsed = data;
+      // when use await this computedAsync values become undefined in several applySubtitle logic
+      runGenericWorkerFunc('parseVttFile')(el.src).then(parsed => {
+        el.parsed = parsed;
+      });
```
ferferga, line 136: This is the same as the previous implementation, there is no reason for this to switch it from await, unless there's something really strange happening down on Vite/Rolldown when transpiling the code.

Do you have steps to reproduce where this works, but await doesn't?

packages/frontend/src/store/player-element.ts
```diff
@@ -206,10 +207,13 @@ class PlayerElementStore extends CommonStore<PlayerElementState, 'isStretched' |
       const trackSrc = this.currentExternalSubtitleTrack.value.src;
       const subtitleTrackPayload = await this._fetchSubtitleTrack(trackSrc);
 
-      if (subtitleTrackPayload[trackSrc]) {
-        /**
-         * video_width works better with ultrawide monitors
-         */
+      /**
+       * only load when no this._asssub, sometime this method called twice intermediacy with no time to _clear,
```
ferferga, line 211: How is this possible, what's the underlying issue? We call this._clear() before

ferferga (commented):
Hey @ghostvar thank you very much for your contribution and welcome to the project! 🎉

Sorry for taking so long to answer, but I'm in the middle of moving, so I had less than usual spare time. Some comments, I believe there are some testing bits (the class editing in JOverlay) left as well.

packages/frontend/src/pages/playback/video.vue
ghostvar: Before I start, I had mistaken to push JOverlay issue to master branch. But about this fix I see component has hover props to control opacity in ui-toolkit:
https://github.com/jellyfin/jellyfin-vue/blob/12dc9da12b69c0ffb4c7bec774b1195dd596f354/packages/ui-toolkit/src/components/JOverlay.vue#L23
So, why need to manually swich uno-opacity-0 to uno-opacity-100 when where was props for it, In my case overlay is stuck and two class is fighting each to other (showing uno-opacity-100 and uno-opacity-0 at same time) so I just fix it.

packages/frontend/src/store/playback-manager.ts
ghostvar: Yes it has SDK showing object SubtitleDeliveryMethod and DeliveryUrl, but instead of giving value there is undefined, as I say in https://github.com/jellyfin/jellyfin-vue/issues/2603#issuecomment-4466936236 still I suspect it different endpoint.

Still had no idea why this happen, maybe because of the lower Jellyfin version? I haven't tested it yet.

packages/frontend/src/store/player-element.ts
ghostvar: Regarding the steps to reproduce this issue, do you mind if I record a video on using await or not?

packages/frontend/src/store/player-element.ts
ghostvar: And also steps to reproduce this one

ferferga:
Hello @ghostvar are you interested in keeping this PR alive?

ghostvar:
Hi @ferferga, Yes I am definetelly insterested in keeping this PR alive. Regarding the steps to reproduce the issue (where this approach works but await doesn't and other issue), would it be okay if I record a short video to demonstrate the exact behavior? Let me know, I will provide it asap.

ghostvar (commented):
Addressing comments

ghostvar:
Hi @ferferga,
First of all, I am so sorry for the 3-month silence! I just realized that my replies from back then were stuck in "Pending" status all this time because I completely missed the "Submit review" button 🤦‍♂️🤦‍♂️.

I am still quite new to open-source contributions, so I honestly thought I was waiting for your reply, while you were waiting for mine! I am really, really apologies for the confusion.

PR #2861, opened 2026-09-08
feat: merge crew entries for people with multiple roles

shbernal:
**Changes**

The server sends one `People` entry per credit, so a person credited for several jobs on the same item is listed once per job — Edgar Wright shows up twice in the crew list of *Scott Pilgrim vs. the World*, same avatar, once as Director and once as Screenplay. This combines the credits belonging to the same person into a single entry listing every role, e.g. "Director / Screenplay / Producer". Actors are untouched, so someone playing multiple characters still gets an entry per character.

It follows what jellyfin-web does since https://github.com/jellyfin/jellyfin-web/pull/8209 (shipped in 12.0), using the same merged types — director, writer, producer, which is also what the server keeps in `TmdbUtils.WantedCrewKinds` — and the same `" / "` separator.

The merging lives in a new `getMergedCrew()` in `utils/items.ts`, used by the `crew` computed on both the item and the series page.

One deliberate extra: the crew filter was `['Director', 'Writer']`, so producers never appeared in the crew list at all, unlike in the web client. They are included now. If that was intentional and you would rather keep producers out, say so and I will drop that part — the merging itself works either way.

**Code assistance**

Written with Claude Code using the Opus 5 model. It audited which clients were missing the behavior, wrote `getMergedCrew()` and the call sites, and verified the result by running the dev server against a Jellyfin 12.0 server and driving it in a headless browser. I reviewed the change myself.

Verified against a real library. Crew list for *Scott Pilgrim vs. the World*, before:

```
Edgar Wright         Director
Michael Bacall       Screenplay
Bryan Lee O'Malley   Graphic Novel
Edgar Wright         Screenplay
```

after:

```
Edgar Wright         Director / Screenplay / Producer
Michael Bacall       Screenplay
Bryan Lee O'Malley   Graphic Novel
J. Miles Dale        Executive Producer
Eric Gitter          Producer
...
```

`vue-tsc` and `eslint` report no new problems on the touched files (the repo has a number of pre-existing type errors in storybook and ui-toolkit; the set is identical with and without this change).

I only reproduced the duplicate entries on Android TV originally, but the same gap exists here, in Swiftfin and in jellyfin-roku, so I am opening the equivalent PR against each of those rather than leaving the fix partial across clients.

**Issues**

N/A — no existing issue. Follows up on https://github.com/jellyfin/jellyfin-web/pull/8209.

### DIFF
--- a/packages/frontend/src/pages/item/[itemId].vue
+++ b/packages/frontend/src/pages/item/[itemId].vue
@@ -306,7 +306,7 @@ import { getLibraryApi } from '@jellyfin/sdk/lib/utils/api/library-api';
 import { getUserLibraryApi } from '@jellyfin/sdk/lib/utils/api/user-library-api';
 import { computed, ref } from 'vue';
 import { useRoute } from 'vue-router';
-import { getItemDetailsLink, getMediaStreams } from '#/utils/items.ts';
+import { getItemDetailsLink, getMediaStreams, getMergedCrew } from '#/utils/items.ts';
 import { getItemizedSelect } from '#/utils/forms.ts';
 import { useBaseItem } from '#/composables/apis.ts';
 import { useItemBackdrop } from '#/composables/backdrop.ts';
@@ -341,9 +341,7 @@ const currentAudioTrack = ref<number>();
 const currentSubtitleTrack = ref<number>();
 
 const crew = computed<BaseItemPerson[]>(() =>
-  (item.value.People ?? []).filter(person =>
-    ['Director', 'Writer'].includes(person.Type ?? '')
-  )
+  getMergedCrew(item.value.People ?? [])
 );
 
 const actors = computed<BaseItemPerson[]>(() =>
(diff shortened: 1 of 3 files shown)


ferferga:
@shbernal Hello, thank you very much for your contribution! Would you mind fixing the code smell pointed out by SonarCloud?

shbernal:
Fixed, thanks. The `continue` is gone.

One thing you should know before merging. After I opened this, a maintainer on jellyfin-androidtv declined the equivalent PR there as server work (jellyfin/jellyfin-androidtv#5802), and I agree with him, so I've opened jellyfin/jellyfin#17896 proposing the API expose a person's credits as a list instead of one `BaseItemPerson` per credit.

I still think this is worth merging. jellyfin-web has had the same workaround since jellyfin/jellyfin-web#8209, the API change is some way off, and vue will need to handle flat `People` from 12.x servers regardless. But it is a stopgap, and `getMergedCrew` should go away once the server sends the credits grouped. Happy to close this instead if you'd rather wait.

ferferga (approved):
Thanks for pointing that out, in fact I remember we've been discussing this for some time. Regardless, it's a nice QoL improvement that is simple enought to land quickly, and we don't even know when the API rewrite will be done, so it's great to have anyway.

Thank you very much for your contribution! Much appreciated!

